PR #2, opened 2021-06-05
adding api tests for e2e testing

chajain:
Adding apis tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,67 @@
-# tests
-Harvester test cases
+# About 
+---
+This is a harvester test framework which uses python pytest library. This framework heavily depends on the pytest fixtures which are defined in conftest.py
+
+## Pre-requisites
+-------------
+
+Before running these tests, bring up a virtual harvester environment following the instructions in repo https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester
+
+## Pytest Fixtures
+-------------
+They help us set up some helper code that should run before any tests are executed, and are perfect for setting-up resources that are needed by the tests. They are defined in conftest.py under the apis folder.
+
+These fixture are executed at the session level and they are used to provide the authentication and to get the api endpoints before executing tests for a particular api
+
+## Running Tests 
+---
+
+- Create a python virtual environment, for example: test_venv
+```shell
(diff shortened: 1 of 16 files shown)


juadk:
Hello, JFYI, I've tested your new API tests inside a two nodes cluster built automatically by openQA (Testing Framework used at SUSE for operating system testing) and all looks good. 
I'll write a confluence page to explain the process asap.
image

mbelur:
Looks good to me!
couple of things which can be addressed  as an enhancement later:
1. Could we use random names for the resources used in each test?
2. We should add cleanup of resources regardless of the test outcome whether success or fail.

README.md
```diff
@@ -1,2 +1,59 @@
...
+This is a harvester test framework which uses python pytest library. This framework heavily depends on the pytest fixtures which are defined in conftest.py
+
+## Pre-requisites
+-------------
+
+Before running these tests, bring up a virtual harvester environment following the instructions in repo https://github.com/harvester/ipxe-examples/tree/main/vagrant-pxe-harvester
+
+## Pytest Fixtures
+-------------
+They help us set up some helper code that should run before any tests are executed, and are perfect for setting-up resources that are needed by the tests. They are defined in coftest.py under the apis folder.
```
bk201, line 12:
```suggestion
They help us set up some helper code that should run before any tests are executed, and are perfect for setting-up resources that are needed by the tests. They are defined in conftest.py under the apis folder.
```
nit

bk201:
LGTM!

Note some response codes and messages might be different after https://github.com/harvester/harvester/pull/919 is merged.
Since the document encourages users to use a virtualenv, It would be nice to have a tox file and add some lint check (like flake8) in the future.

bk201 (approved):

guangyee:
> LGTM!
> 
> Note some response codes and messages might be different after [harvester/harvester#919](https://github.com/harvester/harvester/pull/919) is merged.
> Since the document encourages users to use a virtualenv, It would be nice to have a tox file and add some lint check (like flake8) in the future.

That's a good suggestion. Will add that in a follow-on PR.

guangyee (approved):
Need to add tox later. But this is a good starting point.

PR #4, opened 2021-06-16
add scenario tests framework

guangyee:
Utilizing pytest and jinja2 templating to simplify the process in
developing scenario tests. Test developers can easily create request
templates and load them into JSON request objects by provide only the
required fields. The templates, along with their corresponding fixtures
are designed to be reusable. With templating, the goal is to reduce the
complexity of having to handcraft the JSON request in every test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,4 @@ __pycache__/
 *.log
 assets
 test_result.html
-
+*venv
--- a/README.md
+++ b/README.md
@@ -1,9 +1,8 @@
 # About 
----
+
 This is a harvester test framework which uses python pytest library. This framework heavily depends on the pytest fixtures which are defined in conftest.py
 
 ## Pre-requisites
--------------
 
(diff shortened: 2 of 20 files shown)


mbelur:
Looks Good! one suggestion is to move the common code like utils, fixture and templates so it can also be consumed by the API tests.

guangyee:
> Looks Good! one suggestion is to move the common code like utils, fixture and templates so it can also be consumed by the API tests.

Good suggestion! We probably will need to re-org the directories to have a top level parent directory `harvester-test`. i.e.

tests/harvester-tests
tests/harvester-tests/fixtures
tests/harvester-tests/templates
tests/harvester-tests/scenarios
tests/harvester-tests/apis

I can do that in a follow-on PR.

mbelur (approved):
LGTM!

guangbochen (approved):
LGTM

PR #5, opened 2021-06-22
Re-align API and scenario tests to share the same fixtures

guangyee:
Reorganize the tests to share the same fixtures and hance reduce code
complexity and duplication.

Also fixed the pep8 errors.


harvester_e2e_tests/apis/test_hosts.py
```diff
@@ -0,0 +1,68 @@
...
+def test_verify_host_maintenance_mode(admin_session, harvester_api_endpoints):
+    resp = admin_session.get(harvester_api_endpoints.list_nodes)
+    assert resp.status_code == 200, 'Failed to list nodes: %s' % (resp.content)
+    host_data = resp.json()
+    # NOTE: always test on the first node for now
+    resp = admin_session.post(
+        host_data['data'][0]['actions']['enableMaintenanceMode'])
+    assert resp.status_code == 204, (
+        'Failed to update node: %s' % (resp.content))
+    polling2.poll(
```
mbelur, line 44: Could we use the utils.poll_for_resource_ready here? we can use the timeout rather than poll_forever option

mbelur (commented):
Could we update the README to add a note on how to run the harvester_e2e_tests which now includes both the api and scenario tests?

harvester_e2e_tests/apis/test_hosts.py
guangyee: good catch! I missed that one.

chajain (approved):
Looks good

guangyee:
> Could we update the README to add a note on how to run the harvester_e2e_tests which now includes both the api and scenario tests?

just the standard `tox -e py38 -- harvester_e2e_tests --endpoint ...` will run both tests. Let me add a note to the README.

mbelur (approved):
LGTM!

PR #8, opened 2021-06-28
Use try-fetch-update loop when updating Kubernetes resources

guangyee:
In Kubernetes paradigm, we must fetch the resources to be updated to get
the latest resource version whenever we encounter an HTTP 409 during resource
update. The reason is that the resource had been updated by another process
and therefore the resourceVersion to be updated is already out of date. In
this case, we must fetch the resource to get the latest resourceVersion and
try again, untill the resource is successfully updated.


mbelur (approved):
LGTM

chajain (approved):
Looks good

PR #12, opened 2021-07-09
Add more VM test cases

guangyee:
1. add test cases for cloud-init for both user and network
2. add test case for creating a VM without starting it


chajain (approved):
goodLooks

PR #13, opened 2021-07-13
Add test cases for VM scheduling behavior

guangyee:
1) test VM schedule only on the nodes that satisfied the CPU requirement
2) test VM schedule only on the nodes that satisfied the memory requirement
3) test VM schedule only on the nodes that satisfied both CPU and memory
   requirements


chajain (approved):
Looks good

PR #14, opened 2021-07-14
Add VM with externally created volume tests

guangyee:
This patch also refactored the common code and moved them to utils so
they can be shared among the tests.


chajain (approved):
Looks good

PR #15, opened 2021-07-14
Add test case for Rancher authentication

guangyee:
Make sure the Harvester user can also login to Rancher.


chajain (approved):
Looks good

PR #22, opened 2021-07-20
Add host management framework to facilitate host test scenarios

guangyee:
Add a framework to enable tests which required host manipulations (i.e.
power-off, power-on, and reboot).

This patch also removes the duplicate VM volume tests.


chajain (approved):
Looks good

PR #23, opened 2021-07-21
set the default values for the parameters from config.yml

mbelur:
This change reads from the config.yml the default values for the
parameters.
This can be overridden with the respective command line option.


chajain (approved):
Looks good

PR #25, opened 2021-07-22
Add tests for VM networking and connectivity

guangyee:
1. make sure we can create a public VLAN
2. make sure we can SSH into the VM using it's public IP
3. make sure we can add a second NIC to the VM


mbelur (approved):
LGTM!

chajain (approved):
Looks good

PR #33, opened 2021-07-30
Add network negative tests

guangyee:
1. Make sure the VLAN IP is still pingable after a reboot.
2. Make sure the management IP of a VM is still pingable from another VM on
   the subnet after a reboot.

This PR also disables admin password reset as it is no longer needed. Admin
password reset is only enforce in the UI, not in the API.


mbelur (approved):
LGTM!

chajain (approved):
Looks Good

PR #34, opened 2021-08-02
P2 VM tests for Volume

chajain:
Create VM with two disk volumes
Create VM with one disk volume and one CDROM volume

### DIFF
--- a/harvester_e2e_tests/fixtures/vm.py
+++ b/harvester_e2e_tests/fixtures/vm.py
@@ -154,3 +154,41 @@ def vm_with_volume(request, admin_session, image, volume, keypair,
                               volume=volume,
                               keypair=keypair)
     return vm_json
+
+
+@pytest.fixture(scope='class')
+def vm_with_two_disks(request, admin_session, image, keypair,
+                      harvester_api_endpoints):
+
+    volume1 = utils.create_volume(
+        request, admin_session, harvester_api_endpoints,
+        description='Test volume', size=8)
+    volume2 = utils.create_volume(
+        request, admin_session, harvester_api_endpoints,
+        description='Test volume 2nd Disk', size=8)
+    vm_json = utils.create_vm(request, admin_session, image,
+                              harvester_api_endpoints,
+                              template='vm_with_two_disks',
+                              volume=volume1,
+                              volume2=volume2,
+                              keypair=keypair)
+    return vm_json
+
+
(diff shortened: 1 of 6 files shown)


harvester_e2e_tests/utils.py
```diff
@@ -336,14 +336,17 @@ def _wait_for_image_become_active():
 def create_vm(request, admin_session, image, harvester_api_endpoints,
               template='basic_vm', keypair=None, volume=None, network=None,
               cpu=1, disk_size_gb=10, memory_gb=1, network_data=None,
-              user_data=None, running=True):
+              user_data=None, running=True, volume2=None):
```
guangyee, line 339: For future consideration, I would suggest doing `additional_volumes=[]` instead of `volume2=None` in case we more volumes to add. We don't have to change it now, just something to keep in mind.

guangyee (approved):
It would be nice to make the additional volumes generic instead of just volume2. I am fine with merging this one as is. But we'll need to refactor this code if additional volumes are needed.

harvester_e2e_tests/fixtures/vm.py
```diff
@@ -154,3 +154,41 @@ def vm_with_volume(request, admin_session, image, volume, keypair,
...
+                              volume2=volume2,
+                              keypair=keypair)
+    return vm_json
+
+
+@pytest.fixture(scope='class')
+def vm_with_disk_cdrom(request, admin_session, image, keypair,
+                       harvester_api_endpoints):
+
+    cdromvol = utils.create_volume(
```
guangyee, line 182: I don't think we need to explicitly create the CDROM volume.

harvester_e2e_tests/templates/vm_with_disk_cdrom.json.j2
```diff
@@ -0,0 +1,165 @@
+{# Template for basic VM creation request. Start the VM instance upon #}
```
guangyee, line 1: I don't think we need this template. The basic_vm template already have a cdrom device. See https://github.com/harvester/tests/blob/main/harvester_e2e_tests/templates/basic_vm.json.j2#L100

guangyee (changes requested):
Sorry I just found out that we already have test cases for external volume and cdrom. We may not need this PR anymore as the cases are covered in the basic VM tests.

chajain:
These tests are already there

PR #36, opened 2021-08-03
test VM CPU and memory overcommit

guangyee:
1. Test VM instance unschedulable when CPU is 10000
2. Test VM instance unschedulable when memory is 10000

Also refactored utils to make the code sharable.


chajain (approved):
Looks Good

PR #38, opened 2021-08-05
Add negative network tests

guangyee:
1. Test creating a network with no name.
2. Test creating a VM with a bogus network (i.e. bogus VLAN ID)


mbelur (approved):
LGTM

PR #39, opened 2021-08-05
fix nodeName KeyError failures

mbelur:
Add additional checks to ensure that the
VM is Running and the host nodeName is available.

### DIFF
--- a/harvester_e2e_tests/utils.py
+++ b/harvester_e2e_tests/utils.py
@@ -252,7 +252,7 @@ def restart_vm(admin_session, harvester_api_endpoints, previous_uid, vm_name,
 def assert_vm_restarted(admin_session, harvester_api_endpoints,
                         previous_uid, vm_name, wait_timeout):
     # give it some time for the VM instance to restart
-    time.sleep(120)
+    time.sleep(180)
 
     def _check_vm_instance_restarted():
         resp = admin_session.get(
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/scenarios/test_vm_negatives.py
```diff
@@ -65,6 +66,24 @@ def test_host_shutdown(self, request, admin_session,
     @pytest.mark.host_management
     def test_delete_vm_while_host_shutdown(self, request, admin_session,
                                            harvester_api_endpoints, basic_vm):
+        def _check_vm_instance_nodename():
+            resp = admin_session.get(
+                harvester_api_endpoints.get_vm_instance % (
```
guangyee, line 71: suggest using utils.lookup_vm_instance(). Also make `vm_instance_josn` and `nonlocal` inside the `_check_vm_instance_nodename()` function so we don't need line 88-89 anymore.

harvester_e2e_tests/scenarios/test_vm_negatives.py
```diff
@@ -82,6 +101,24 @@ def test_delete_vm_while_host_shutdown(self, request, admin_session,
     @pytest.mark.host_management
     def test_vm_after_host_reboot(self, request, admin_session,
                                   harvester_api_endpoints, basic_vm):
+        def _check_vm_instance_nodename():
```
guangyee, line 104: maybe we can move this code to utils.py so they can be shared. Perhaps putting it into `lookup_vm_instance()` and add an `ensure_running` parameter?

guangyee (approved):
lgtm

PR #40, opened 2021-08-05
Fix host maintenance test

guangyee:
Use direct API endpoints from the template instead of relying on the links.
Also account for HTTP 409 on update by using polling.


mbelur (approved):
LGTM!

PR #41, opened 2021-08-12
Edit VM from YMAL to add Advanced Config

chajain:
Edit a VM from the form to add user data
Edit a VM from the form to add network data


harvester_e2e_tests/scenarios/test_create_vm.py
```diff
@@ -219,6 +220,51 @@ def test_create_update_vm_machine_type(request, admin_session, image, keypair,
...
+                                    user_data_with_guest_agent, network_data):
+    # make sure the VM instance is successfully created
+    vm_instance_json = utils.lookup_vm_instance(
+        admin_session, harvester_api_endpoints, single_vm)
+    vm_name = single_vm['metadata']['name']
+    previous_uid = vm_instance_json['metadata']['uid']
+    vol_data = single_vm['spec']['template']['spec']['volumes']
+    assert 'cloudInitNoCloud' not in vol_data
+    userdata = user_data_with_guest_agent.replace('\\n', '\n')
+    netdata = network_data.replace('\\n', '\n')
```
guangyee, line 235: I don't we need to do this as the fixture has already returning the correct format. See
https://github.com/harvester/tests/blob/main/harvester_e2e_tests/fixtures/vm.py#L44

harvester_e2e_tests/scenarios/test_create_vm.py
```diff
@@ -219,6 +220,51 @@ def test_create_update_vm_machine_type(request, admin_session, image, keypair,
                 utils.delete_vm(request, admin_session,
                                 harvester_api_endpoints, vm_json)
 
+
+def test_create_update_vm_user_data(request, admin_session, image, single_vm,
```
guangyee, line 224: I think we need to make the test with the `public_network` marker since we should verify that the updated VM can be SSH into with the updated user credential.

harvester_e2e_tests/scenarios/test_create_vm.py
```diff
@@ -219,6 +220,51 @@ def test_create_update_vm_machine_type(request, admin_session, image, keypair,
...
+    found = False
+    for i in updated_vol_data:
+        if i['name'] == 'cloudinitdisk':
+            if (i['cloudInitNoCloud']['networkData'] == netdata and
+                    i['cloudInitNoCloud']['userData'] == userdata):
+                found = True
+                break
+
+    assert found, 'Failed to update VM %s to add userDate & NetworkData' % (
+        vm_name)
```
guangyee, line 266: We should reboot the VM to ensure the changes works by successfully SSH into the VM.

PR #47, opened 2021-08-26
Adding terraform test for creating an image

chajain:
Added a test to create images using terraform commands


chajain:
Creating another PR

PR #49, opened 2021-09-01
Added Windows Image - Single VM and Multiple VM test

yarunachalam:


config.yml
```diff
@@ -27,3 +27,4 @@ rancher-endpoint: ~
 
 node-scripts-location: 'scripts/vagrant'
 
+win-image-url: 'http://10.84.144.252/meera/ws2016.qcow2'
```
mbelur, line 30: can we set the default value of win-image-url to None and in the test check if it's not configured as None, then run the test?

mbelur (changes requested):

config.yml
guangyee: We'll need a conditional market for this. I'll implement it in a follow-on patch.

guangyee (approved):
LGTM

config.yml
mbelur: ok sounds good.

mbelur (approved):

PR #52, opened 2021-09-08
Fix test breakages

guangyee:
Fix test breakages relating to:

1) API changes for the images where "sourceType" is required.
2) timing delaying for the volume state change when their owning
   VM is deleted
3) implement enable/disable maintenance as they are not longer part
   of action links


chajain (approved):
Looks good

mbelur (approved):
LGTM

PR #53, opened 2021-09-13
Test - Edit a VM and add install Enable usb tablet option

yarunachalam:
Test - Edit a VM and add install Enable usb tablet option


guangyee (approved):
looks good!

PR #55, opened 2021-09-15
Terraform apis/scenarios tests

chajain:
Added the terraform apis tests for images, volumes, keypair and network
Added scenario test for creating virtual machine using terraform


harvester_e2e_tests/templates/provider.tf.j2
```diff
@@ -0,0 +1,13 @@
+terraform {
+  required_version = ">= 0.13"
+  required_providers {
+    harvester = {
+      source  = "registry.terraform.io/harvester/harvester"
```
guangyee, line 5: to make it self-contained, shouldn't this be pointing to a local directory, relative to tests root?

guangyee (approved):
Very nice!

PR #59, opened 2021-09-30
Add Harvester Cluster Network Test

yarunachalam:
Add Harvester Cluster Network Test

### DIFF
--- a/harvester_e2e_tests/apis/test_networks.py
+++ b/harvester_e2e_tests/apis/test_networks.py
@@ -95,6 +95,7 @@ def test_create_edit_network(request, admin_session, harvester_api_endpoints,
 
 
 @pytest.mark.terraform
+@pytest.mark.public_network
 def test_create_network_using_terraform(request, admin_session,
                                         harvester_api_endpoints,
                                         network_using_terraform):
--- a/harvester_e2e_tests/fixtures/network.py
+++ b/harvester_e2e_tests/fixtures/network.py
@@ -36,12 +36,21 @@ def enable_vlan(request, admin_session, harvester_api_endpoints):
         vlan_json['config'] = {}
     if 'defaultPhysicalNIC' not in vlan_json['config']:
         vlan_json['config']['defaultPhysicalNIC'] = None
-    vlan_json['enable'] = True
-    vlan_json['config']['defaultPhysicalNIC'] = vlan_nic
-    utils.poll_for_update_resource(request, admin_session,
-                                   harvester_api_endpoints.update_vlan,
-                                   vlan_json,
-                                   harvester_api_endpoints.get_vlan)
+
+    if utils.is_marker_enabled(request, 'terraform'):
+        utils.create_clusternetworks_terraform(
+            request,
+            admin_session,
+            harvester_api_endpoints,
+            'resource_clusternetworks',
+            vlan_nic)
+    else:
+        vlan_json['enable'] = True
+        vlan_json['config']['defaultPhysicalNIC'] = vlan_nic
+        utils.poll_for_update_resource(request, admin_session,
+                                       harvester_api_endpoints.update_vlan,
+                                       vlan_json,
+                                       harvester_api_endpoints.get_vlan)
 
 
 def _cleanup_network(admin_session, harvester_api_endpoints, network_id,
@@ -164,4 +173,7 @@ def network_using_terraform(request, admin_session,
     yield network_json
 
     if not request.config.getoption('--do-not-cleanup'):
-        utils.destroy_resource(request, admin_session)
+        utils.destroy_resource(
+            request,
+            admin_session,
+            'harvester_network.' + network_json['metadata']['name'])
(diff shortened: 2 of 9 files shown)


harvester_e2e_tests/apis/test_networks.py
```diff
@@ -99,3 +100,31 @@ def test_create_network_using_terraform(request, admin_session,
...
+@pytest.mark.terraform
+def test_cluster_network_using_terraform(request, admin_session,
+                                         harvester_api_endpoints):
+
+    network_data = utils.create_clusternetworks_terraform(
+                       request,
+                       admin_session,
+                       harvester_api_endpoints,
+                       'resource_clusternetworks',
+                       'eth0')
```
guangyee, line 114: we shouldn't hardcode this to `eth0`, we should make it configurable and default to `harvester-mgmt` as that's the default bonded NIC name for the management network.

harvester_e2e_tests/apis/test_networks.py
```diff
@@ -99,3 +100,31 @@ def test_create_network_using_terraform(request, admin_session,
...
+    assert 'eth0' in network_data['config']['defaultPhysicalNIC']
+    if not request.config.getoption('--do-not-cleanup'):
+        utils.destroy_resource(request, admin_session, 'dir-only')
+
+    # undo cluster network update back to original
+    network_data = utils.create_clusternetworks_terraform(
+                        request, admin_session,
+                        harvester_api_endpoints,
+                        'resource_clusternetworks',
+                        'vlan')
```
guangyee, line 126: shouldn't this come from the `--vlan-nic` config?

harvester_e2e_tests/apis/test_networks.py
yarunachalam: fixed. this should come from --vlan-nic config

harvester_e2e_tests/apis/test_networks.py
yarunachalam: after cluster_network testing done. Undo cluster_network. Instead of hardcording vlan - I have get the existing vlan-nic value and update it back

yarunachalam: after sync up with @guangyee . We dont need this undo cluster network. 
Also added, enable vlan for terraform and also covered testcase for already existing public network.

harvester_e2e_tests/apis/test_networks.py
```diff
@@ -99,3 +100,23 @@ def test_create_network_using_terraform(request, admin_session,
                                         harvester_api_endpoints,
                                         network_using_terraform):
     pass
+
+
+@pytest.mark.terraform
+@pytest.mark.public_network
+def test_cluster_network_using_terraform(request, admin_session,
+                                         harvester_api_endpoints):
```
guangyee, line 108: all you have to do is to add the `network` fixture in the args as we are already creating network using terraform.

```
def test_cluster_network_using_terraform(request, admin_session,
                                                                   harvester_api_endpoints, network):

File: harvester_e2e_tests/apis/test_networks.py [hunk 4]
```diff
@@ -99,3 +100,23 @@ def test_create_network_using_terraform(request, admin_session,
...
+                       request,
+                       admin_session,
+                       harvester_api_endpoints,
+                       'resource_clusternetworks',
+                       vlan_nic)
+
+    time.sleep(30)
+    assert vlan_nic in network_data['config']['defaultPhysicalNIC']
+    if not request.config.getoption('--do-not-cleanup'):
+        utils.destroy_resource(request, admin_session, 'dir-only')
```
guangyee, line 122: no need for all that code as you are already creating networking in the fixture

harvester_e2e_tests/fixtures/network.py
```diff
@@ -36,12 +36,25 @@ def enable_vlan(request, admin_session, harvester_api_endpoints):
...
         vlan_json['config']['defaultPhysicalNIC'] = None
-    vlan_json['enable'] = True
-    vlan_json['config']['defaultPhysicalNIC'] = vlan_nic
-    utils.poll_for_update_resource(request, admin_session,
-                                   harvester_api_endpoints.update_vlan,
-                                   vlan_json,
-                                   harvester_api_endpoints.get_vlan)
+
+    for item in request.session.items:
+        if item.get_closest_marker('terraform') is not None:
```
guangyee, line 41: No need to do the for loop here. I would put the marker lookup logic into utils.py.

harvester_e2e_tests/fixtures/network.py
```diff
@@ -36,12 +36,25 @@ def enable_vlan(request, admin_session, harvester_api_endpoints):
...
+            utils.create_clusternetworks_terraform(
+                request,
+                admin_session,
+                harvester_api_endpoints,
+                'resource_clusternetworks',
+                vlan_nic)
+            if not request.config.getoption('--do-not-cleanup'):
```
guangyee, line 48: No need for cleanup.

harvester_e2e_tests/apis/test_networks.py
yarunachalam: I can remove the assert and --do-not-cleanup

harvester_e2e_tests/fixtures/network.py
yarunachalam: I can remove the assert and --do-not-cleanup

harvester_e2e_tests/fixtures/network.py
yarunachalam: fixed

PR #61, opened 2021-10-04
backout changes to conditionally install terraform provider

guangyee:
Lets always install the latest Harvester terraform providers for now.
Tox {posargs} is just a string replacement, which doesn't play well in
bash command if it contains a quote.


mbelur (approved):
LGTM

PR #62, opened 2021-10-04
Fix various tests breakages

guangyee:
1. "rancher-endpoint" config is no longer needed as Rancher auth test cases
   are no longer applicable since v0.3.0.
2. Fix the expectation for creating an image with no URL.
3. Make the default values in config.yml closely match what's in the
   vagrant environment.
4. Skip user tests are user management feature is deferred.
5. Fix host enable/disable maintenance mode.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -11,19 +11,16 @@ password: 'admin'
 do-not-cleanup: False
 
 # harvester_cluster_nodes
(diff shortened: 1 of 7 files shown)


chajain (approved):
LGTM

mbelur (approved):
LGTM

PR #64, opened 2021-10-07
Improve VLAN networking tests

guangyee:
1. For simplicity, let's use the 'default' namespace for all our tests
   unless the given test is namespace-specific.
2. Code around https://github.com/harvester/harvester/issues/1059 for now by
   rebooting the VM for the public IP to be accessible.


mbelur (approved):
LGTM

PR #65, opened 2021-10-08
Make the hosts tests more robust

guangyee:
By returning the host back to it's original state after the test with
better error handling. Also, when the host is shutdown, it's status
could be "unavailable" so we need to account for that.

### DIFF
--- a/harvester_e2e_tests/apis/test_hosts.py
+++ b/harvester_e2e_tests/apis/test_hosts.py
@@ -15,6 +15,7 @@
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
 
+import copy
 from harvester_e2e_tests import utils
 import pytest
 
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/apis/test_hosts.py
```diff
@@ -66,61 +67,80 @@ def test_update_first_node(request, admin_session, harvester_api_endpoints):
...
-            host_data_to_delete = data
+    host_data_to_delete = host_data['data'][0]
     # delete the host
     utils.delete_host(request, admin_session, harvester_api_endpoints,
                       host_data_to_delete)
     resp = admin_session.get(harvester_api_endpoints.list_nodes)
     assert resp.status_code == 200, 'Failed to list nodes: %s' % (resp.content)
     host_data = resp.json()
     assert host_data_to_delete not in host_data['data']
+    # TODO(gyee): do we need to make sure the VIP still works?
```
mbelur, line 143: we may want to update the configuration entry "harvester_cluster_nodes" in the config.yml after since the subsequent test runs fail the first test test_get_host fails on the assertion

mbelur (changes requested):

harvester_e2e_tests/apis/test_hosts.py
guangyee: good point! I think we may need to run this test separately, last in the pipeline, since it is so destructive.

PR #68, opened 2021-10-13
Delete the last host

guangyee:
For delete host test, delete the last one instead of the first one
to avoid destroying the VIP.

Also, turned off SSL certificate verification for all APIs as we are using
self-signed certificate out-of-the-box.


chajain (approved):
Looks good

PR #69, opened 2021-10-14
Installing terraform inside the terraform_test_artifacts folder

chajain:


terraform_test_artifacts/terraform_install.sh
```diff
@@ -0,0 +1,37 @@
...
+fi
+
+TERDIR="$PWD/terraform_test_artifacts"
+if [ -d "$TERDIR"/bin ]; then
+  echo "Removing bin folder"	
+  rm -rf ${TERDIR}/bin
+fi
+
+pushd "$TERDIR"
+`wget ${DOWNLOAD_URL}`
```
guangyee, line 29: wget ${DOWNLOAD_URL} -o terraform_bin.zip

terraform_test_artifacts/terraform_install.sh
```diff
@@ -0,0 +1,42 @@
...
+USAGE="${0}: [<version>] 
+Where:
+  <version>: version of to dowload. If absent, it will download the latest.
+"
+
+DOWNLOAD_URL=
+if [ $# -eq 0 ] ; then
+	# lookup the latest release download URL
+	DOWNLOAD_URL="https://releases.hashicorp.com/terraform/1.0.9/terraform_1.0.9_linux_amd64.zip"
+	VER="1.0.9"
```
guangyee, line 14: Instead of hardcoding the version, you can get it using github API. For example:

`curl -s https://api.github.com/repos/hashicorp/terraform/releases/latest | grep '"name":' | awk '{print $2}' | sed 's/.*v//' | sed 's/".*//'`

terraform_test_artifacts/terraform_install.sh
```diff
@@ -0,0 +1,42 @@
...
+elif [ $# -eq 1 ] ; then
+	VER=$1
+	DOWNLOAD_URL="https://releases.hashicorp.com/terraform/${VER}/terraform_${VER}_linux_amd64.zip"
+else
+        echo "$USAGE"
```
guangyee, line 19: We should keep the spacing consistent.

guangyee (commented):
Just a couple more suggestion. Otherwise, the code looks good. I've also tested the PR locally.

PR #77, opened 2021-10-20
updated workflow for manual tests

noahgildersleeve:
This will update the git workflow to update the Github Pages branch with up to date docs when they are changed.


noahgildersleeve:
I moved the workflow for GH-pages to another workflow yaml so that it would only run if the docs directory changes.

PR #85, opened 2021-10-20
Use a naming convention when creating a VLAN network

guangyee:
As all network must have a unique VLAN ID, we want to name it such that
we won't run into a situation where tests are creating different networks
with the same VLAN ID and failing.


noahgildersleeve:
What happens if you use the utility method to create a network with a duplicate VLAN. Does it fail?

PR #91, opened 2021-10-28
Include markers for test cases based on priority

yarunachalam:
- Include markers for test cases based on priority.
- Include comment with test case number and description

### DIFF
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -27,12 +27,17 @@ pytest_plugins = [
 ]
 
 
+@pytest.mark.images_p1
+@pytest.mark.p1
 def test_list_images(admin_session, harvester_api_endpoints):
     resp = admin_session.get(harvester_api_endpoints.list_images)
     assert resp.status_code == 200, (
         'Failed to list images: %s' % (resp.content))
 
 
+@pytest.mark.images_p2
+@pytest.mark.p2
+# This negative test covers images-3-Create Image no data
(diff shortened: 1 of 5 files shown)


harvester_e2e_tests/apis/test_images.py
```diff
@@ -27,12 +27,17 @@
...
+@pytest.mark.p1
 def test_list_images(admin_session, harvester_api_endpoints):
     resp = admin_session.get(harvester_api_endpoints.list_images)
     assert resp.status_code == 200, (
         'Failed to list images: %s' % (resp.content))
 
 
+@pytest.mark.images_p2
+@pytest.mark.p2
+# This negative test covers images-3-Create Image no data
```
guangyee, line 40: Lets use python docstring format for these. https://www.python.org/dev/peps/pep-0257/#toc-entry-6

For example,

```
def test_create_image_no_data(admin_session, harvester_api_endpoints):
"""
Test creating images with no data.

Covers:
* Images-3: Create Image no data
"""
```

harvester_e2e_tests/apis/test_images.py
```diff
@@ -27,12 +27,17 @@
 ]
 
 
+@pytest.mark.images_p1
+@pytest.mark.p1
 def test_list_images(admin_session, harvester_api_endpoints):
```
guangyee, line 32: Add test case cover docstring here.

guangyee (changes requested):
The markers looks good. As for the test cases description, lets use Python docstring format.

yarunachalam:
created PR 94 with all above comments

PR #92, opened 2021-11-02
Use harvester provider directly from the registry

guangyee:
No need to install it explicitly as it is now in the registry.

https://registry.terraform.io/providers/harvester/harvester/latest


mbelur (approved):
LGTM

PR #93, opened 2021-11-05
fix terraform tests

mbelur:
Tests that would call import would fail with
no terraform.tfstate file.
So this change moves the terraform init to
be performed earlier so it's applicable for
both the import and apply commands.

Also simplify the script to handle
the "import" argument

Also fixed the variable used for
the destroy all resources.

Also cleanup dev.tfrc which is no longer needed


guangyee (approved):
awesome, thanks for fixing it!

PR #99, opened 2021-11-11
Fix terraform tests for vlan network

mbelur:
These tests handle a situation where a nettwork with the vlan ID is already allocated and we need handle this scenario in the test by doing a lookup first and if found, import the resource and not try to create it as creation fails with VLAN ID allocated already message.


guangyee (approved):
lgtm

chajain (approved):
Looks good

PR #107, opened 2021-11-16
change the default admin password based on the new limitation for a m…

mbelur:
…in of 12 chars


guangyee (approved):
Thanks for fixing it!

PR #111, opened 2021-11-18
add .nojekyll for github pages

lanfon72:
adding `.nojekyll` in the root, which is used to bypass routing path starts with underscore `_`.

ref: https://github.blog/2009-12-29-bypassing-jekyll-on-github-pages/

### DIFF
--- a/.github/workflows/gh-pages-push-workflow.yml
+++ b/.github/workflows/gh-pages-push-workflow.yml
@@ -46,6 +46,7 @@ jobs:
           git rm -r --ignore-unmatch .
           popd > /dev/null
           mv docs/public/* "$gh_pages_worktree"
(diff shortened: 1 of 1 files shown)


guangbochen:
do we have any files or directory start with underscores?

lanfon72:
@guangbochen yes, the new folder starts with `_` for ordering.

guangbochen:
can u share the link or take a image of the error, just want to double-check after the fix, thanks

lanfon72:
for currently, the `https://harvesterhci.io/tests/manual/_incoming/` is not working, it will work in local build because there is no limit for URL path starts with `_`.

noahgildersleeve (approved):
LGTM

PR #112, opened 2021-11-18
reset admin password

mbelur:


harvester_e2e_tests/fixtures/session.py
```diff
@@ -23,6 +23,10 @@
 def admin_session(request, harvester_api_endpoints):
     username = request.config.getoption('--username')
     password = request.config.getoption('--password')
+    old_password = request.config.getoption('--old-password')
+
+    # reset admin password
+    reset_admin_password(harvester_api_endpoints, old_password, password)
```
chajain, line 29: Do we need a first_login check here as the authentication fails when you try to run the tests the second time?

chajain (changes requested):

harvester_e2e_tests/fixtures/session.py
mbelur: Will take a look if the first_login check API is available

mbelur: I agree that this is not a good experience to modify the configuration file on a subsequent run. For now, we will leave it as-is as the default "admin" password fine from a API perspective.

PR #115, opened 2021-11-23
backup tests using different vm fixture

mbelur:
Using vms_with_vlan_as_default_network fixture
for backup tests instead of vm_with_one_vlan
to avoid adding the restart_vm workaround due
to bug#1059


guangyee (approved):
lgtm

PR #120, opened 2021-12-02
Add airgapped, update rancher integration and 1.0 feature test case

TachunLin:
Add and update harvester rancher integration test add the following

- Add new airgapped test scenario in case 54, 55, 56
- Update airgapped test scenario in case 37, 38
- Update 02 integrate rancher in harvester settings
- Update enable harvester node driver

Add new test case for v1.0.0 feature for 
- #1218  Missing http proxy settings on rke2 and rancher pod
- #1401 Support volume hot-unplug 
- #1330 Enhance the import experience to Rancher


lanfon72 (commented):
I think the `.hugo_build.lock` is not necessary?

For #1218, should we separate them into different test cases?
from the issue comments, I think it just asked to test in different verify point.

TachunLin:
> I think the `.hugo_build.lock` is not necessary?
> 
> For #1218, should we separate them into different test cases? from the issue comments, I think it just asked to test in different verify point.

Thanks for the check, I have just updated and commit the fix according to it including: 

1. Remove .hugo_build.lock file
2. Combine #1218 to one test case under _incomfing
3. Fix #1401 title and issue link under _incomfing
4. Add a title for test case 47 in harvester-rancher folder

lanfon72 (approved):
LGTM

PR #122, opened 2021-12-09
Make Rancher image version configurable

guangyee:
This patch also imporves session handling by retrying on the common errors.


guangyee:
Combined with https://github.com/harvester/tests/pull/123

PR #124, opened 2021-12-17
Improve Rancher integration test

guangyee:
1. Refactor the Rancher integration tests into their own file.
2. Require external Rancher instance. Running a Rancher VM in Harvester is
   reliable, especially in networking. As such, the Rancher VM setup code is
   being removed.
3. Retry on 500 on all HTTP methods.


mbelur (approved):
LGTM

PR #129, opened 2022-01-04
use --rancher-cluster-wait-timeout instead of --wait-timeout

guangyee:
For cluster creation and deletion as Rancher RKE cluster usually take
longer to create and remove. This will prevent unnecessary timeouts which
can lead to dangling clusters in Rancher.


chajain (approved):
Looks good

PR #130, opened 2022-01-06
add UI tests into `ui_tests`

lanfon72:
subtree added from https://github.com/lanfon72/ui-tests
keep git history for tracking purpose.


lanfon72:
use another branch ( see #131 ) to avoid redundant commit history.

PR #131, opened 2022-01-06
add UI tests into `ui_tests`

lanfon72:
subtree added from https://github.com/lanfon72/ui-tests
keep git history for tracking purpose.


noahgildersleeve (approved):
I think there is some work to do on changing the architecture of some of the code, but let's get this merged in so that we can get the changes in the main repo.

TachunLin (approved):
- Can Install required python packages based on requirement.txt on python 3.9.6
- Can trigger chrome install by webdrivermanager
- Tested all test suites inside the `atests` folder

Let's merge it to the main and work on it.

PR #147, opened 2022-01-20
Remove restart VM workaround dual DHCP NICs

guangyee:
Dual DHCP NICs tests are de-prioritized as it is not a normal use case.
The expectation is that the default route will be setup for the first DHCP
NIC only. Subsequent DHCP routes must be manually added to the VM. Therefore,
this patch removes the restart VM workaround, as well as tests that are no
longer relevant.


noahgildersleeve (approved):
LGTM

PR #157, opened 2022-02-09
added in pydoc3 to requirements and possibly fixed the GH Actions pus…

noahgildersleeve:
…h workflow

### DIFF
--- a/cypress/utils/utils.ts
+++ b/cypress/utils/utils.ts
@@ -17,7 +17,7 @@ export const validateZip = (filename: string) => {
         console.log(zipEntry.toString()); // outputs zip entries information
         // if (zipEntry.entryName == "my_file.txt") {
         //     console.log(zipEntry.getData().toString("utf8"));
-        // }
+        // } 
     });
 }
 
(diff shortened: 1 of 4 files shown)


cypress/utils/utils.ts
```diff
@@ -17,7 +17,7 @@ export const validateZip = (filename: string) => {
         console.log(zipEntry.toString()); // outputs zip entries information
         // if (zipEntry.entryName == "my_file.txt") {
         //     console.log(zipEntry.getData().toString("utf8"));
-        // }
+        // }
```
yasker, line 20: extra space?

noahgildersleeve: That was to trigger the rebuild. I might actually add the CI workflows as watched files to avoid having to do this. Currently we only trigger GH pages builds if we edit any of the following
- cypress/
- harvester_e2e_tests/
- docs/

That means if we're updating the GH actions it won't trigger a deploy

noahgildersleeve: I pushed an update to the triggers so we won't have to do that again

yasker: OK. I think we need to formalize the directory and test name a bit later, e.g. it should be called `frontend e2e tests` and `backend e2e tests` instead of `cypress` vs `harvester_e2e_tests` IMO. But we can do that later.

yasker (approved):

PR #158, opened 2022-02-10
added in makefile for local creation of doc sites and updated README.md

noahgildersleeve:


README.md
```diff
@@ -316,3 +316,23 @@ The site will be accessible at http://localhost:1313.
 [terraform]: https://www.terraform.io/
 [AWS S3]: https://aws.amazon.com/s3/
 [Harvester manual test cases]: https://github.com/harvester/tests/tree/main/docs/content/manual
+
```
yasker, line 319: Should we replace the Preview Website section above?

noahgildersleeve: Good point. I'll update that.

noahgildersleeve: Updated

PR #159, opened 2022-02-11
added test skeletons and updated typedoc.json to only check the test …

noahgildersleeve:
…specs

updated typedoc to include all new tests. Also removed erroneous pageobject


TachunLin (approved):
LTGM.

PR #163, opened 2022-02-14
Cypress: Add vm create case

n313893254:


cypress/pageobjects/components/component.po.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  isDisabled(): Cypress.Chainable<boolean> {
+    return this.self().invoke('attr', 'disabled').then((disabled: string) => disabled === 'disabled');
+  }
+
+  checkVisible(): Cypress.Chainable<boolean> {
+    return this.self().should('be.visible');
+  }
+}
```
noahgildersleeve, line 33: You can access any cypress chainable in the code by just chaining off of a `cy.get` selector. so if you want to check if something is disabled you can just do `cy.get(this.button).should('be.disabled');` You can do the same thing with checkVisible `cy.get(this.button).should('be.visible')`

cypress/pageobjects/virtualmachine.po.ts
```diff
@@ -0,0 +1,68 @@
...
+    cy.intercept('GET', '/v1/harvester/harvesterhci.io.virtualmachineimages').as('loadVMCreate');
+    cy.wait('@loadVMCreate')
+  }
+
+  public setDefaultValue() {
+    this.namespace().select('default')
+    this.name().input('test-vm-name-automation')
+    this.description().input('test-vm-description-automation')
+    this.cpu().input('2')
+    this.memory().input('4')
```
noahgildersleeve, line 26: These should probably be accepted as parameters from calling the method in the test spec not statically set.

cypress/integration/po.types.ts
```diff
@@ -0,0 +1 @@
+export type CypressChainable = Cypress.Chainable;
```
noahgildersleeve, line 1: We should have the CypressChainable type already included in the @types section of the package.json for the cypress project.

noahgildersleeve (changes requested):

cypress/pageobjects/components/component.po.ts
noahgildersleeve: I just saw that's what you had done in the methods. I'm resolving this comment.

cypress/integration/po.types.ts
noahgildersleeve: I now see why you added in this. Resolving comment.

noahgildersleeve (approved):
LGTM

PR #166, opened 2022-02-16
Add Cypress test case

n313893254:
- [create-a-vm-with-start-vm-on-creation-checked](https://harvesterhci.io/tests/manual/virtual-machines/create-a-vm-with-start-vm-on-creation-checked/
)
- [add-a-network-to-an-existing-vm-with-only-1-network](https://harvester.github.io/tests/manual/virtual-machines/add-a-network-to-an-existing-vm-with-only-1-network/)


noahgildersleeve (approved):
LGTM

PR #173, opened 2022-02-23
Add test case restore-backup-create-new-vm-in-another-namespace

FrankYang0529:
related issue: https://github.com/harvester/harvester/issues/1027


noahgildersleeve (approved):
LGTM

PR #208, opened 2022-03-07
Add test cases for extra disks

weihanglo:
- extra-disk-using-raw-disk

Related issues:
- https://github.com/harvester/tests/issues/178
- https://github.com/harvester/harvester/issues/1839
- https://github.com/harvester/harvester/issues/1430
- https://github.com/harvester/node-disk-manager/pull/26


noahgildersleeve (changes requested):
Could you please move the test into the `docs/content/manual/_incoming/` directory? We're trying to keep all new tests for the release in that directory so that we can keep track of new features and bugfixes per release.

noahgildersleeve (approved):
LGTM

PR #211, opened 2022-03-07
Add manual test: large-amount-extra-disks

weihanglo:
related: 
- https://github.com/harvester/node-disk-manager/pull/32
- https://github.com/harvester/harvester/issues/1718


noahgildersleeve (changes requested):
Could you please move the test into the `docs/content/manual/_incoming/` directory? We're trying to keep all new tests for the release in that directory so that we can keep track of new features and bugfixes per release.

noahgildersleeve (approved):
LGTM

PR #220, opened 2022-03-08
New manual test: 1874-extra-disk-swap-path

weihanglo:
Related: harvester/harvester#1874


noahgildersleeve (changes requested):
Could you please move the test into the `docs/content/manual/_incoming/` directory? We're trying  to keep all new tests for the release in that directory so that we can keep track of new features and bugfixes per release.

noahgildersleeve (approved):
LGTM

PR #251, opened 2022-03-24
Cypress: Fix ComponentPo bug

n313893254:
It will cause `new LabeledInputPo('.labeled-input')` doesn't work as expect.


noahgildersleeve (approved):
LGTM

PR #259, opened 2022-03-31
Fix terraform import clusternetwork

futuretea:
After terraform-provider-harvester v0.2.0+, we don't need to add namespace during importing clusternetwork


lanfon72 (approved):
LGTM

PR #260, opened 2022-04-01
doc: Update Code Comment For API Reference

irishgordo:
- updates code comment on api_endpoints to reference more recent API
docs

Resolves: docs/adjust-swagger-api-doc-location-reference

### DIFF
--- a/harvester_e2e_tests/fixtures/api_endpoints.py
+++ b/harvester_e2e_tests/fixtures/api_endpoints.py
@@ -25,7 +25,7 @@ pytest_plugins = [
 
 
 # NOTE: see Swagger API doc here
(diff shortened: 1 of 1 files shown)


irishgordo:
I'm hopeful that the:
```
https://docs.harvesterhci.io/latest/reference/api/
```
will always just reference the latest.
I tried it out, currently seems to redirect to:
```
https://docs.harvesterhci.io/v1.0/reference/api/
```
Not sure entirely if that's something that works :sweat_smile:

noahgildersleeve (approved):
LGTM

irishgordo:
Wooooo!

PR #266, opened 2022-04-06
Add tests for #2023 Legacy ISO

tlehman:
- Add test cases for [issue 2023](https://github.com/harvester/harvester/issues/2023) (one for legacy BIOS and one for UEFI)
image

### DIFF
--- a/docs/content/manual/_incoming/2023-boot-installer-legacy-and-uefi.md
+++ b/docs/content/manual/_incoming/2023-boot-installer-legacy-and-uefi.md
@@ -1,0 +1,16 @@
+---
+title: Boot installer under Legacy BIOS and UEFI
+---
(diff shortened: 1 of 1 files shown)


tlehman:
Update: the boot failure under BIOS was from commit [69eeff..](https://github.com/tlehman/harvester-installer/commit/60eeffee6575bbadfbd27512daedd324b6c5afd1) on my branch where I was trying to fix a grub issue for a user.

I tried the same qemu BIOS boot test on commit [77493b...](https://github.com/harvester/harvester-installer/commit/77493b0952556bbae2769ab11e8ecbef89b2225e) (current origin master) and the BIOS boot test passed.

So the failure I found this morning is because of my proposed grub2.cfg changes. Currently **this test does not reproduce the exact issue in issue 2023**. It's still worth having, but a repro for 2023 will require an HP DL360G6 or G8

PR #283, opened 2022-04-13
Fix typo for Harvester

tlehman:
# Testing

I tested this by `cd`ing into `docs/` and running `hugo server`, then verifying that the site generates correctly, and that those 4 pages are listed and render correctly.


noahgildersleeve (approved):
LGTM

PR #337, opened 2022-05-11
Cypress: Run specific in order

DaiYuzeng:
We intend to run specific in order via the `testFiles` of Cypress's configuration, but the specific which is missing `describe`(only has `it`) will run at first.

截屏2022-05-11 下午5 29 28

We will Add missing `describe` to cover `it`.

@noahgildersleeve @lanfon72 @TachunLin Please let me know if you have concerns, thanks!

### DIFF
--- a/cypress/cypress.json
+++ b/cypress/cypress.json
@@ -4,7 +4,17 @@
     "fixturesFolder": "./fixtures",
     "pluginsFile": "./plugins/index.js",
     "supportFile": "./support/index.js",
-    "testFiles": "**/*.spec.ts",
+    "testFiles": [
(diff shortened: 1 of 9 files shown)


WuJun2016 (approved):
File Structure  LGTM.

Some of the tests in there that run with errors need to be solved by other pr.

Note: To ensure that all test cases run in the order we specify, all `it` needs to be wrapped in `describe`,  Otherwise, some strange problem will occur and the execution will not follow the specified order.

@noahgildersleeve     cc @lanfon72  @TachunLin

n313893254 (approved):
LGTM

PR #345, opened 2022-05-18
Add script for test report and Add command of stopping spec's test on GUI

DaiYuzeng:
There are two new features:

- Generate report
- Stop failed test on GUI via new command `stopOnFailed`

### Generate report

`yarn run report` can generate report now and the report will be saved as `mochawesome-report/mochawesome.html`

The content of HTML looks like:
image

There are records about Test status/Error message/Spent times of test.

### Stop test

Dependent tests should be stopped if it's failed.
Use code below to stop it.

```
before(() => {
    cy.stopOnFailed();
})
```

### DIFF
--- a/cypress/package.json
+++ b/cypress/package.json
@@ -26,7 +26,7 @@
   },
   "scripts": {
     "delete:reports": "rm cypress/results/* || true",
-    "prereport": "yarn run delete:reports",
+    "prereport": "npm run delete:reports",
     "report": "node scripts/run.js"
   }
 }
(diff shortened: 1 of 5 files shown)


WuJun2016:
@DaiYuzeng   For a description of how to use `cy.stopFailed` please add it to `cypress/README.md`  （You can wait until the pr is merged before adding  https://github.com/harvester/tests/pull/343）, thanks

cypress/package.json
```diff
@@ -20,5 +23,10 @@
   },
   "engines": {
     "node": "^12.22 || ^14.13 || >=16"
+  },
+  "scripts": {
+    "delete:reports": "rm cypress/results/* || true",
+    "prereport": "yarn run delete:reports",
```
lanfon72, line 29: `yarn` is not required by default, we can use `npm run delete:reports` instead.

DaiYuzeng: thanks!

WuJun2016 (approved):
LGTM

PR #346, opened 2022-05-18
* [CY] update test files in order

lanfon72:
Changes:
- Create and move test cases into relevant folders
- Update `import` to use non-relative import [^1]
- Update **testFiles** in `cypress.json` in a specific order

Folder name will be mapped into relevant labels, the dependencies as below:
- VM requires:
    - image < volume < dashboard (host general functions)
    - network < settings
- backup and Vm settings (SSH Key, applying Templates and Cloud Config Templates) require:
    - VM

The execution order will be:
image

To notice that **SSH Key** and **Cloud Config Template** are special case that we will not only test that they could be created but also test that they will be applied on VM.

[^1]: https://www.typescriptlang.org/docs/handbook/module-resolution.html#relative-vs-non-relative-module-imports


WuJun2016:
@lanfon72  login （first login） should be executed first, otherwise if it is a new environment, there will be problems，

lanfon72:
@WuJun2016 yeah, login should always be the first test case. (so I move it out of folder)

cypress/cypress.json
```diff
@@ -6,15 +6,16 @@
...
-      "login.spec.ts",
-      "settings.spec.ts",
-      "images.spec.ts",
-      "hosts.spec.ts",
-      "virtualmachines/*.ts",
-      "cloud-config-templates.spec.ts",
-      "ssh-keys.spec.ts",
-      "network.spec.ts",
-      "support.spec.ts"
+      "login.spec.ts"
```
WuJun2016, line 9: `"login.spec.ts"`   =>  `"login.spec.ts",`

WuJun2016 (approved):
LGTM

PR #352, opened 2022-05-24
To fix daily e2e testing on pytest

lanfon72:
Changes:
- To handle `polling2` exception correctly
- To check VM is running before get the hosted node name

this is based on **harvester-install-and-test-e2e-daily#253**

---
Cases of `polling2` exception:
- [x] test_create_vm.py::test_create_windows_vm::setup
- [x] test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke1_cluster::setup
- [x] test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke2_cluster::setup
- [x] test_create_vm.py::test_host_maintenance_mode
- [x] test_vm_networking.py::TestVMNetworking::test_add_vlan_network_to_vm

Cases of `KeyError 'nodeName'`:
- [x] test_vm_negatives.py::TestHostDown::test_vm_after_host_reboot
- [x] test_vm_negatives.py::TestHostDown::test_delete_vm_while_host_shutdown

Case of AssertionError:
- [x] test_vm_networking.py::test_create_vm_using_terraform
- [ ] test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_add_prj_owner_user

### DIFF
--- a/harvester_e2e_tests/fixtures/network.py
+++ b/harvester_e2e_tests/fixtures/network.py
@@ -70,11 +70,14 @@ def _cleanup_network(admin_session, harvester_api_endpoints, network_id,
 
     # NOTE(gyee): there's no way we know how many VMs the network is currently
     # attached to. Will need to keep trying till all the VMs had been deleted
-    success = polling2.poll(
-        _delete_network,
-        step=5,
-        timeout=wait_timeout)
-    assert success, 'Unable to cleanup network: %s' % (network_id)
+    try:
+        polling2.poll(
+            _delete_network,
+            step=5,
+            timeout=wait_timeout)
+    except polling2.TimeoutException as e:
+        errmsg = 'Unable to cleanup network: %s' % (network_id)
+        raise AssertionError(errmsg) from e
 
 
 def _lookup_network(request, admin_session, harvester_api_endpoints, vlan_id):
--- a/harvester_e2e_tests/scenarios/test_create_vm.py
+++ b/harvester_e2e_tests/scenarios/test_create_vm.py
@@ -709,12 +709,14 @@ def test_host_maintenance_mode(request, admin_session, image, keypair,
                     return True
             return False
 
-        success = polling2.poll(
-            _check_vm_instance_migrated,
-            step=5,
-            timeout=request.config.getoption('--wait-timeout'))
-        assert success, 'Timed out while waiting for VM to be migrated: %s' % (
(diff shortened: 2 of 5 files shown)


lanfon72:
For those two AssertionErrors, I will need more time to investigate, especially terraform one.

irishgordo:
I'll look at this more in the morning - it's definitely gonna be great if daily e2e's start passing again!

I will mention that for like the `test_create_rke1_cluster` and `rke2` that rely on the fixtures - the fact that they have to download the image that will be used for those clusters from somewhere - I know that if we were to run those tests right now against an air-gapped harvester & rancher, they would fail because the test wouldn't be able to download the image.  Like all the image.py fixtures would fail in an air-gapped environment:
https://github.com/harvester/tests/blob/main/harvester_e2e_tests/fixtures/image.py

I think in the future when we implement more rancher_integration tests into either CI or our nightly runs on something that is airgapped we could add additional logic to handle things if it's running in an airgapped environment.

lanfon72:
> I think in the future when we implement more rancher_integration tests into either CI or our nightly runs on something that is airgapped we could add additional logic to handle things if it's running in an airgapped environment.

The CI is not air-gapped environment, but we can still improve the pytest's e2e test cases for air-gapped free.

TachunLin (approved):
Tested the following test scenarios on local kvm machines.  
Some cases not yet able to verified since need to settle down local test environment like host management script and terraform. 

As per discussed, I think we can merge it and check whether the failure have been resolved on daily E2E test. WDYT. 

- [x] **`PASS`** test_create_vm.py::test_create_windows_vm::setup
  image

- [ ] **`PASS`** test_vm_networking.py::TestVMNetworking::test_add_vlan_network_to_vm
  image

- [ ] **`FAILED`** test_create_vm.py::test_host_maintenance_mode
  image

- [ ] **`PASS`** test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke1_cluster::setup
- [ ]  **`FAILED`** test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke2_cluster::setup

  image
  image
  image

* Found something strange is when RKE2 cluster provisioing stuck in the join boostrap, 
The E2E job seems not able to finish the result. 
  image

lanfon72:
This PR not really fix those root cause but the error message, we would need fix this first to let the error message more clearly to triage.

PR #353, opened 2022-05-25
[CY] add create vlan1 test

WuJun2016:


TachunLin (changes requested):
Checked on the fresh vagrant pxe install harvester node
* Without network enabled 
image

* Running the create vlan test will cause failure since network interface not ready
image

* If enabled network, the create vlan can run successfully
image

Not sure whether this case already have the pre-req test to enable network 
If not, I would suggest can have a check on this to prevent failure.

WuJun2016:
> Checked on the fresh vagrant pxe install harvester node
> 
> * Without network enabled
>   image
> * Running the create vlan test will cause failure since network interface not ready
>   image
> * If enabled network, the create vlan can run successfully
>   image
> 
> Not sure whether this case already have the pre-req test to enable network If not, I would suggest can have a check on this to prevent failure.

The network is already enabled in the setting page,
https://github.com/harvester/tests/blob/8a5065823ec887e567b172c0e630ea0b19b6fa66/cypress/testcases/settings/settings.spec.ts#L64
The reason you are running the error is that I forgot to remove the it.only

TachunLin (approved):
LGTM. To run the network test suite, user must ensure already execute the `configure global vlan` test have already executed on settings.spec 

image

PR #354, opened 2022-05-26
regression: UI

lanfon72:
This PR will cover all of UI regression test cases.
Changes:
- exclude all files under `cypress/cypress`, which automatically generated by cypress.
- explicitly check the header after `cy.login()`
- add new task `readYaml` for checking yaml content.
- add new tasks `deleteFolder` and `deleteFile`.
- remove partially implemented **generate support bundle** and doc comments
- implement test cases:
  1. Links of Community Section
  2. Download KubeConfig File
  3. Generate Support Bundle (by the limitation, it will only verify form's behavior)

Limitation:
- Unable to catch the generated support bundle file[^1], the example[^2] is not work for our dashboard.
- `context` be treated as `describe`, so there is no scope for **sub-tests**, every `it` will trigger `before/beforeEach/after/afterEach`. (thus, merge sub test cases of links into single test case)

[^1]: https://github.com/cypress-io/cypress/issues/8619
[^2]: https://github.com/cypress-io/cypress-example-recipes/blob/master/examples/testing-dom__download/cypress/integration/form-submission-spec.js


lanfon72:
I should update the `Generate Support Bundle` test case to be more clearly, please don't merge it before new commits.

TachunLin (approved):
Verified the new tasks and test cases running correctly
Also did not show error message after test complete 

image

noahgildersleeve:
@lanfon72 I added in a test case for downloading the support bundle. I think I got it working without a timeout triggering. I found a solution that involved running an event listener [here](https://github.com/cypress-io/cypress/issues/14857#issuecomment-790765826). I added in comments in the code. Take a look when you get a chance and let me know if you have any questions or concerns. The solution is a little fragile, but I think it should probably work. The issue is that it has three overlapping timeouts that it balances between.

cypress/testcases/dashboard/support.spec.ts
```diff
@@ -31,9 +31,8 @@ describe("Support Page", () => {
       const kubeconfig = `${Cypress.config("downloadsFolder")}/local.yaml`
       page.downloadKubeConfigBtn.click()
 
-      // cy.readFile(kubeconfig)
-      //   .should("exist")
-      cy.verifyDownload('local.yaml', {timeout: constants.timeout.downloadTimeout});
```
lanfon72: I tend to use `.readFile`, it will print file's content to log. (so we can track it if need)

cypress/testcases/dashboard/support.spec.ts
```diff
@@ -63,32 +62,62 @@ describe("Support Page", () => {
...
+          cy.log("Total Files:", files)
+          return {dirs, files}
+        })
+        // .then(({dirs, files}) => {}) // new verfiers here
 
-          page.inputSupportBundle('this is a test bundle')
-            .get("@generateBtn").click()
-            // Verify that the download has completed
-            .verifyDownload('supportbundle', {timeout: constants.timeout.downloadTimeout, contains: true})
-            .intercept('/v1/harvester/supportbundles*/download', (req) => {
```
lanfon72: The API not always be called, so I will not check it.

irishgordo:
@lanfon72 - I'm wondering if this is something we can also scaffold on the backend for our tests.
I started to dig into it a bit here:
https://github.com/irishgordo/tests/commit/a6ac4be4d702b0bdef4a03ada99cd319b969197c

I know for: https://github.com/harvester/harvester/pull/2321 - the introduction of "fleet-agent-*" and "fleet-controller-*" as directories inside of "logs", that could be something we could check for since I believe the feature will be backported.

lanfon72:
@irishgordo yeah it is good if API's test can cover it too.
We have only few test cases about support bundle, I think the PR is new test case that we need to cover.

PR #360, opened 2022-05-31
To fix daily e2e #260

lanfon72:
Changes:
- skip know issue on `scenarios/test_vm_negatives.py::test_delete_vm_while_host_shutdown`
- remove unnecessary exception chaining
- add more context information when error raised
- fix potential bugs
    1. `test_delete_host` should not test on single node, and the **mark** should be `trylast`, not `last`
    2. parser's **type** should use constructor instead of pure string literal

Case of Errors/Fails on daily\#260:
~~- [ ] `test_create_vm.py::test_create_windows_vm` unable to test on local environment, it use specific image on Jenkins~~
~~- [ ]  `test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke2_cluster` **fail** on local environment, still investigating~~
- [x] `test_vm_networking.py::test_create_vm_using_terraform` **pass** on local environment after update image version, it should be the root cause.
- [x] `test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_create_rke1_cluster` **pass** on local environment, add more context information for debug
- [x] `test_create_vm.py::test_host_maintenance_mode` **pass** on local environment
- [x] `test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_add_prj_owner_user` **pass** on local environment, should be the environment issue
- [x] `test_vm_negatives.py::TestHostDown::test_delete_vm_while_host_shutdown` **fail** on local environment, known issue
- [x] `test_vm_negatives.py::TestHostDown::test_vm_after_host_reboot` **pass** on local environment, should be the environment issue

### DIFF
--- a/harvester_e2e_tests/scenarios/test_create_vm.py
+++ b/harvester_e2e_tests/scenarios/test_create_vm.py
@@ -716,7 +716,7 @@ def test_host_maintenance_mode(request, admin_session, image, keypair,
                 timeout=request.config.getoption('--wait-timeout'))
         except polling2.TimeoutException as e:
             errfmt = 'Timed out while waiting for VM to be migrated: %s'
-            raise AssertionError(errfmt % vm_json['metadata']['name']) from e
+            raise AssertionError(errfmt % vm_json['metadata']['name'])
 
         vmi_json_after_migrate = utils.lookup_vm_instance(
             admin_session, harvester_api_endpoints, vm_json)
--- a/harvester_e2e_tests/scenarios/test_create_vm.py
+++ b/harvester_e2e_tests/scenarios/test_create_vm.py
@@ -714,7 +714,7 @@ def test_host_maintenance_mode(request, admin_session, image, keypair,
                 _check_vm_instance_migrated,
                 step=5,
                 timeout=request.config.getoption('--wait-timeout'))
-        except polling2.TimeoutException as e:
+        except polling2.TimeoutException:
             errfmt = 'Timed out while waiting for VM to be migrated: %s'
             raise AssertionError(errfmt % vm_json['metadata']['name'])
 
--- a/harvester_e2e_tests/scenarios/test_rancher_integration.py
+++ b/harvester_e2e_tests/scenarios/test_rancher_integration.py
@@ -116,10 +116,13 @@ def _delete_rke1_cluster(request, rancher_admin_session, rancher_api_endpoints,
 
 def _wait_for_cluster_ready(request, rancher_admin_session,
                             rancher_api_endpoints, cluster_name):
+    cluster_json = dict()
+
     def _wait_for_cluster_ready_status():
         resp = rancher_admin_session.get(
             rancher_api_endpoints.get_provisioning_cluster % (cluster_name))
         if resp.status_code == 200:
+            nonlocal cluster_json
             cluster_json = resp.json()
             if ('status' in cluster_json and
                     'ready' in cluster_json['status']):
@@ -130,9 +133,19 @@ def _wait_for_cluster_ready(request, rancher_admin_session,
             _wait_for_cluster_ready_status,
             step=5,
             timeout=request.config.getoption('--rancher-cluster-wait-timeout'))
-    except polling2.TimeoutException as e:
-        errmsg = 'Timed out while waiting to import Harvester.'
-        raise AssertionError(errmsg) from e
+    except polling2.TimeoutException:
+        from datetime import datetime
+        from json import dumps
+        fmt = "%Y-%m-%dT%H:%M:%SZ"
+
+        def _key(item):
+            t = item['lastUpdateTime']
+            return datetime.strptime(t, fmt) if t else datetime(2000, 1, 1)
+
+        items = sorted(cluster_json['status']['conditions'], key=_key)
+        errmsg = ('Timed out while waiting to import Harvester.\n'
+                  f'Last 5 conditions: {dumps(items[-5:])}')
+        raise AssertionError(errmsg)
 
 
 def _import_harvester_cluster_into_rancher(request, admin_session,
--- a/harvester_e2e_tests/scenarios/test_rancher_integration.py
+++ b/harvester_e2e_tests/scenarios/test_rancher_integration.py
@@ -365,12 +365,16 @@ def rke1_cluster(request, rancher_admin_session, rancher_api_endpoints, image,
 
 
 @pytest.fixture(scope='class')
-def rke2_cluster(request, rancher_admin_session, rancher_api_endpoints, image,
+def rke2_cluster(request, admin_session, harvester_api_endpoints,
+                 rancher_admin_session, rancher_api_endpoints, image,
                  network, cloud_credential,
                  external_rancher_with_imported_harvester):
     test_environment = request.config.getoption('--test-environment')
     cluster_id = external_rancher_with_imported_harvester
     cloud_credential_id = cloud_credential
+
+    utils.assert_image_ready(request, admin_session, harvester_api_endpoints,
+                             image['metadata']['name'])
 
     # Harvester Kubeconfig
     rke2_cluster_name = 'rke2-' + test_environment
--- a/harvester_e2e_tests/utils.py
+++ b/harvester_e2e_tests/utils.py
@@ -362,7 +362,7 @@ def delete_image_by_name(request, admin_session,
             timeout=request.config.getoption('--wait-timeout'))
     except polling2.TimeoutException as e:
         errmsg = 'Timed out while waiting for image to be deleted'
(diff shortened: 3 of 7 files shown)


harvester_e2e_tests/utils.py
```diff
@@ -460,9 +462,10 @@ def _check_vm_ready():
             _check_vm_ready,
             step=5,
             timeout=request.config.getoption('--wait-timeout'))
-    except polling2.TimeoutException as e:
-        errmsg = 'Timed out while waiting for VM to be ready.'
-        raise AssertionError(errmsg) from e
+    except polling2.TimeoutException:
+        errmsg = ('Timed out while waiting for VM to be ready.\n'
+                  f"Stucking in Phase {resp_json['status']['phase']}")
```
irishgordo, line 467: We don't necessarily need to change this :smile: - this is more of a general comment:

We're leveraging the value of `resp_json.status.phase` from the Python dictionary, we may encounter a `KeyError` if the value doesn't exist which could throw another error while trying to generate `errmsg`.
[we could use the get() functionality on the dict](https://docs.python.org/3/library/stdtypes.html#dict.get)
We may want to try to just set it as something like:
```
f"Sticking in Phase {resp_json.get('status',{}).get('phase', 'default_phase_not_found')}"
```
Where we could create a temporary dict if we can't find the dictionary associated with 'status' and then if we can't find 'phase' to use within the `errmsg` we can just use some default string of some sort.

harvester_e2e_tests/scenarios/test_rancher_integration.py
```diff
@@ -130,9 +133,19 @@ def _wait_for_cluster_ready_status():
...
+    except polling2.TimeoutException:
+        from datetime import datetime
+        from json import dumps
+        fmt = "%Y-%m-%dT%H:%M:%SZ"
+
+        def _key(item):
+            t = item['lastUpdateTime']
+            return datetime.strptime(t, fmt) if t else datetime(2000, 1, 1)
+
+        items = sorted(cluster_json['status']['conditions'], key=_key)
```
irishgordo, line 145: We don't necessarily need to change this :smile: - this is more of a general comment:

There is the chance within the `cluster_json` dict we may also run into the possibility of it not having either 'status' or 'conditions'.  If that happened it would throw a `KeyError` within the assignment of `items` - we could possibly specify defaults like I mentioned on the other comment below.

https://docs.python.org/3/library/stdtypes.html#dict.get

harvester_e2e_tests/scenarios/test_create_vm.py
```diff
@@ -714,9 +714,9 @@ def _check_vm_instance_migrated():
                 _check_vm_instance_migrated,
                 step=5,
                 timeout=request.config.getoption('--wait-timeout'))
-        except polling2.TimeoutException as e:
+        except polling2.TimeoutException:
             errfmt = 'Timed out while waiting for VM to be migrated: %s'
-            raise AssertionError(errfmt % vm_json['metadata']['name']) from e
+            raise AssertionError(errfmt % vm_json['metadata']['name'])
```
irishgordo, line 719: We don't necessarily need to change this :smile: - this is more of a general comment:

I just noticed on the dict of `vm_json` we may also experience a `KeyError` while trying to build the `AssertionError` message.  Like the few other comments below.

irishgordo (changes requested):
This looks good!  Thank you for all the effort in working to fix the daily e2e tests!  I did leave some comments, we may not necessarily need to make changes but I definitely am curious about when we create error messages how we may approach injecting variables into those messages.

harvester_e2e_tests/utils.py
lanfon72: the key will never missed, or it would be a critical issue for the cluster.
you can reference `Line:703` of inner function for more details.

harvester_e2e_tests/scenarios/test_rancher_integration.py
lanfon72: The key will never missed, or it would be a critical issue for the cluster.
If you have more concern, you can test it on the API endpoint `https://<your_rancher_URL>/v1/provisioning.cattle.io.clusters/fleet-default/<The_cluster_ID>`.

By the case, it will apply to RKE1 and RKE2 cluster which we attempt to create.

lanfon72:
added `utils.assert_image_ready` which will be used in `test_rancher_integration.rke2_cluster`.

To fix the case:
image

We should make sure the image has been downloaded successfully before create RKE2 cluster.
All of current test cases related to **image** didn't detect the potential error.

But I think this is not the root cause of the fail, I still investigating it.

irishgordo:
@lanfon72 is that error bubbling up from the image not being able to download from opensuse.org?
I noticed with:
https://github.com/lanfon72/tests/blob/88312c0dbf264956bde4e95ca148b9110e1f0551/harvester_e2e_tests/utils.py#L1184-L1193
we could pass in a `--image-cache-url`. 
I wrote a small simplehttp server that grabs the files and will serve things locally with Python:
```
import os
import requests
import http.server
import socketserver

GLOBAL_PORT_FOR_HTTP_SERVER = 8800
GLOBAL_REMOTE_URL_FOR_LEAP = "https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.3/images/openSUSE-Leap-15.3.x86_64-NoCloud.qcow2"
GLOBAL_REMOTE_URL_FOR_TUMBLEWEED = "https://download.opensuse.org/tumbleweed/iso/openSUSE-Tumbleweed-NET-x86_64-Current.iso"
DIRECTORY_GIVEN = None

class ServerHandler(http.server.SimpleHTTPRequestHandler):
    def __init__(self, *args, **kwargs):
        super().__init__(*args, directory=DIRECTORY_GIVEN, **kwargs)


def check_directory(directory_provided):
    is_existing_directory = os.path.isdir(directory_provided)
    if not is_existing_directory:
        print("the directory provided does not exist...")
        return False 
    else:
        if os.listdir(directory_provided):
            print("the directory is not empty... assuming files are already present...")
            return False

def download_file_to_local_directory(url_to_download, directory_local_with_file_name_locally):
    print(f"downloading... {url_to_download} ...")
    data = requests.get(url_to_download)
    with open(directory_local_with_file_name_locally, 'wb') as file:
        file.write(data.content)
    print(f"finished downloading... {url_to_download} ...")
    return True

def start_server(directory_provided):
    global DIRECTORY_GIVEN
    DIRECTORY_GIVEN = directory_provided
    print('starting the server...')
    with socketserver.TCPServer(("", GLOBAL_PORT_FOR_HTTP_SERVER), ServerHandler) as httpd:
        print(f"serving the server at port: {GLOBAL_PORT_FOR_HTTP_SERVER} - with directory: {DIRECTORY_GIVEN}...")
        httpd.serve_forever()

def main():
    print('..setting up :) ...')
    directory_provided = input('Please provide the absolute path to an empty directory to use...\n')
    result_of_check = check_directory(directory_provided)
    if result_of_check == False:
        start_server(directory_provided)
    else: 
        leap_download_result = download_file_to_local_directory(GLOBAL_REMOTE_URL_FOR_LEAP, os.path.join(directory_provided, 'openSUSE-Leap-15.3.x86_64-NoCloud.qcow2'))
        tumbleweed_download_result = download_file_to_local_directory(GLOBAL_REMOTE_URL_FOR_TUMBLEWEED, os.path.join(directory_provided, 'openSUSE-Tumbleweed-NET-x86_64-Current.iso'))
        if tumbleweed_download_result != True or leap_download_result != True:
            print("...files didn't download... exiting")
            exit()
        start_server(directory_provided)

if __name__ == "__main__":
    print('executing temp_download_script.py directly...')
    main()
```
Then if you re-ran the tests with "--image-cache-url" and passed in for instance: `http://localhost:8800", it should try to pull both OpenSUSE Leap and OpenSUSE Tumbleweed from their.
Based around what I see here:
https://github.com/harvester/tests/commit/d3e3ffe01e30e7bf3e7dee39326652fe9ad238e3

lanfon72:
@irishgordo yes, we can use the cache sever for image, but we still need to fail fast in rancher integration, that's the reason I commit https://github.com/harvester/tests/pull/360/commits/88312c0dbf264956bde4e95ca148b9110e1f0551 for the fix.

lanfon72:
By several tests, `test_create_rke2_cluster` still failed even I wait more than 3 hours.
But If I use UI to create RKE2 (with same configuration, but extend disk to **30Gi**), RKE2 can be provisioned successfully.

I will test the API directly, then fix `test_create_vm_using_terraform` next week.

lanfon72:
Let's make this PR merged to get new fails :smiling_face_with_tear:

irishgordo (approved):

PR #372, opened 2022-06-13
[DevOps] feat: Supporting 2096 Work With CI Building Test Dockerfile To Run Tests Against Harvester-Installer CI

irishgordo:
* Edits to support additional Makefile contexts
* dockerfile for ci e2e that holds tox testing content

Resolves: feat/2096-Supporting-E2E-CI-Work-For-Harvester-Installer


irishgordo:
Relates to: 
https://github.com/harvester/harvester/issues/2096

irishgordo:
The idea is that this would just allow us to build a dockerfile that would build a docker image that would contain all the python & pip dependencies to run tests from the image directly as another option in comparison to setting up a virtualenv via something like `python3 -m venv the_venv_directory`

irishgordo:
If you wanted to test for instance that the dockerfile is preforming as expected, you could run an arbitrary test on the container like:
```
docker run -t -i --rm harvester_installer_e2e_ci_image:latest pytest harvester_e2e_tests/scenarios/test_rancher_integration.py::TestRancherIntegrationWithExternalRancher::test_import_harvester_into_rancher -vv --endpoint https://192.168.0.131 --password testtesttest --rancher-endpoint https://helm-install.test --rancher-admin-password testtesttest
```

irishgordo:
Additionally, this could be tested with something like this (you'd want to replace the values) but:
```
docker run -t -i --rm harvester_installer_e2e_ci_image:latest pytest -k 'not verify_host_maintenance_mode and not host_mgmt_maintenance_mode and not host_reboot_maintenance_mode and not host_poweroff_state and not create_images_using_terraform and not create_keypairs_using_terraform and not create_edit_network and not create_network_using_terraform and not create_volume_using_terraform' --junitxml=result_harvester_api_latest.xml -m 'not delete_host' harvester_e2e_tests/apis --username ${baseSettingsYaml.harvester_dashboard.admin_user} --password ${baseSettingsYaml.harvester_dashboard.admin_password} --endpoint https://${baseSettingsYaml.harvester_network_config.vip.ip} --harvester_cluster_nodes ${baseSettingsYaml.harvester_cluster_nodes}"
```

Any feedback, suggestions, etc. - super open to iterate and make any adjustments :smile:

PR #379, opened 2022-06-15
Created Basic Support Bundle Checks

irishgordo:
- adding basic support_bundle api test functionality
- per PEP8, it's acceptable to have line length up to 99 characters,
  excluding docstrings, increased flake8 tox.ini configuration of
  increased max line length

Resolves: 350-collect-fleet-logs-in-support-bundle

### DIFF
--- a/harvester_e2e_tests/apis/test_support_bundle.py
+++ b/harvester_e2e_tests/apis/test_support_bundle.py
@@ -1,0 +1,142 @@
+# Copyright (c) 2022 SUSE LLC
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of version 3 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.   See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, contact SUSE LLC.
+#
+# To contact SUSE about this file by physical or electronic mail,
+# you may find current contact information at www.suse.com
+
+import json
+import os
+import re
+import tempfile
+import zipfile
+
+import polling2
+import pytest
+from harvester_e2e_tests import utils
+
+pytest_plugins = [
+    'harvester_e2e_tests.fixtures.api_endpoints',
+    'harvester_e2e_tests.fixtures.user',
+    'harvester_e2e_tests.fixtures.session',
+    'harvester_e2e_tests.fixtures.support_bundle'
+]
+
+
+def _create_bundle(admin_session, harvester_api_endpoints):
+    """Helper function to create 'happy path' support bundle
+
+    Args:
+        admin_session (dict): fixture that gets exteneded from requests
+        harvester_api_endpoints (dict): constants harvester api endpoints
+    """
+    request_json = utils.get_json_object_from_template('support_bundle')
+    resp = admin_session.post(
+        harvester_api_endpoints.create_support_bundle, json=request_json)
+    assert resp.status_code == 201
+    resp_json = json.loads(resp.text)
+    name = resp_json.get('metadata').get('name')
+    return name, resp_json
+
+
+def test_create_missing_description(
+        admin_session, harvester_api_endpoints):
+    """Tests that description is enforced on support bundle creation
+
+    Args:
+        admin_session (dict): fixture that gets exteneded from requests
+        harvester_api_endpoints (dict): constants harvester api endpoints
+    """
+    request_json = utils.get_json_object_from_template('support_bundle')
+    del request_json['spec']['description']
+    resp = admin_session.post(
+        harvester_api_endpoints.create_support_bundle, json=request_json)
+    assert resp.status_code == 422
+    assert 'is invalid: spec.description: Required value' in resp.text
+
+
+def test_delete(admin_session, harvester_api_endpoints):
+    """Tests creating a support bundle then it's deletion
+
+    Args:
+        admin_session (dict): fixture that gets exteneded from requests
+        harvester_api_endpoints (dict): constants harvester api endpoints
+    """
+    name, _resp_json = _create_bundle(admin_session, harvester_api_endpoints)
+    delete_resp = admin_session.delete(
+        harvester_api_endpoints.delete_support_bundle + name
+    )
+    assert delete_resp.status_code == 200
+
+
+@pytest.mark.p1
+@pytest.mark.slow
+def test_creates_successful_bundle_and_downloads(
+        admin_session, harvester_api_endpoints):
+    """Creates a support bundle, validates it progresses to 100%,
+    downloads in memory as a zip file, audits the zip file to
+    ensure all needed files are present
+
+    Args:
+        admin_session (dict): fixture that gets exteneded from requests
+        harvester_api_endpoints (dict): constants harvester api endpoints
+    """
+    name, _resp_json = _create_bundle(admin_session, harvester_api_endpoints)
+    progress = 0
+
+    def _wait_for_progress_to_be_ready():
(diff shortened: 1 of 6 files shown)


harvester_e2e_tests/apis/test_support_bundle.py
```diff
@@ -0,0 +1,109 @@
...
+        '/download')
+    with open(zip_file_location, 'wb') as support_zip_file:
+        support_zip_file.write(resp_download.content)
+    list_of_files_in_support_bundle = None
+    with zipfile.ZipFile(zip_file_location, 'r') as support_zip_file:
+        list_of_files_in_support_bundle = support_zip_file.namelist()
+    cattle_fleet_local_system_fleet_agent_log = False
+    cattle_fleet_system_fleet_controller_log = False
+    cattle_fleet_system_gitjob_log = False
+    for file in list_of_files_in_support_bundle:
```
lanfon72, line 97: In here, I would suggest to use regex.match:
```python
patterns = [r"logs/cattle-fleet-local-system/fleet-agent-.*/fleet-agent.log",
            r"logs/cattle-fleet-system/fleet-controller-.*/fleet-controller.log",
            r"logs/cattle-fleet-system/gitjob-.*/gitjob.log"]
# or the simple one: patterns: [r".*cattle-fleet-system.*fleet-agent.log", ...]
# I would recommend that one file use one pattern, it would be more easy to update.
matches = []
for f in files:
    for pattern in patterns:
        matches.extend([f] if re.match(pattern, f) else [])

errmsg  = ("Some file(s) not found,"
           f"files: {matches}\n"
           f"patterns: {patterns}")
assert len(matches) == len(patterns), errmsg
```

And we can just add new pattern if there have any new file need to be verified.

tox.ini
```diff
@@ -25,5 +25,8 @@ deps = flake8
 commands =
   flake8
 
+[flake8]
+per-file-ignores = harvester_e2e_tests/apis/test_support_bundle.py: E501
```
lanfon72, line 29: In my opinion, we should not exclude single file for _E501_, it would break the window then other files would like to be the one.
(In most cases, you would still have ways to fit _E501_)

lanfon72 (changes requested):

harvester_e2e_tests/apis/test_support_bundle.py
```diff
@@ -0,0 +1,109 @@
...
+        resp_progress_check = admin_session.get(
+            harvester_api_endpoints.view_support_bundle +
+            name_to_use_for_download)
+        json_progress_result = json.loads(resp_progress_check.text)
+        progress = json_progress_result.get('status', {}).get('progress', 0)
+        if progress != 100:
+            return False
+        return True
+
+    successful_support_bundle_generation = polling2.poll(
```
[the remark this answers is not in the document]
lanfon72: this is the minimum change requirement

harvester_e2e_tests/apis/test_support_bundle.py
```diff
@@ -0,0 +1,109 @@
...
+        _wait_for_progress_to_be_ready,
+        step=5,
+        timeout=300
+    )
+    assert successful_support_bundle_generation, ('Timed out waiting for ' +
```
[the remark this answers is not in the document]
irishgordo: the addition of the `+` was to make the pep8 linter happy, to create a wrap because of [E501, for flake8/pep8/autopep linting](https://www.flake8rules.com/rules/E501.html) - I can refactor to remove that :smile:

harvester_e2e_tests/apis/test_support_bundle.py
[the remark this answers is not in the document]
irishgordo: ahh, good catch, I can implement a try-except block, and raise an error with a useful message - if we do timeout on the operation :smile:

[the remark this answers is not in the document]
irishgordo: I agree, that refactor of the variable naming makes sense to me :+1: :smile:

harvester_e2e_tests/apis/test_support_bundle.py
```diff
@@ -0,0 +1,109 @@
...
+    zip_file_location = os.path.join(tempdir, 'support_bundle.zip')
+    resp_download = admin_session.get(
+        harvester_api_endpoints.download_support_bundle +
+        name_to_use_for_download +
+        '/download')
+    with open(zip_file_location, 'wb') as support_zip_file:
+        support_zip_file.write(resp_download.content)
+    list_of_files_in_support_bundle = None
+    with zipfile.ZipFile(zip_file_location, 'r') as support_zip_file:
+        list_of_files_in_support_bundle = support_zip_file.namelist()
```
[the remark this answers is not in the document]
irishgordo: agree :smile: - that naming makes more sense

harvester_e2e_tests/apis/test_support_bundle.py
irishgordo: I'll investigate the implementation of the regex :smile: 

I do like how that it's fairly easy to read through what is taking place without the regex implementation.  

We do end up with two nested for loops if we loop through every file and every pattern.  But I don't see that being a problem :smile: 

You're right that the extensibility with the `patterns` - should support bundle in the future implement additional things, we'd want to be able to extend this test.

tox.ini
irishgordo: I'll look to see what we can refactor to shrink the line length. If we're open to it, I'd think that we could expand the limit for the 'maximum' line length to something like 120 characters.  But after I make a few changes, we may be able to remove the exclusion because the file will meet all E501 requirements :smile:

harvester_e2e_tests/apis/test_support_bundle.py
lanfon72: the extensibility is one of the reason to use regex, another reason is, consider we are verifying multiple files, and there is no sub-test functionality in pytest, so in this case, we are unable to catch multiple files missed.

irishgordo: :smile:  - definitely :+1: 

I did realize we needed to adjust the regex a little bit.
After some testing, for our 'gitjob' regex for instance we would need to adjust it from:
```
r"logs/cattle-fleet-system/gitjob-.*/gitjob.log"
```
to:
```
r"^.*/logs/cattle-fleet-system/gitjob-.*/gitjob.log"
```
in order to match/capture the test string of:
```
supportbundle_87739721-c996-40f9-9317-3ab213cb3fc1_2022-06-22T18-29-15Z/logs/cattle-fleet-system/gitjob-cc9948fd7-6t7q2/gitjob.log
```
since `zip_file.namelist()` prints out the full path, top level directory included.

tox.ini
irishgordo: I'm thinking even after some cleanup all lines are less than 90 characters:
```
pep8 run-test: commands[0] | flake8
./harvester_e2e_tests/apis/test_support_bundle.py:129:80: E501 line too long (85 > 79 characters)
./harvester_e2e_tests/apis/test_support_bundle.py:130:80: E501 line too long (89 > 79 characters)
```
Instead of:
```
per-file-ignores = harvester_e2e_tests/apis/test_support_bundle.py: E501
```
I'll be pushing a change that tells Flake8 we allow up to 99 - I'm willing to change that though if we feel that's too much.  I've added that configuration in the `tox.ini`.

Though, via [Flake8 Rules](https://www.flake8rules.com/rules/E501.html) it seems that a maximum length of 99 characters is excepted.

From https://peps.python.org/pep-0008/#maximum-line-length :
> Some teams strongly prefer a longer line length. For code maintained exclusively or primarily by a team that can reach agreement on this issue, it is okay to increase the line length limit up to 99 characters, provided that comments and docstrings are still wrapped at 72 characters.

In this case it's not the 'docstrings' that are longer, it's our RegEx Pattern's length that is causing the `tox -e pep8` issues.

lanfon72: I'm up for expand to 99.

lanfon72 (approved):
LGTM

PR #396, opened 2022-06-29
add test case for 1418

chrisho:
related issues:

- https://github.com/harvester/harvester/issues/1418


TachunLin (changes requested):
In step 1, Perhaps we can add more details steps about how and where to add topology lables in Harvester nodes. 
While in step 3, we also need more details about where to check the topology labels on guest-cluster node

chrisho:
> 

up

> In step 1, Perhaps we can add more details steps about how and where to add topology lables in Harvester nodes. While in step 3, we also need more details about where to check the topology labels on guest-cluster node

updated

TachunLin (approved):
Thanks for the update. LGTM.

PR #412, opened 2022-07-19
Fix daily testing errors

lanfon72:
Changes:
- For `test_vm_after_host_reboot` case, change to wait VM before get the vm instance information, it should resolve the `nodeName` missing sometimes
- for `test_create_vm_using_terraform` case, change to skip network deletion when it is deleted via terraform

### the case of `test_create_vm_using_terraform`
image

According the error message above, it is not caused by the test case itself, but the teardown execution order.
When you look into to the [test_vm_networking.py](https://github.com/harvester/tests/blob/main/harvester_e2e_tests/scenarios/test_vm_networking.py), the first test case is `TestVMNetworking`, and its `test_ssh_to_vm` will trigger VLAN creation (via the argument `vms_with_vlan_as_default_network`).
The fixture of [network](https://github.com/harvester/tests/blob/f35236f80f18efcfb0f3c542440e29d468bba462/harvester_e2e_tests/fixtures/network.py#L118) be marked as `session` scope, it will be teardown after all test cases executed.
so Test case execution order will be:
```
# test_vm_networking.py
-> TestVMnetworking
=> Setup fixtures including network
-> execute test cases
=> Teardown fixutures which scope is not *session*
...
-> execute test_create_vm_using_terraform
=> destroy all resources including vlan network
=< Teardown fixtures which scope is *session*
```

It is the reason why I add the checker to skip `_cleanup_network` execution.
(I think we should fix it in a graceful way, but consider other solutions might affect more test cases, this fix should be the minimal affection.)

### DIFF
--- a/harvester_e2e_tests/fixtures/network.py
+++ b/harvester_e2e_tests/fixtures/network.py
@@ -127,9 +127,13 @@ def network(request, admin_session, harvester_api_endpoints, enable_vlan):
     yield network_data
 
     if not request.config.getoption('--do-not-cleanup'):
-        _cleanup_network(admin_session, harvester_api_endpoints,
-                         network_data['id'],
-                         request.config.getoption('--wait-timeout'))
+        # XXX: we would need to check the network not be deleted terraform yet
+        if not utils.is_marker_enabled(request, 'terraform') and \
+            _lookup_network(request, admin_session, harvester_api_endpoints,
+                            vlan_id) is not None:
+            _cleanup_network(admin_session, harvester_api_endpoints,
+                             network_data['id'],
+                             request.config.getoption('--wait-timeout'))
 
 
 @pytest.fixture(scope='class')
--- a/harvester_e2e_tests/scenarios/test_vm_negatives.py
+++ b/harvester_e2e_tests/scenarios/test_vm_negatives.py
@@ -100,10 +100,10 @@ class TestHostDown:
     def test_vm_after_host_reboot(self, request, admin_session,
                                   harvester_api_endpoints, basic_vm):
         # make sure the VM instance is successfully created and running
-        vm_instance_json = utils.lookup_vm_instance(
-            admin_session, harvester_api_endpoints, basic_vm)
         utils.assert_vm_ready(request, admin_session, harvester_api_endpoints,
                               basic_vm["metadata"]["name"], True)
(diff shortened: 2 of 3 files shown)


harvester_e2e_tests/fixtures/network.py
```diff
@@ -127,9 +127,13 @@ def network(request, admin_session, harvester_api_endpoints, enable_vlan):
...
 
     if not request.config.getoption('--do-not-cleanup'):
-        _cleanup_network(admin_session, harvester_api_endpoints,
-                         network_data['id'],
-                         request.config.getoption('--wait-timeout'))
+        # XXX: we would need to check the network not be deleted terraform yet
+        if not utils.is_marker_enabled(request, 'terraform') and \
+            _lookup_network(request, admin_session, harvester_api_endpoints,
+                            vlan_id) is not None:
+            _cleanup_network(admin_session, harvester_api_endpoints,
```
irishgordo, line 134: Ahh, this is smart, to check to make sure the Network exists before we attempt to clean it up!
If so, should we indent this method call to be within the nested result of the `_lookup_network() is not None:` conditional?
Something like:
```
        if not utils.is_marker_enabled(request, 'terraform') and \
            _lookup_network(request, admin_session, harvester_api_endpoints,
                            vlan_id) is not None:
                _cleanup_network(admin_session, harvester_api_endpoints,
                                 network_data['id'],
                                 request.config.getoption('--wait-timeout'))
```
But maybe it's just a flake8 formatting thing, I'm not sure.

irishgordo (changes requested):

harvester_e2e_tests/fixtures/network.py
lanfon72: No, we don't need to make another indent for the `_cleanup_network...`.
when you look into Line 131-133, it means:
```python
if not utils.is_marker_enabled(request, 'terraform') and _lookup_network(request, admin_session, harvester_api_endpoints, vlan_id) is not None:
    _cleanup_network(....)
```
In the end of Line 131, we use backslash(`\`) as the line continuation character, but we don't need to another backslach(`\`) for Line 132, because it is wrapped in parentheses.

irishgordo: Ahh you're right I completely missed the `\` 😅

irishgordo (approved):

PR #430, opened 2022-08-02
Update Manual Test For Standard User

irishgordo:
* standard user should not be able to do anything initially, updating
  manual test to cover that aspect

Resolves: doc/update-standard-user-no-access-test


lanfon72 (approved):
LGTM

PR #439, opened 2022-08-11
fix daily e2e errors

lanfon72:
### Changes
- [x] Update `api_endpoints.json.j2` to fix test cases in **harvester_e2e_tests/apis/test_vm_templates.py**
- [x] Fix test scenario: most available CPU
    - Update `basic_vm.json.j2` to inject **CPU** correctly
    - Update assert **available_cpu** to `1`, because we will expect that the created VM will initial with `1` CPU
    - Implement `parse_unit` to handle resource size
    - Update `lookup_hosts_with_most_available_cpu` to use **Reserved** resource instead of **Used**
- [x] Fix test scenario: most available Memory
    - Update assert **available_memory** to `1`, because we will expect that the created VM will initial with `1` Memory
    - Implement `format_unit` to handle resource size convert
    - Update `lookup_hosts_with_most_available_memory` to use **Reserved** resource instead of **Used**, to avoid over-commit, also reduce maximum memory to _99%_ before floor the size.


lanfon72:
I had tested that the `overcommit-config` will not affect the API of VM creation, so I don't need to update overcommit-config to _100%_ before execute those test scenarios.

irishgordo:
Screenshot from 2022-08-15 16-58-56
Screenshot from 2022-08-15 16-58-39
Screenshot from 2022-08-15 16-23-25

Working to try to test these edits on the bare-metal - noticing a few issues with "unsupported protocol scheme" on the images.

Also was thinking we might see if we can update the pytest parameter for the cloud image for ubuntu:
https://github.com/harvester/tests/blob/main/harvester_e2e_tests/scenarios/test_create_vm.py#L84

From `http:` to `https:` testing that as the test_create_vm.py runs

lanfon72:
I think it is another issue, you can use `-k "vm_on_available"` to test relevant test case only.

irishgordo:
@lanfon72 ahhh okay :+1:  - I'll pass in that test flag and see the test session results.
I without the flag, it did happen to get a bit further I believe than last time:
```
harvester_e2e_tests/scenarios/test_create_vm.py ...sEE..........^CERROR: got KeyboardInterrupt signal
_________________________________________________________________________ summary _________________________________________________________________________
ERROR:   py38: keyboardinterrupt

```
before I interrupted it

lanfon72:
You can use verbose (**-vvv**) parameter and the console will display more details, if _image-cache-url_ is not configured, it will spend time to download images every time.

irishgordo:
I was able to validate that:
- test_create_vm_on_available_cpu_node
- test_update_vm_on_available_cpu_node
- test_create_vm_on_available_memory_node
- test_update_vm_on_available_memory_node
- test_create_vm_on_available_cpu_and_memory_nodes

Pass.

Screenshot from 2022-08-15 18-30-50

:smile:

irishgordo:
I'm going to try to let all the other tests run and pass in extra verbosity.
Testing Env right now is:
v1.0.3 & v2.6.7-rc6

irishgordo:
@lanfon72 I did manage to re-run the test.

[local-test-run.log](https://github.com/harvester/tests/files/9347890/local-test-run.log)
 
1 failed, 15 passed, 1 skipped, 2 errors in 6486.22s (1:48:06)

Surrounding:
test_create_multiple_vms (had two errors)
test_host_maintenance_mode (was the failed test)

We can ignore `test_host_maintenance_mode` since I only have a single node running for the test - if I remember, a single node can't be put into maintenance mode.

It looked like the errors with test_create_multiple_vms were due to timeouts -but that may have been due to the fact it was waiting for the image to become available (ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img) and it was never downloaded due to:
```
 Head "ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img": unsupported protocol scheme "" 
```

I'm not sure if the `create_image` function might have some difficulties with the ubuntu image.
The polling timeout just failed at looking for the VM to become ready:
```
>           polling2.poll(
                _check_vm_ready,
                step=5,

```
OpenSUSE Leap images had no issues downloading, only Ubuntu.

lanfon72:
@irishgordo this PR will not change the function of `utils.create_image`, I think it is a existing issue and we could fix it in another PR.

As you tested, this PR will fix following test cases:
- test_create_vm_on_available_cpu_node
- test_update_vm_on_available_cpu_node
- test_create_vm_on_available_memory_node
- test_update_vm_on_available_memory_node
- harvester_e2e_tests/apis/test_vm_templates.py::test_create_verify_vm_template

For the `test_vm_templates.py`, it will fix the error of
image

For other 4 cases, the old implementation was not fit our criteria.
For test cases about CPU, created VM will not assign CPU amounts, and the function of calculation (`lookup_hosts_with_most_available_cpu`) fetch the `Used` resources, but we should use `Reserved` resources; and the `update_on_available` should assert the available CPU counts most `>1` because before we update the VM resource, it will be created with `1` CPU.
For test cases about Memory, similar to CPU's cases, this PR fix the function of calculation to make it hit our test criteria.

irishgordo (approved):

PR #487, opened 2022-08-31
[WIP][DRAFT] Adding Support Core Plugin

irishgordo:
* this adds the support-core plugin to the infrastructure scripting

Resolves: feat/support-core-plugin-introduction


irishgordo:
We should determine if we would like to try to pin versions to the plugins.
Historically, it looks like this hasn't taken place. 

```
jenkins-plugin-cli --plugins support-core:1206.v14049fa_b_d860
```
But that would mean shifting from the `jenkins install-plugin` over to the jenkinsci tool plugin installation manager:
https://github.com/jenkinsci/plugin-installation-manager-tool

irishgordo:
This was a manual change and has been implemented.

PR #579, opened 2022-10-26
added multi-nic update test

noahgildersleeve:
fixed formatting

### DIFF
--- a/docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
+++ b/docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
@@ -1,0 +1,23 @@
+---
+title: Upgrade Harvester on node that has bonded NICs for management interface
+---
+
+* Related issues: [#3045](https://github.com/harvester/harvester/issues/3045) [BUG] Harvester Upgrade 1.0.3 to 1.1.0 does not handle multiple SLAVE in BOND for management interface
+
+## Category: 
+* Upgrade
+
+## Environment Setup
+
+- This is to be done on a Harvester cluster where the NICs were configured to be bonded on install for the management interface
+
+## Verification Steps
+
+1. Install Harvester with at least 2 bonded NICs on install in `balance-tlb`
+
+    ![image](https://user-images.githubusercontent.com/83787952/198138366-472d4432-839a-4d9d-a942-d55fe8f6f6d8.png)
+1. Start an upgrade if the responder has one setup or set up a custom upgrade via the instructions [here](https://docs.harvesterhci.io/v1.1/upgrade/automatic/#prepare-an-air-gapped-upgrade)
+
+## Expected Results
+
+- The upgrade should complete successfully and the Harvester cluster should pass sanity checks


docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,23 @@
...
+---
+
+* Related issues: [#3045](https://github.com/harvester/harvester/issues/3045) [BUG] Harvester Upgrade 1.0.3 to 1.1.0 does not handle multiple SLAVE in BOND for management interface
+
+## Category: 
+* Upgrade
+
+## Environment Setup
+
+- This is to be done on a Harvester cluster where the NICs were configured to be bonded on install for the management interface
```
khushboo-rancher, line 12: It will be good if we add other env details like how many node etc.

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,23 @@
...
+
+## Verification Steps
+
+1. Install Harvester with at least 2 bonded NICs on install in `balance-tlb`
+
+    ![image](https://user-images.githubusercontent.com/83787952/198138366-472d4432-839a-4d9d-a942-d55fe8f6f6d8.png)
+1. Start an upgrade if the responder has one setup or set up a custom upgrade via the instructions [here](https://docs.harvesterhci.io/v1.1/upgrade/automatic/#prepare-an-air-gapped-upgrade)
```
irishgordo, line 19:
```suggestion
1. Start an upgrade if the [upgrade-responder](https://github.com/harvester/upgrade-responder) has one already available or set up a custom upgrade via the instructions [here](https://docs.harvesterhci.io/v1.1/upgrade/automatic/#prepare-an-air-gapped-upgrade)
```

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,23 @@
...
+## Category: 
+* Upgrade
+
+## Environment Setup
+
+- This is to be done on a Harvester cluster where the NICs were configured to be bonded on install for the management interface
+
+## Verification Steps
+
+1. Install Harvester with at least 2 bonded NICs on install in `balance-tlb`
```
irishgordo, line 16:
```suggestion
1. Install a previously released version of Harvester with at least 2 bonded NICs on install in `balance-tlb`
```

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,23 @@
...
+
+    ![image](https://user-images.githubusercontent.com/83787952/198138366-472d4432-839a-4d9d-a942-d55fe8f6f6d8.png)
```
irishgordo, line 18: I think this picture is great! :smile:   I really dig it's use here to be able to help show what the installation screen should hold.
It renders really well, when I tested locally with Hugo - 
Screenshot from 2022-11-02 10-41-04

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,23 @@
...
+1. Start an upgrade if the responder has one setup or set up a custom upgrade via the instructions [here](https://docs.harvesterhci.io/v1.1/upgrade/automatic/#prepare-an-air-gapped-upgrade)
+
+## Expected Results
+
+- The upgrade should complete successfully and the Harvester cluster should pass sanity checks
```
irishgordo, line 23: QU: Do we need to elaborate on what the sanity checks should be?
I know through - https://harvester.github.io/tests/manual/ - we don't really have a "Sanity Checks" section.
I would figure, though, we might want to ensure elements of networking related work - maybe like VLAN 1 check to ensure that's active on the management network too?

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
irishgordo: @khushboo-rancher - I think right now this would have to be restricted to purely virtualized testing (at least for now :sweat_smile: ), since we do not have a bare metal environment that has two 10GiB NIC cards present in it. - Though I think it could be done on either "if" a bare metal environment was set up with 2 10GiB NIC cards in the node.  

I think single-node would be fine since the management network is restricted to the "create" node.  And since currently after the creation of the mgmt based network we don't allow for it to be modified from the UI directly - to add additional network card ports, x-ref:
Screenshot from 2022-11-02 10-54-31

irishgordo (changes requested):
I'm excited we're updating our docs to cover this!  This looks good and renders well locally on Hugo.  Had a few small suggestions, qu, and comment.

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
noahgildersleeve: Hmm. good point. We do have sanity checks, but they are only outlined in the parent page in the manual/vm section. I'll update that.

noahgildersleeve: Added link.

noahgildersleeve:
Updated PR with suggested changes.

irishgordo (approved):
@noahgildersleeve  looks good :+1: :smile: ! 
LGTM!

khushboo-rancher:
In general LGTM, one minor thing. Could you also add creating VMs and some other basic operation before the upgrade?

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,33 @@
...
+    - Verify all 3 VMs pass [health checks](https://harvester.github.io/tests/manual/virtual-machines/)
+    - Turn off and on the VMs
+
+## Verification Steps
+
+1. Start an upgrade if the [upgrade-responder](https://github.com/harvester/upgrade-responder) has one already available or set up a custom upgrade via the instructions [here](https://docs.harvesterhci.io/v1.1/upgrade/automatic/#prepare-an-air-gapped-upgrade)
+
+## Expected Results
+
+- The upgrade should complete successfully and the Harvester cluster should pass [health checks](https://harvester.github.io/tests/manual/virtual-machines/)
```
irishgordo, line 33: Would we also want to add that the 3 created VMs should be able to pass the health checks?
https://harvester.github.io/tests/manual/virtual-machines/

Since before upgrade we'd validate the VMs being able to pass health checks - then after the upgrade, maybe we'd want to do the same?

docs/content/manual/_incoming/3045-upgrade-with-bonded-nic.md
```diff
@@ -0,0 +1,33 @@
...
+## Category: 
+* Upgrade
+
+## Environment Setup
+
+- This is to be done on a Harvester cluster where the NICs were configured to be bonded on install for the management interface. This can be done in one of two ways.
+    - Single node virtualized environment
+    - Bare metal environment with at least two NICs (this should really be done on 10gig NICs, but can be done on gigabit)
+- Both NICs should be on the same VLAN/network with the same subnet
+- Before you do the upgrade you should do the following
```
irishgordo, line 16:
```suggestion
- Before you do the upgrade you should do the following:
```
(nit) - but not needed :smile:

irishgordo (changes requested):

irishgordo (approved):
Looks good! :+1: :smile: !
LGTM!

khushboo-rancher (approved):

PR #589, opened 2022-11-02
[Cypress] Fix constant.url and window.byId

n313893254:
- As the Harvester route change, we also need to change the constant URL 
- Rancher remove the lookup.js and did not support the `window.byId` function

There are still have 2 area that needs to change:
- Network requires select cluster network
- Export image requires select storage class


n313893254:
Closed in favor https://github.com/DaiYuzeng/tests/pull/2

PR #605, opened 2022-11-11
Fix cypress run order

n313893254:
Preset image and network can make sure to create VM as expected.


TachunLin (approved):
Tested on `v1.1.1-rc1` with Rancher `v2.7.0`
Use `cypress run` to execute the entire test suite in a row , running image and network prior to virtual-machines 

The create vm tests can execute correctly in virtual-machine and template test suite 
image


image

image

PR #632, opened 2022-11-25
[CY] verify UI url

WuJun2016:
1. add Verify the Harvester UI URL:  https://harvester.github.io/tests/manual/ui/verify-url/

### DIFF
--- a/cypress/constants/constants.ts
+++ b/cypress/constants/constants.ts
@@ -20,3 +20,31 @@ export const PageUrl = {
     virtualMachine:   '/harvester/c/local/kubevirt.io.virtualmachine',
     vmNetwork:        '/harvester/c/local/harvesterhci.io.networkattachmentdefinition'
 }
+
+export const MenuNav = {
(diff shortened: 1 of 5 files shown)


TachunLin:
The test suite pass the first test but failed at the second
  image



The second test case `Verify the Harvester icon on the left top corner` failed at  `[testcases/dashboard/nav.spec.ts:21:8]`
  image
  
  ```
  The command was expected to run against origin https://192.168.122.239 
  but the application is at origin about://blank.
  ```

After removing the `experimentalSessionAndOrigin: true,` in `cypress.config.ts` 
The second case can execute correctly in sequence. 
image

TachunLin (changes requested):
By removing the `experimentalSessionAndOrigin: true`, in `cypress.config.ts`. 
The test suite can run correctly. 
Not sure whether removing would impact any existing test cases.

WuJun2016:
@TachunLin  It's been fixed. I tested it several times and it works.  PTAL again, thanks

TachunLin (approved):
LGTM. Tested several time on v1.1.1-rc2 without error. 

image

image

PR #639, opened 2022-11-30
update e2e testing scenarios

lanfon72:
## Changes
- Implement `ClusterNetworkManager` in apiclient for relevant API calls
- Update fixture `enable_vlan` to create cluster network when the Harvester version over `v1.0.3`

### need to update terraform test cases and resources
- [ ] harvester_e2e_tests/integration/test_terraform.py::test_create_network_using_terraform::setup  
- [ ] harvester_e2e_tests/integration/test_terraform.py::test_create_volume_using_terraform::setup   
- [ ] harvester_e2e_tests/scenarios/test_vm_networking.py::test_create_vm_using_terraform::setup   
### randomly failed, need to figure a better way to test
- [ ] harvester_e2e_tests/scenarios/test_create_vm.py::test_create_vm_on_available_cpu_node  
- [ ] harvester_e2e_tests/scenarios/test_create_vm.py::test_update_vm_on_available_cpu_node  
### `affinity` will be added, but while update VM, affinity will be removed automatically, need to check whether it is API issue or not
- [ ] harvester_e2e_tests/scenarios/test_vm_networking.py::TestVMNetworking::test_remove_vlan_network_from_vm 
- [ ] harvester_e2e_tests/scenarios/test_vm_networking.py::test_update_vm_vlan_network_to_management

### DIFF
--- a/apiclient/harvester_api/api.py
+++ b/apiclient/harvester_api/api.py
@@ -5,7 +5,8 @@ from pkg_resources import parse_version
 from requests.packages.urllib3.util.retry import Retry
 
 from .managers import (HostManager, KeypairManager, ImageManager, SettingManager,
-                       NetworkManager, VolumeManager, TemplateManager, SupportBundlemanager)
+                       NetworkManager, VolumeManager, TemplateManager, SupportBundlemanager,
+                       ClusterNetworkManager)
 
 
 class HarvesterAPI:
@@ -42,6 +43,7 @@ class HarvesterAPI:
         self.templates = TemplateManager(self)
         self.supportbundle = SupportBundlemanager(self)
         self.settings = SettingManager(self)
+        self.clusternetworks = ClusterNetworkManager(self)
 
     @property
     def cluster_version(self):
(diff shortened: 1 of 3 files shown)


lanfon72:
Tested on `harvester-runtests#80`

khushboo-rancher (changes requested):
Can you revert the config.yml changes?

khushboo-rancher:
Instead of modifying the existing fixtures, let's create a new directories and have new fixtures there. Because in future it will become tough and tedious work to remove the old code.

lanfon72:
> Instead of modifying the existing fixtures, let's create a new directories and have new fixtures there. Because in future it will become tough and tedious work to remove the old code.

Extending old fixtures is the simplest way to fix multiple test cases, otherwise we will need to update all test cases.
(it would almost equals to re-write those test cases if use new fixtures)

noahgildersleeve (approved):
LGTM. Tested apis/test_networks.py and it passed and properly parsed my v1.1.1 cluster and correctly set VM network and handled cluster netowrks.

khushboo-rancher:
@lanfon72 Can you squash all the commits to 1 commit and then we can merge this.

lanfon72:
> @lanfon72 Can you squash all the commits to 1 commit and then we can merge this.

It requires to update in repository settings.
image

PR #644, opened 2022-12-02
[CY] delete deleteProgramlly function

WuJun2016:
1. When we want to test the UI deletion function, we should call the `delete` method in the `Cru-Resource Class`.
2. We can call the deleteFromStore method when we simply want to delete a previously created object.

So the `deleteProgramlly` method is no longer needed


n313893254 (approved):
LGTM

PR #648, opened 2022-12-06
* [CI] remove legacy python 3.6

lanfon72:


TachunLin (approved):
LGTM. 
Since remove python 3.6 won't affect the e2e-rework.  
And the latest OS version have been updated to Ubuntu 22.04 which would no longer support python 3.6 

We could remove it to prevent future CI build failure.

irishgordo:
This is great!
Thanks for the fix @lanfon72 !
Excited to see us removing Python 3.6, especially since 22.04 doesn't have support and also Python 3.6, hit it's end-of-life date back in Dec. 2021.

PR #654, opened 2022-12-13
[e2e] add integration test cases for VM

lanfon72:
## Changes
- [x] Fix bug of `fake_image_file` fixture
- [x] Add fixture to access VM's console
- [x] Add create minimal VM
    - [x] Image fixture for VM
- Add operation test cases
    - [x] pause
    - [x] unpause
    - [x] stop
    - [x] start
    - [x] restart
    - [x] softreboot
    - [x] migrate
    - [x] abort migrate
    - [x] delete
- **NEW** config parameter `opensuse-image-url` to point out the openSUSE image for testing
- **NEW** fixture `images.py` to replace legacy fixture `image.py`
- **UPDATE** fixture `unique_name` in api_client to `module` scope

### DIFF
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -606,7 +606,7 @@ class VirtualMachineManager(BaseManager):
     # operators: start, restart, stop, migrate, pause, unpause, softreboot
     PATH_fmt = "v1/harvester/kubevirt.io.virtualmachines/{ns}{uid}"
     # guestinfo, network
-    VMI_fmt = "apis/{VM_API}/namespaces/{ns}/virtualmachineinstances/{uid}"
+    VMI_fmt = "v1/harvester/kubevirt.io.virtualmachineinstances/{ns}/{uid}"
     # operators: guestosinfo, console(ws), vnc(ws)
     VMIOP_fmt = "apis/subresources.{VM_API}/namespaces/{ns}/virtualmachineinstances/{uid}/{op}"
 
@@ -623,11 +623,11 @@ class VirtualMachineManager(BaseManager):
             return resp.status_code, resp.content
 
     def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
-        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace, VM_API=self.API_VERSION)
+        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._get(path, raw=raw, **kwargs)
 
     def get_status(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
-        path = self.VMI_fmt.format(uid=name, ns=namespace, VM_API=self.API_VERSION)
+        path = self.VMI_fmt.format(uid=name, ns=namespace)
         return self._get(path, raw=raw, **kwargs)
 
     def create(self, name, vm_spec, namespace=DEFAULT_NAMESPACE, *, raw=False):
@@ -644,7 +644,7 @@ class VirtualMachineManager(BaseManager):
         return self._update(path, vm_spec, raw=raw, as_json=as_json, **kwargs)
 
     def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
-        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace, VM_API=self.API_VERSION)
+        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._delete(path, raw=raw)
 
     def clone(self, name, new_vm_name, namespace=DEFAULT_NAMESPACE, *, raw=False):
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -612,6 +612,16 @@ class VirtualMachineManager(BaseManager):
 
     Spec = VMSpec
 
+    def download_virtctl(self, *, raw=False, **kwargs):
+        code, info = self._get(f"apis/subresources.{self.VM_API}/v1/version")
+        version, platform = info['gitVersion'], info['platform']
+        resp = self.api.session.get("https://github.com/kubevirt/kubevirt/releases/download/"
+                                    f"{version}/virtctl-{version}-{platform}", **kwargs)
+        if raw:
+            return resp
+        else:
+            return resp.status_code, resp.content
+
     def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace, VM_API=self.API_VERSION)
         return self._get(path, raw=raw, **kwargs)
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -150,7 +150,7 @@ class ImageManager(BaseManager):
 
     def create_by_file(self, name, filepath, namespace=DEFAULT_NAMESPACE,
                        description="", display_name=None):
-        file = Path(filepath)
+        file = Path(filepath).expanduser()
 
         data = self.create_data(name, "", description, "upload", namespace, display_name)
         self.create("", namespace, json=data)
@@ -624,6 +624,10 @@ class VirtualMachineManager(BaseManager):
 
     def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace, VM_API=self.API_VERSION)
+        return self._get(path, raw=raw, **kwargs)
+
+    def get_status(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
+        path = self.VMI_fmt.format(uid=name, ns=namespace, VM_API=self.API_VERSION)
         return self._get(path, raw=raw, **kwargs)
 
     def create(self, name, vm_spec, namespace=DEFAULT_NAMESPACE, *, raw=False):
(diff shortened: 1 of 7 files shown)


lanfon72:
tested on `harvester-runtests#96`

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -0,0 +1,374 @@
...
+    yield f"{data['metadata']['namespace']}/{unique_image_id}"
+
+    code, data = api_client.images.delete(unique_image_id)
+
+
+@pytest.fixture(scope="module")
+def unique_vm_name(unique_name):
+    return f"vm-{unique_name}"
+
+
```
noahgildersleeve, line 63: There needs to be a test skel here with the steps per [test skel instructions](https://github.com/harvester/tests#user-content-backend-test-skeletons)

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -0,0 +1,374 @@
...
+        sleep(3)
+    else:
+        raise AssertionError(
+            f"Failed to create Minimal VM({cpu} core, {mem} RAM) with errors:\n"
+            f"Phase: {data.get('status', {}).get('phase')}\t"
+            f"Status: {data.get('status')}\n"
+            f"API Status({code}): {data}"
+        )
+
+
```
noahgildersleeve, line 90: There needs to be a test skel here with the steps per [test skel instructions](https://github.com/harvester/tests#user-content-backend-test-skeletons). I took a look at the subsequent tests and see why there aren't test skels for all of them (they are small and self expalantory). However we should probably add a more overarching test skel with steps at the higher level then.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -0,0 +1,374 @@
...
+                f"Failed to abort VM({unique_vm_name})'s migration, stuck on {states[-1]}\n"
+                f"API Status({code}): {data}"
+            )
+
+        assert cur_host == data['status']['nodeName'], (
+            f"Failed to abort VM({unique_vm_name})'s migration,"
+            f"VM been moved to {data['status']['nodeName']} is not the origin host {cur_host}\n"
+        )
+
+    def test_delete(self, api_client, unique_vm_name, wait_timeout):
```
noahgildersleeve, line 330: Do we need to add the `--do-not-cleanup` check, or are we not doing that anymore?

noahgildersleeve (changes requested):
There's just a couple of things (mostly test skels). Other than that the tests passed for me against a 3 node ipxe-examples `v.1.1.1` install

harvester_e2e_tests/integration/vmconsole.py
```diff
@@ -0,0 +1,60 @@
...
+            if re.search(rf"{self.name} login:\s+?", out, re.M):
+                break
+            print(out)
+            sleep(5)
+        else:
+            raise TimeoutError(-1, "Login timeout: Unable to catch login hints.\n"
+                               f"self.{out}")
+
+        out = self.execute_command(self.user)
+        breakpoint()
```
irishgordo, line 43: Should we remove the `breakpoint()`?

irishgordo (changes requested):

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: Consider following cases:
- A Test case failed, tear down steps (the _cleanup_) will fail too.
- A Test case passed, we will not look into the remaining environment, so _cleanup_ option is not necessary.

as above cases, we will not need the option anymore.

noahgildersleeve: I thought that the `--do-not-cleanup` was mostly for being able to get logs from the environment. I don't have much of an opinion either way I just wanted to check. If we are changing the policy on that though we should update the README.md and remove the requirement for `--do-not-cleanup` in new tests. CC @khushboo-rancher

noahgildersleeve: I do agree that it probably won't matter in a lot of the new tests. the `--do-not-cleanup` option was a bigger deal in the old involved complicated scenarios.

noahgildersleeve: Resolving due to not needing the `--do-not-cleanup` in this test.

harvester_e2e_tests/integration/vmconsole.py
```diff
@@ -0,0 +1,60 @@
...
+
+    def execute_command(self, command, *, timeout=None):
+        if not self.proc:
+            self.login(timeout)
+        proc = self.proc
+        proc.stdin.write(command.encode() + b"\n")
+        proc.stdin.flush()
+        sleep(1)
+        outbytes = len(proc.stdout.peek())
+        return proc.stdout.read(outbytes).decode()
```
irishgordo, line 60: This is awesome!!!
I really like the idea that we have here with leveraging `virtctl` to interact with the VMs on Harvester given that we have the respective `kubeconfig` and other credentials for the VM.
Really excellent find @lanfon72 ! :+1: :raised_hands: 

This reduces so much complexity and overhead of needing to deal with waiting for the VM to be up, grabbing the IPv4 associated with the VM and then juggling something like a Paramiko connection out to it.  Especially if we're doing something fairly complex with the VM post cloud-init stage.  Since we're leveraging Kubevirt behind the scenes on Harvester, `virtctl` as a binary that gets leveraged in this context-manager wrapper is just awesome!

Like if we wanted to do something more complex with actually writing a ton of data to the VM's disk directly once it's up prior to migrating, say like writing 5GB of data so the disk is at least somewhat full:
```
    cpu, mem = 1, 2
    vm = api_client.vms.Spec(cpu, mem)
    vm.add_image("disk-0", image_id)

    code, data = api_client.vms.create(unique_vm_name, vm)

        code, data = api_client.vms.get_status(unique_vm_name)
        if 200 == code and "Running" == data.get('status', {}).get('phase'):
            # some logic to check whether we have network to get virtctl or we need to use a file, if we have network
            # maybe we grabbed it earlier during the testing phase and set it at a global level 
            # as to not try to regrab it again - maybe it's just globally accesible or a fixture?
            with VMConsole(VIRTCTL_GLOBAL, unique_vm_name, ubuntu, ubuntupw, 600) as unique_vm_console:
                write_file_result = unique_vm_console.execute_command('dd if=/dev/urandom of=/home/ubuntu/sample.txt bs=5G count=1')
                # and then other checks or something...
        sleep(3)
    else:
        raise AssertionError(
            f"Failed to create Minimal VM({cpu} core, {mem} RAM) with errors:\n"
            f"Phase: {data.get('status', {}).get('phase')}\t"
            f"Status: {data.get('status')}\n"
            f"API Status({code}): {data}"
        )
```

`execute_command` will really help us to do more complicated testing with VMs on Harvester.
If we want to later, we might be able to expand the `sleep(1)` to be more of an exponential backoff - or just bring in a utility package like [backoff](https://github.com/litl/backoff) and add the decorator to the class function/method - if we're executing a command that may be like "long running" or something like fetching a lot of content via `wget` or something to avoid being rate-limited.

irishgordo (changes requested):

apiclient/harvester_api/managers.py
```diff
@@ -612,10 +612,24 @@ class VirtualMachineManager(BaseManager):
...
+    def download_virtctl(self, *, raw=False, **kwargs):
+        code, info = self._get(f"apis/subresources.{self.VM_API}/v1/version")
+        version, platform = info['gitVersion'], info['platform']
+        resp = self.api.session.get("https://github.com/kubevirt/kubevirt/releases/download/"
+                                    f"{version}/virtctl-{version}-{platform}", **kwargs)
+        if raw:
+            return resp
+        else:
+            return resp.status_code, resp.content
+
```
irishgordo, line 624: Maybe later we can have a more alternative url friendly version of acquiring `virtctl` - like if we had a file-server offline that already had it, so the location for the release wouldn't be tied directly to `github.com/kubevirt/releases/download/...` but something like `http://111.222.333.444:8080/virtctl-1.0.0-amd64`?
Maybe that URL path could just be abstracted to something in the `config.yml` if we need to later.

lanfon72: The `virtctl` binary should always use the version reflect to the cluster, to prevent any incompatible functions.
(It is the reason the URL is fixed)

In airgaped test case, we can add an configure variable to download the `virtctl` file.  But I think the output of serial console should be the same between two environment, so we can just test those cases in general environment.

harvester_e2e_tests/integration/vmconsole.py
lanfon72: Commands executing through the subprocess, backoff will not work.

The Console is not much stable to execute command, so I didn't use it to manipulate VMs.
(But the implementation will still help us to monitor VM's information while rebooting)

irishgordo: We could look into finding a way to make the console more stable later on :smile:. 
I agree that the implementation will help us monitor VM's info during rebooting.
If we could later on use it to manipulate VMs more, that could help expand the depth of our tests, like with the example I was sorta thinking about surrounding us being able to actually write a file to the disk via the console execute_command, vs needing to juggle SSH connection out through a separate library like Paramiko or something.

khushboo-rancher (approved):
LGTM

harvester_e2e_tests/integration/test_vm_functions.py
noahgildersleeve: This is really a fixture and not  a test. Resolving

noahgildersleeve (approved):
LGTM

Issue #657, opened 2022-12-14
[BUG][DevOps]Fix CI/harvester-installer Jenkins restart issue


irishgordo:
Current Loadout:
```
            .-/+oossssoo+/-.               ubuntu@test-jenkins-ci-main-node 
        `:+ssssssssssssssssss+:`           -------------------------------- 
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 22.04.1 LTS x86_64 
    .ossssssssssssssssssdMMMNysssso.       Host: SYS-5018A-FTN4 123456789 
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Kernel: 5.15.0-56-generic 
  +ssssssssshmydMMMMMMMNddddyssssssss+     Uptime: 1 day, 20 hours, 9 mins 
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Packages: 700 (dpkg), 4 (snap) 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Shell: bash 5.1.16 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Resolution: 1024x768 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   Terminal: /dev/pts/0 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   CPU: Intel Atom C2758 (8) @ 2.400GHz 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   GPU: 02:00.0 ASPEED Technology, Inc. ASPEED Graphics Family 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Memory: 293MiB / 32077MiB 
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/
  +sssssssssdmydMMMMMMMMddddyssssssss+                             
   /ssssssssssshdmNNNNmyNMMMMhssssss/                              
    .ossssssssssssssssssdMMMNysssso.
      -+sssssssssssssssssyyyssss+-
        `:+ssssssssssssssssss+:`
            .-/+oossssoo+/-.
```

Issue:
```
fatal: [master]: FAILED! => {
    "changed": false,
    "invocation": {
        "module_args": {
            "daemon_reexec": false,
            "daemon_reload": false,
            "enabled": null,
            "force": null,
            "masked": null,
            "name": "jenkins",
            "no_block": false,
            "scope": "system",
            "state": "restarted"
        }
    },
[...]

PR #658, opened 2022-12-14
bugfix: Fixing Broken Jenkins Provisioning On Debian (non-proxy & non-ssl)

irishgordo:
* fixing jenkins provisioning, new versions are all systemd based services needing additional edits

Resolves: bug/fix-infrastructure-provisioning
Co-Authored-By: ttpcodes <alice.nguyen@suse.com>

Fixes:
- https://github.com/harvester/tests/issues/657
- https://github.com/harvester/tests/issues/659
- https://github.com/harvester/tests/issues/661
- https://github.com/harvester/tests/issues/662

### DIFF
--- a/CI/harvester-installer/roles/jenkins/tasks/install_jenkins_on_Debian.yml
+++ b/CI/harvester-installer/roles/jenkins/tasks/install_jenkins_on_Debian.yml
@@ -57,6 +57,19 @@
     mode: 0644
   when: JENKINS_USE_PROXY
 
+- name: jenkins | build drop in system-d config dir
+  ansible.builtin.shell: >
+    mkdir -p /etc/systemd/system/jenkins.service.d
+
+- name: jenkins | load in the override.conf
+  template:
+    src: override.conf.j2
+    dest: /etc/systemd/system/jenkins.service.d/override.conf
+    mode: 0600
+    force: yes
+  vars:
+    JENKINS_USING_PROXY: "{{ JENKINS_USE_PROXY }}"
+
 - name: jenkins | install_jenkins_on_Debian |
         Create init.groovy.d to bootstrap Jenkins
   file:
@@ -75,10 +88,18 @@
     group: jenkins
     mode: 0755
 
-- name: jenkins | install_jenkins_on_Debian | Restart Jenkins
-  systemd:
+- name: jenkins | install_jenkins | Restart jenkins
+  ansible.builtin.systemd:
     name: jenkins
     state: restarted
+    daemon_reload: yes
+    force: yes
+    enabled: yes
+    masked: no
+  register: result
+  retries: 5
+  delay: 30
+  until: result is success
 
 - name: jenkins | install_jenkins_on_Debian | Enable Jenkins port 8080
   shell: ufw allow 8080
--- a/CI/harvester-installer/roles/jenkins/tasks/install_vagrant_on_Debian.yml
+++ b/CI/harvester-installer/roles/jenkins/tasks/install_vagrant_on_Debian.yml
@@ -1,7 +1,10 @@
 ---
+# NOTE: this is strange...I have no idea why the `-1` on the package has to be present now...but it seems they changed
+# the way they are serving the packages... idk... here's to being "optimistic" that they won't change their file
+# serving pattern again in the future...
 - name: jenkins | install_vagrant_on_Debian | Install vagrant package
   apt:
-    deb: https://releases.hashicorp.com/vagrant/{{ LATEST_VAGRANT_VERSION }}/vagrant_{{ LATEST_VAGRANT_VERSION }}_x86_64.deb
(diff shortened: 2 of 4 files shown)


irishgordo:
New bug discovered:
https://github.com/harvester/tests/issues/659
Updating description...

irishgordo:
Also discovered:
https://github.com/harvester/tests/issues/660

But could be for future iteration...

irishgordo:
Additionally discovered:
https://github.com/harvester/tests/issues/661

CI/harvester-installer/roles/jenkins/tasks/install_jenkins_on_Debian.yml
```diff
@@ -3,7 +3,7 @@
         Install prerequisite packages for Jenkins
   apt:
     name: [ca-certificates, openjdk-11-jdk, curl, apt-transport-https, gnupg,
-           python3, python3-pip, figlet]
+           python3, python3-pip, figlet, sshpass]
```
noahgildersleeve, line 6: Where is sshpass being used?

CI/harvester-installer/roles/jenkins/templates/override.conf.j2
```diff
@@ -0,0 +1,7 @@
+# https://www.jenkins.io/doc/book/system-administration/systemd-services/
+[Service]
+Environment="JAVA_OPTS=-Djenkins.install.runSetupWizard=false -Dcasc.jenkins.config=/var/lib/jenkins/casc_configs -Djava.awt.headless=true"
+Environment="CASC_JENKINS_CONFIG=/var/lib/jenkins/casc_configs/"
+{% if JENKINS_USING_PROXY %}
+Environment="JENKINS_OPTS=--httpListenAddress=127.0.0.1"
+{% endif %}
```
noahgildersleeve, line 7: Needs a newline at end of file

CI/harvester-installer/roles/jenkins/tasks/install_vagrant_on_Debian.yml
```diff
@@ -1,7 +1,10 @@
 ---
+# NOTE: this is strange...I have no idea why the `-1` on the package has to be present now...but it seems they changed
+# the way they are serving the packages... idk... here's to being "optimistic" that they won't change their file
+# serving pattern again in the future...
 - name: jenkins | install_vagrant_on_Debian | Install vagrant package
   apt:
-    deb: https://releases.hashicorp.com/vagrant/{{ LATEST_VAGRANT_VERSION }}/vagrant_{{ LATEST_VAGRANT_VERSION }}_x86_64.deb
+    deb: https://releases.hashicorp.com/vagrant/{{ LATEST_VAGRANT_VERSION }}/vagrant_{{ LATEST_VAGRANT_VERSION }}-1_amd64.deb
```
noahgildersleeve, line 7: Is this going to the version, then {{version_number}}-1, or did the URL change somehow? It almost looks like that is a second file that had a `-1` added to it to prevent filename collision.

noahgildersleeve (changes requested):

CI/harvester-installer/roles/jenkins/tasks/install_jenkins_on_Debian.yml
irishgordo: I've removed that as it's not "needed" for the fix, that's more of the future state dependency for the air-gap work.
So as a separation of concerns, I've cleaned that up :smile: :+1:  good catch! @noahgildersleeve

CI/harvester-installer/roles/jenkins/templates/override.conf.j2
irishgordo: Good catch, I've updated that, thanks @noahgildersleeve :smile:

CI/harvester-installer/roles/jenkins/tasks/install_vagrant_on_Debian.yml
irishgordo: The `-1` is this new weird thing that I mentioned in the code-comment above this:
```
# NOTE: this is strange...I have no idea why the `-1` on the package has to be present now...but it seems they changed
# the way they are serving the packages... idk... here's to being "optimistic" that they won't change their file
# serving pattern again in the future..
```
Like their file serving seems to have changed with Hashicorp.
For instance, if you notice here:
https://releases.hashicorp.com/vagrant/2.3.4/
They append the `-1` to the end of the deb package, which is what we need.
And they seem to have also done that for:
https://releases.hashicorp.com/vagrant/2.3.3/
So I'm "guessing" that moving forward this will be Hashicrops file serving way of distributing the packages

noahgildersleeve (approved):

noahgildersleeve:
LGTM

khushboo-rancher (approved):

Issue #646, opened 2022-12-06
[TEST] Rework on the image test cases


TachunLin:
## What's the test to develop? Please describe

Implementation of the image related test cases.

## Describe the items of the test development (DoD, definition of done) you'd like

- [x] Create image from Volume
- [x] Create ISO, raw, windows image.
- [x] Create with invalid image
- [x] Edit images which is not in use and which is in use.
- [x] Delete image and create with same name.

TachunLin:
Done the initial draft of following test plans: 
1. Create ISO, raw, windows image.
    - test_create_image_url
    - test_upload_image_file
2. Create with invalid image 
    - test_create_invalid_url
    - test_create_invalid_file
3. Delete image and create with same name.
    - test_delete_image_recreate

TachunLin:
Add the following test cases 
1. Export volume to image
    - test_create_image_from_volume 
2. Edit images which is not in use and which is in use
    - test_edit_image_in_use
    - test_edit_image_not_in_use

TachunLin:
Compare network backend rework with existing daily api test to find duplicate. 

|                                   | Daily API Test | Backend Rework |
|-----------------------------------|:--------------:|:--------------:|
| test_create_image_from_volume     |                |        V       |
| test_create_image_url (qcow2)     |        V       |        V       |
| test_create_image_url (img)       |                |        V       |
| test_upload_image_file            |        V       |        V       |
| test_delete_image_recreate (URL)  |                |        V       |
| test_delete_image_recreate (File) |                |        V       |
| test_create_invalid_url           |                |        V       |
| test_create_invalid_file          |                |        V       |
| test_edit_image_in_use            |        V       |        V       |
| test_edit_image_not_in_use        |                |        V       |
| test_delete                       |        V       |        V       |

PR #663, opened 2022-12-16
[e2e] add initial integration test cases for image 

TachunLin:
Add the initial image backend test of the following test plans:

- [x] Create ISO, raw, windows image.
  - test_create_image_url
  - test_upload_image_file
- [x] Create with invalid image
  - test_create_invalid_url
  - test_create_invalid_file
- [x] Delete image and create with same name.
  - test_delete_image_recreate
  
- [x] Export volume to image
  - test_create_image_from_volume
  
- [x] Edit images which is not in use and which is in use
  - test_edit_image_in_use
  - test_edit_image_not_in_use

### DIFF
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -210,6 +210,13 @@ class VolumeManager(BaseManager):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._delete(path, raw=raw)
 
+    def export(self, name, image_name, storage_class, namespace=DEFAULT_NAMESPACE, *, raw=False):
+        export_spec = {"displayName": image_name, "namespace": namespace,
+                       "storageClassName": storage_class}
+        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
+        params = dict(action="export")
+        return self._create(path, params=params, json=export_spec, raw=raw)
+
 
 class TemplateManager(BaseManager):
     # get, create, delete
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -15,10 +15,12 @@
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
 
+
 from time import sleep
 from datetime import datetime, timedelta
 
 import pytest
+
 
 pytest_plugins = [
     "harvester_e2e_tests.fixtures.api_client",
@@ -87,11 +89,14 @@ class TestImages:
                 "test-label": "42"
             },
             "annotations": {
-                "test-annotation": "dummy"
-            }
+                "test-annotation": "dummy",
+                "field.cattle.io/description": 'test description'
+            },
+
         }
 
         code, data = api_client.images.update(unique_name, dict(metadata=updates))
+
         assert 200 == code, (f"Failed to update image with error: {code}, {data}")
 
         unexpected = list()
--- a/harvester_e2e_tests/integration/test_images.py
+++ b/harvester_e2e_tests/integration/test_images.py
@@ -1,14 +1,15 @@
-import pytest
+
+import os
+
 from time import sleep
 from datetime import datetime, timedelta
+import pytest
+
 
 pytest_plugins = [
-    'harvester_e2e_tests.fixtures.api_endpoints',
-    'harvester_e2e_tests.fixtures.api_client',
-    'harvester_e2e_tests.fixtures.volume',
-    'harvester_e2e_tests.fixtures.session',
+    "harvester_e2e_tests.fixtures.api_client",
     'harvester_e2e_tests.fixtures.image',
-    'harvester_e2e_tests.fixtures.images',
+    'harvester_e2e_tests.fixtures.vm',
 ]
 
 
@@ -19,9 +20,290 @@ def test_create_with_url(api_client, image_opensuse, unique_name, wait_timeout, 
     """
     Test if you can create an image from a URL.
 
-    Prerequisite:
-    Setting opensuse-image-url set to a valid URL for
-    an opensuse image.
+    def teardown_class(self):
+        os.system('rm *.gzps')
+        os.system('rm *.qcow2')
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="create_image_from_volume")
+    def test_create_image_from_volume(self, api_client, unique_name, export_storage_class):
+        """
+    Test create image from volume
+
+    Steps:
+        1. Create a volume "test-volume" in Volumes page
+        2. Export the volume to image "export-image"
+        3. Check the image "export-image" exists
+        4. Cleanup image "export-image" on Images page
+        5. Cleanup volume "test-volume" on Volumes page
+        """
+
+        spec = api_client.volumes.Spec(1)
+        code, data = api_client.volumes.create(unique_name, spec)
+
+        assert 201 == code, (code, data)
+
+        api_client.volumes.get(unique_name)
+        sleep(5)
+        api_client.volumes.export(unique_name, unique_name, export_storage_class)
+        sleep(30)
+
+        image_id = self.get_export(api_client, unique_name)
+
+        self.delete_image(api_client, image_id, 10)
+
+        self.delete_volume(api_client, unique_name, 10)
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="create_image_url")
+    def test_create_image_url(self, api_client, qcow2_image_url, iso_image_url):
+        """
+        Test create raw and iso type image from url
+
+        Steps:
+        1. Open image page and select default URL
+        2. Input qcow2 image file download URL, wait for download complete
+        3. Check the qcow2 image exists
+        4. Input iso image file download URL, wait for download complete
+        5. Check the iso image exists
+        """
+
+        self.create_image_url(api_client, 'opensuse', qcow2_image_url, 600)
+        self.create_image_url(api_client, 'k3os', iso_image_url, 600)
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="create_image_file")
+    def test_upload_image_file(self, api_client):
+        """
+        Test create raw and iso type image from file
+
+        Steps:
+        1. Open image page and select default URL
+        2. Upload an qcow2 file type image from file
+        3. Check the the uploaded file exists
+        """
+
+        self.upload_file(api_client, 'test-file-upload', 'test-upload.qcow2', '10M')
+
+        self.get_image(api_client, 'test-file-upload')
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(depends=["create_image_url", "create_image_file"])
+    def test_delete_image_recreate(self, api_client, iso_image_url):
+        """
+        Test create raw and iso type image from file
+
+        Steps:
+        1. Check the image created by URL exists
+        2. Delete the newly created image
+        3. Create an iso file type image from URL
+        4. Check the iso image exists
+        5. Upload an qcow2 file type image
+        5. Delete the newly uploaded file
+        6. Upload a new qcow2 file type image
+        """
+
+        self.get_image(api_client, 'k3os')
+        self.delete_image(api_client, 'k3os', 15)
+
+        self.create_image_url(api_client, 'k3os', iso_image_url, 600)
+
+        self.get_image(api_client, 'test-file-upload')
+        self.delete_image(api_client, 'test-file-upload', 15)
+
+        self.upload_file(api_client, 'test-file-upload', 'test-upload.qcow2', '10M')
+
+    @pytest.mark.images_p1
+    def test_create_invalid_url(self, api_client, unique_name):
+        """
+        Test create image from invalid url
+
+        Steps:
+        1. Create image with invalid URL. e.g. - https: // test.img
+        2. Check the Image state show as Failed
+        """
+        self.create_with_invalid_url(api_client, unique_name)
+
+    @pytest.mark.images_p1
+    def test_create_invalid_file(self, api_client):
+        """
+        Test create upload image from invalid file type
+
+        Steps:
+        1. Try to upload invalid image file to images page, something like dmg, or tar.gzps
+        2. Check should get an error
+        """
+
+        os.system('''cat <<-EOF >>test-invalid.tar.gzps \ntest\nEOF''')
+        resp = api_client.images.create_by_file('test-invalid-file', 'test-invalid.tar.gzps')
+
+        assert 500 == resp.status_code, (
+            f"Failed to upload unsupported file type:{resp.status_code}, {resp.content}"
+        )
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="edit_image_in_use")
+    def test_edit_image_in_use(self, api_client, unique_name):
+        """
+        Test can edit image which already in use
+
+        Steps:
+        1. Check the image created from URL exists
+        2. Create a VM and use the existing image created from url
+        3. Update the image labels and description
+        4. Check can change the image content
+        """
+
+        self.get_image(api_client, 'k3os')
+
+        spec = api_client.vms.Spec(1, 1)
+        spec.add_image('k3os', "default/k3os")
+        code, data = api_client.vms.create(unique_name, spec)
+
+        assert 201 == code, (f"Failed to create vm with error: {code}, {data}")
+
+        updates = {
+            "labels": {
+                "usage-label": "yes"
+            },
+            "annotations": {
+                "field.cattle.io/description": 'edit image in use'
+            },
+
+        }
+
+        self.update_by_input(api_client, 'k3os', updates)
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="edit_image_not_in_use")
+    def test_edit_image_not_in_use(self, api_client, qcow2_image_url, iso_image_url):
+        """
+        Test can edit image which not in use
+
+        Steps:
+        1. Check the image created from URL exists
+        2. Update the image labels and description
+        3. Check can change the image content
+        """
+
+        self.get_image(api_client, 'k3os')
+
+        updates = {
+            "labels": {
+                "usage-label": "no"
+            },
+            "annotations": {
+                "field.cattle.io/description": 'edit image not in use'
+            },
+
+        }
+
+        self.update_by_input(api_client, 'opensuse', updates)
+
+    def update_by_input(self, api_client, unique_name, updates):
+
+        code, data = api_client.images.update(unique_name, dict(metadata=updates))
+
+        assert 200 == code, (f"Failed to update image with error: {code}, {data}")
+
+        unexpected = list()
+        for field, pairs in updates.items():
+            for k, val in pairs.items():
+                if data['metadata'][field].get(k) != val:
+                    unexpected.append((field, k, val, data['metadata'][field].get(k)))
+
+        assert not unexpected, (
+            "\n".join(f"Update {f} failed, set key {k} as {v} but got {n}"
+                      for f, k, v, n in unexpected)
+        )
+
+    def upload_file(self, api_client, unique_name, file_name, size):
+        create_image_cmd = "qemu-img create " + file_name + " " + size
+        os.system(create_image_cmd)
+        resp = api_client.images.create_by_file(unique_name, file_name)
+
+        assert 200 == resp.status_code, (
+            f"Failed to upload fake image with error:{resp.status_code}, {resp.content}"
+        )
+
+    def get_export(self, api_client, display_name):
+        code, data = api_client.images.get()
+        assert 200 == code, (code, data)
+
+        for item in data['items']:
+            if item['spec']['displayName'] == display_name:
+                image_name = item['metadata']['name']
+                break
+            else:
+                raise AssertionError(
+                    f"Failed to find image {display_name}"
+                )
+        return image_name
+
+    def create_image_url(self, api_client, display_name, image_url, wait_timeout):
+        code, data = api_client.images.create_by_url(display_name, image_url)
+
+        assert 201 == code, (code, data)
+        image_spec = data.get('spec')
+
+        assert display_name == image_spec.get('displayName')
+        assert "download" == image_spec.get('sourceType')
+
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+
+        while endtime > datetime.now():
+            code, data = api_client.images.get(display_name)
+            image_status = data.get('status', {})
+
+            assert 200 == code, (code, data)
+            if image_status.get('progress') == 100:
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to download image {display_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+    def delete_image(self, api_client, unique_name, wait_timeout):
+        code, data = api_client.images.delete(unique_name)
+
+        assert 200 == code, (f"Failed to delete image with error: {code}, {data}")
+
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+
+        while endtime > datetime.now():
+            code, data = api_client.images.get(unique_name)
+            if code == 404:
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to delete image {unique_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+    def delete_volume(self, api_client, unique_name, wait_timeout):
+        code, data = api_client.volumes.delete(unique_name)
+
+        assert 200 == code, (f"Failed to delete volume with error: {code}, {data}")
+
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        while endtime > datetime.now():
+            code, data = api_client.volumes.get(unique_name)
+            if code == 404:
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to delete volume {unique_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+    def get_image(self, api_client, unique_name):
+        code, data = api_client.images.get()
+
+        assert len(data['items']) > 0, (code, data)
 
     1. Create an image from URL.
     2. Check for 201 response.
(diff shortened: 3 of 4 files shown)


TachunLin:
Verified can pass the test suite on local Harvester environment of v1.1.2 master 
image

TachunLin:
Verified can pass the test suite on Jenkins jobs `harvester-runtest` 82 on dolphin bare metal clusters
image

apiclient/harvester_api/managers.py
```diff
@@ -203,6 +203,13 @@ def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._delete(path, raw=raw)
 
+    def export(self, volume_name, image_name, namespace=DEFAULT_NAMESPACE, *, raw=False):
+        export_spec = {"displayName": image_name,
+                       "namespace": namespace, "storageClassName": "harvester-longhorn"}
+
+        path = self.PATH_fmt.format(uid="", ns=namespace) + "/{}?action=export".format(volume_name)
```
lanfon72, line 210: We should use `params` for query string instead of string concatenation.
```python
path = self.PATH_fmt.format(uid="", ns=namespace) + "/{}?action=export".format(volume_name)
# equals to
path = self.PATH_fmt.format(uid=f"/{volume_name}", ns=namespace)
# without `?action=export` query string
```

`VirtualMachineManager` have some examples for query string, it would be something like
```python
params = dict(action="export")
self._create(path, params=params)
```

apiclient/harvester_api/managers.py
```diff
@@ -203,6 +203,13 @@ def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._delete(path, raw=raw)
 
+    def export(self, volume_name, image_name, namespace=DEFAULT_NAMESPACE, *, raw=False):
+        export_spec = {"displayName": image_name,
+                       "namespace": namespace, "storageClassName": "harvester-longhorn"}
```
lanfon72, line 208: for the method definition, consider the document would like:
```
class harvester_api.VolumeManger
    export(volume_name, image_name, namespace='default', *, raw=False)
        # the description of the method
```

I would prefer to use `name` instead of `volume_name` for the first argument.

`storageClassName` should be one of the argument, same as the UI provides

harvester_e2e_tests/apis/test_images.py
```diff
@@ -19,6 +19,7 @@
 from datetime import datetime, timedelta
 
 import pytest
+import os
```
lanfon72, line 22: standard library should be placed in the first group.[^pep8]

[^pep8]: https://peps.python.org/pep-0008/#imports

harvester_e2e_tests/apis/test_images.py
```diff
@@ -85,11 +86,31 @@ def test_update(self, api_client, unique_name):
...
+            for k, val in pairs.items():
+                if data['metadata'][field].get(k) != val:
+                    unexpected.append((field, k, val, data['metadata'][field].get(k)))
+
+        assert not unexpected, (
+            "\n".join(f"Update {f} failed, set key {k} as {v} but got {n}"
+                      for f, k, v, n in unexpected)
+        )
+
+    def test_update_by_input(self, api_client, unique_name, updates):
```
lanfon72, line 110: fixture `updates` is missing, when modifying old test cases, we should always test it.

for the two test cases, you might need check it again, update then test it.

harvester_e2e_tests/apis/test_images.py
```diff
@@ -137,6 +158,31 @@ def test_create_with_reuse_display_name(self, api_client, unique_name, fake_imag
 
         _ = api_client.images.delete(unique_name)
 
+    @pytest.mark.dependency(name=["upload_image"])
+    def test_upload_file(self, api_client, unique_name, file_name, size):
```
lanfon72, line 162: where are fixtures `file_name` and `size`?

harvester_e2e_tests/apis/test_images.py
```diff
@@ -161,3 +207,29 @@ def test_create_with_invalid_url(api_client, unique_name):
     assert "no such host" in image_conds[0].get("message")
 
     api_client.images.delete(unique_name)
+
+
+def create_image_url(api_client, display_name, image_url, wait_timeout):
```
lanfon72, line 212: do we provide `display_name` and `image_url` as a fixture?

and for a test, function name should starts with `test`

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+
+pytest_plugins = [
+    "harvester_e2e_tests.fixtures.api_client",
+    'harvester_e2e_tests.fixtures.image',
+    'harvester_e2e_tests.fixtures.vm',
+]
+
+
+@pytest.mark.images_p1
+@pytest.mark.dependency(name="create_image_from_volume")
```
lanfon72, line 18: We should use new markers such like:
```python
@pytest.mark.p0
@pytest.mark.images
```
you might need check all markers for following test cases.

the dependency mark is not necessary when there don't have any following test cases depends on it.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
+
+from harvester_e2e_tests.apis import test_images, test_volumes
+
+from time import sleep
+
+import pytest
+import os
```
lanfon72, line 7: https://peps.python.org/pep-0008/#imports

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+
+
+pytest_plugins = [
+    "harvester_e2e_tests.fixtures.api_client",
+    'harvester_e2e_tests.fixtures.image',
+    'harvester_e2e_tests.fixtures.vm',
```
lanfon72, line 13: I don't see any fixtures are requested from `vm` fixture.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+    Steps:
+    1. Create a volume "test-volume" in Volumes page
+    2. Export the volume to image "export-image"
+    3. Check the image "export-image" exists
+    4. Cleanup image "export-image" on Images page
+    5. Cleanup volume "test-volume" on Volumes page
+    """
+
+    volume_name = 'test-volume'
+    export_name = 'export-image'
```
lanfon72, line 32: `volume_name` and `export_name` should use unique name instead of the **hard-coded** one.
we already have `unique_name` fixture for you to make it possible.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+
+    testvolumes = test_volumes.TestVolumes()
+    testimages = test_images.TestImages()
+
+    testvolumes.test_create(api_client, volume_name)
```
lanfon72, line 37: That's not the right way to do the test.

We should **NOT** use any test cases in another test case.
Test cases should be isolated and not be chained.
IT IS the reason why we need to refactor our e2e testing.

following test cases have same issue, you will need check and update them too.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+        test
+        EOF
+    ''')
+    resp = api_client.images.create_by_file('test-invalid-file', 'test-invalid.tar.gzps')
+
+    assert 500 == resp.status_code, (
+        f"Failed to upload unsupported file type:{resp.status_code}, {resp.content}"
+    )
+
+    os.system('rm test-img.tar.gzps')
```
lanfon72, line 149: this is not the right way to implement the test.
what if the test case failed? the `rm` command will never executed.

you should use NamedTemporaryFile[^1] instead.

[^1]: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+    3. Update the image labels and description
+    4. Check can change the image content
+    """
+    testimages = test_images.TestImages()
+
+    testimages.test_get(api_client, 'k3os')
+
+    spec = api_client.vms.Spec(1, 1)
+    spec.add_image('k3os', "default/k3os")
+    code, data = api_client.vms.create('test-k3os', spec)
```
lanfon72, line 170: This is not the right way to implement it.
what if there already have a VM named `test-k3os`?

we should leverage `unique_name` to make names unique as possible.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,212 @@
...
+        "labels": {
+            "usage-label": "yes"
+        },
+        "annotations": {
+            "field.cattle.io/description": 'edit image in use'
+        },
+
+    }
+
+    testimages.test_update_by_input(api_client, 'k3os', updates)
```
lanfon72, line 184: If you create the test `test_update_by_input` is for the case, then you should not add it to API tests.

lanfon72 (changes requested):
If you have problems to implement those test cases correctly, you can schedule a time for us to discuss those issues.

apiclient/harvester_api/managers.py
TachunLin: Update related lines according to comment

apiclient/harvester_api/managers.py
TachunLin: Update related lines according to comment, add `storage_class` for `storageClassName`

harvester_e2e_tests/apis/test_images.py
TachunLin: Update related import os to the first group in integration/test_images.py

harvester_e2e_tests/apis/test_images.py
TachunLin: The `updates` used in  `test_update_by_input` act as parameter called from integration/test_images.py 
Moved the `test_update_by_input`, `test_upload_file` and `test_get_export` from apis/test_images.py to integration/test_images.py to isolate the test and wrap them in `TestBackendImages` class

harvester_e2e_tests/apis/test_images.py
TachunLin: The `file_name` and `size` value did not set as fixture since they were sent as parameter value via test function call.

harvester_e2e_tests/integration/test_images.py
TachunLin: Use `unique_name` for `volume_name` and `export_name`

harvester_e2e_tests/integration/test_images.py
TachunLin: Move explicit request of other test case to in each test plan to integration/test_image only and wrap into class

harvester_e2e_tests/integration/test_images.py
TachunLin: Add the file clean up in class level function `def teardown_class(self):` 

When using the `fake_image_file` fixture in tests 
  ```
  @pytest.mark.images_p1
      def test_fake_image(self, api_client, unique_name, fake_image_file):
  
          # testimages = test_images.TestImages()
          resp = api_client.images.create_by_file(unique_name, fake_image_file)
  
          assert 200 == resp.status_code, (
              f"Failed to upload fake image with error:{resp.status_code}, {resp.content}"
          )
  
  ``` 
Test execution encounter failure with error message
  ```
  Timeout waiting for the datasource file processing begin
  ```
  image

We have related issue to track the status.  
https://github.com/harvester/harvester/issues/3259 

In order to create failed file type, thus I try using os command line as to achieve it.

harvester_e2e_tests/integration/test_images.py
TachunLin: Update vm name from `test-k3os` to `unique_name`

harvester_e2e_tests/integration/test_images.py
TachunLin: Move test_update_by_input from `apis/test_images` to `integration/test_images` and wrap in `TestBackendImages` class

TachunLin:
Update according to the review comment. 
Tested can pass all test plans on local environment on Harvester v1.1.1 
  image

Will test again on Jenkins jobs

TachunLin:
Rework according to review comment, update the following changes
- [x] Remove duplicate test plans, according to https://github.com/harvester/tests/issues/646#issuecomment-1375606021
  - test_upload_image_file
  - test_create_invalid_url
  - test_edit_image_in_use
  - test_edit_image_not_in_use
- [x] Remove sleep action and use wait action instead
- [x] Move helper function outside of test class
- [x] Remove tear down action 
- [x] Use fixture to handle iso name 
- [x] Use unique name to handle create valid and invalid file upload


* Verified on local kvm environment
image

*  Verified on dolphin bare machines 
image

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+        assert 201 == code, (code, data)
+
+        get_volume(api_client, unique_name)
+
+        api_client.volumes.export(unique_name, unique_name, export_storage_class)
+
+        image_id = get_export(api_client, unique_name, wait_timeout)
+
+        delete_image(api_client, image_id, wait_timeout)
```
lanfon72, line 246: we should also tear down the created volume.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="create_image_url")
+    def test_create_image_url(self, api_client, request, iso_image_url,
+                              qcow2_image_url, qcow2_name, iso_name):
```
lanfon72, line 251: we can use `pytest.mark.parametrize`[^1] or parameterized fixture[^2].

[^1]: https://docs.pytest.org/en/6.2.x/parametrize.html 
[^2]: https://docs.pytest.org/en/6.2.x/fixture.html#fixture-parametrize

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+        5. Delete the newly uploaded file
+        6. Upload a new qcow2 file type image
+        """
+        wait_timeout = request.config.getoption('--wait-timeout')
+
+        get_image(api_client, qcow2_name)
+        delete_image(api_client, qcow2_name, wait_timeout)
+
+        create_image_url(api_client, qcow2_name, iso_image_url, wait_timeout)
+        get_image(api_client, qcow2_name)
```
lanfon72, line 289: when creating image, we should also check the image State becomes **Ready**

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+    assert "Initialized" == image_conds[0].get("type")
+    assert "False" == image_conds[0].get("status")
+    assert "no such host" in image_conds[0].get("message")
+
+    api_client.images.delete(unique_name)
+
+
+class TestBackendImages:
+
+    @pytest.mark.images_p1
```
lanfon72, line 219: `images_p1` is deprecated markers, update them to use `images` and `p0/p1/p2`, to notice the highest priority is `p0`.

and can just mark the class, it will mark all methods inside.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+        5. Check the iso image exists
+        """
+        wait_timeout = request.config.getoption('--wait-timeout')
+
+        create_image_url(api_client, qcow2_name, qcow2_image_url, wait_timeout)
+        create_image_url(api_client, iso_name, iso_image_url, wait_timeout)
+
+    @pytest.mark.images_p1
+    @pytest.mark.dependency(name="delete_image_recreate", depends=["create_image_url"])
+    def test_delete_image_recreate(self, api_client, request,
```
lanfon72, line 269: This test case should also use pytest.mark.parametrize or parameterized fixture.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+        get_image(api_client, unique_name)
+        delete_image(api_client, unique_name, wait_timeout)
+
+    @pytest.mark.images_p1
+    def test_create_invalid_file(self, api_client, unique_name, fake_invalid_image_file):
+        """
+        Test create upload image from invalid file type
+
+        Steps:
+        1. Try to upload invalid image file to images page, something like dmg, or tar.gzps
```
lanfon72, line 315: The suffix of `fake_invalid_image_file` is not one of `dmg`, `tar` or `gzip`.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+    def test_delete_image_recreate(self, api_client, request,
+                                   iso_image_url, unique_name, qcow2_name, fake_image_file):
+        """
+        Test create raw and iso type image from file
+
+        Steps:
+        1. Check the image created by URL exists
+        2. Delete the newly created image
+        3. Create an iso file type image from URL
+        4. Check the iso image exists
```
lanfon72, line 278: it looks we don't test iso file?

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+            "labels": {
+                "usage-label": "yes"
+            },
+            "annotations": {
+                "field.cattle.io/description": 'edit image in use'
+            },
+
+        }
+
+        update_by_input(api_client, qcow2_name, updates)
```
lanfon72, line 375: If the function will only be called once, we could just place codes in here.

and we should remove the VM/volume/image after test.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+    @pytest.mark.dependency(name="edit_image_in_use", depends=["create_image_url"])
+    def test_edit_image_in_use(self, api_client, request, unique_name, qcow2_name):
+        """
+        Test can edit image which already in use
+
+        Steps:
+        1. Check the image created from URL exists
+        2. Create a VM and use the existing image created from url
+        3. Update the image labels and description
+        4. Check can change the image content
```
lanfon72, line 335: To make the image `in use` state, we can just create a image volume from the image, it would be the same behavior when creating a VM with the image.

lanfon72 (changes requested):
Please rebase main branch and revert `fixtures/image.py`, it is legacy fixture and has been deprecated. (ref: https://github.com/harvester/tests#contribute)

harvester_e2e_tests/integration/test_images.py
TachunLin: Thanks for the review. 
Remove `import os` and corresponding function `upload_file`

harvester_e2e_tests/integration/test_images.py
TachunLin: Thanks for the review. Remove uncessary `vm` fixture

harvester_e2e_tests/integration/test_images.py
TachunLin: Thanks for the review. 
Mark test class and test cases with `pytest.mark.p0` fixture

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+        Test create image from volume
+
+        Steps:
+        1. Create a volume "test-volume" in Volumes page
+        2. Export the volume to image "export-image"
+        3. Check the image "export-image" exists
+        4. Cleanup image "export-image" on Images page
+        5. Cleanup volume "test-volume" on Volumes page
+        """
+        wait_timeout = request.config.getoption('--wait-timeout')
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review. 
Use fixtures in `api_client`

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+        spec = api_client.volumes.Spec(1)
+        code, data = api_client.volumes.create(unique_name, spec)
+
+        assert 201 == code, (code, data)
+
+        get_volume(api_client, unique_name)
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review. 
Add the `check volume ready` mechanism to wait for volume state in ready

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+        api_client.volumes.export(unique_name, unique_name, export_storage_class)
+
+        image_id = get_export(api_client, unique_name, wait_timeout)
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review. 
Rewrite the `get_export` function to wait for image in ready state

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+
+@pytest.fixture(scope='session')
+def export_storage_class():
+    storage_class = "harvester-longhorn"
+    return storage_class
+
+
+@pytest.fixture(scope="session")
+def fake_image_file():
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review, remove `qcow2_url` , `iso_image_url`  and `fake_image_file` 
The remaining are needed and have different usage with api_client

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,375 @@
...
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.images.get()
+        assert 200 == code, (code, data)
+
+        for item in data['items']:
+            if item['spec']['displayName'] == display_name:
+                image_name = item['metadata']['name']
+                endtime = datetime.now()
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review. 
Rewrite the `get_export` function to wait for image in ready state

harvester_e2e_tests/integration/test_images.py
TachunLin: Thanks for the review. 
Use `pytest.mark.parametrize` for `test_create_image_url` and `test_delete_image_recreate`

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+    code, data = api_client.images.get()
+
+    assert len(data['items']) > 0, (code, data)
+
+    code, data = api_client.images.get(unique_name)
+    assert 200 == code, (code, data)
+    assert unique_name == data['metadata']['name']
+
+
+def create_with_invalid_url(api_client, unique_name):
```
lanfon72, line 176: function not used.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+            "labels": {
+                "usage-label": "yes"
+            },
+            "annotations": {
+                "field.cattle.io/description": 'edit image in use'
+            },
+
+        }
+
+        update_by_input(api_client, qcow2_name, updates)
```
lanfon72, line 380: It looks `update_by_input` will only be called once, same as `get_export`.

To notice each test case is isolated and independent, when creating a helper function, it should not have any side effects (like assertion inside), otherwise we will need to debug the helper function rather than the test case; and when we update the helper function to fix the error, it will affect other test cases which calling the function.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+        for item in data['items']:
+            if item['spec']['displayName'] == display_name:
+                if "progress" in item['status']:
+                    if item['status']['progress'] == 100:
+                        image_name = item['metadata']['name']
+
+            else:
+                raise AssertionError(
+                    f"Failed to find image {display_name}"
+                )
```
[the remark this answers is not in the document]
bk201: @TachunLin We need to fix this. We need some pause between API calls and also handle the timeout case.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+        4. Cleanup image "export-image" on Images page
+        5. Cleanup volume "test-volume" on Volumes page
+        """
+
+        spec = api_client.volumes.Spec(1)
+        code, data = api_client.volumes.create(unique_name, spec)
+
+        assert 201 == code, (code, data)
+
+        code, data = api_client.volumes.get(unique_name)
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review and suggestion. Remove the duplicated call to get volume state

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+            if data["status"]["phase"] == "Bound":
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to delete volume {unique_name} bound in {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+        get_volume(api_client, unique_name)
```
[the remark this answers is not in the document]
TachunLin: Thanks for the review. Remove the duplicate call of get_volume since it's already covered in previous while loop

harvester_e2e_tests/integration/test_images.py
```diff
@@ -1,47 +1,385 @@
...
+
+        delete_volume(api_client, unique_name, wait_timeout)
+        delete_image(api_client, image_id, wait_timeout)
+
+    @pytest.mark.p0
+    @pytest.mark.dependency(name="create_image_url")
+    @pytest.mark.parametrize("image_name, image_url",
+                             [("qcow2-", qcow2_base_url + qcow2_image_name),
+                              ("iso-", iso_base_url + iso_image_name)])
+    def test_create_image_url(self, image_name, image_url, unique_name, api_client, wait_timeout):
```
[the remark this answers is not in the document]
bk201: @lanfon72 let's keep this to a future refactor, thanks!

harvester_e2e_tests/integration/test_images.py
TachunLin: Thanks for the review, move the function content of `update_by_input` and `get_export` to the test case blocks

harvester_e2e_tests/integration/test_images.py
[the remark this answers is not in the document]
TachunLin: Appreciated for the review and suggestion. Since if we update the shared image fixtures  by using the request options and config.yaml content. We may need to reorganize the structure in related test cases. 
If possible, can we keep this into future refactor. I will create another PR to optimize it. many thanks.

TachunLin:
Thanks for the review and suggestion. 
I have update according to review and verify the test suite can pass Jenkins run test jobs.

image

lanfon72:
image


The first failed case (_harvester_e2e_tests/apis/test_images.py::test_create_with_invalid_url_) is expected because we introduce retry mechanism.

when test on a used environment , `TestBackendImages::test_create_image_from_volume` failed as I mentioned in https://github.com/harvester/tests/pull/663#discussion_r1106140389

And Noah added test case `test_create_with_url` in https://github.com/harvester/tests/pull/673, I have no idea why we going to remove it.

TachunLin:
@lanfon72 , thanks for the check, I will examine the test result and update them accordingly on the used test environment. 

@noahgildersleeve , for the existing test case `test_create_with_url` in the `integration/test_images.py` file
Since the incoming test case `test_create_image_url` would also cover the test scenario of uploading opensuse image
Just want to confirm with you whether we can replace the existing image test case added by you or you'd prefer to keep both of them.

TachunLin:
Thanks for the review and checking, I have commited the following changes to
1. Resolve export image failure in an used environment contain existing images
2. Restore the test case `test_create_with_url` create by Noah.   

* And verified on the Jenkins pipeline `harvester-runtests` job 211 with `test_images` parameter filter.  
image

lanfon72 (approved):
In general LGTM

for the existing test case, It looks nothing will be changed but just update the layout, any reason for it?
image

TachunLin:
Really appreciate for the check and reminder. 
The reason to alter the layout is I add the `test_create_with_url` code block directly from the current test repository main web page , did not notice some lines and indent after paste to the editor. 

I just use the `raw` content to restore the `test_create_with_url` code block and verified on the Jenkins pipeline harvester-runtests job 212 again, the result looks fine

image

harvester_e2e_tests/integration/test_images.py
```diff
@@ -12,6 +18,346 @@
...
+    code, data = api_client.images.get(unique_name)
+    assert 200 == code, (code, data)
+    assert unique_name == data['metadata']['name']
+
+
+@ pytest.mark.p0
+class TestBackendImages:
+
+    @ pytest.mark.p0
+    @ pytest.mark.dependency(name="create_image_from_volume")
```
khushboo-rancher, line 142: I think space between @ and pytest is not required. It should be `@pytest.mark`

khushboo-rancher (commented):
LGTM, just a minor comment.

harvester_e2e_tests/integration/test_images.py
TachunLin: Appreciated for the check, remove the space between @ and pytest at line 138

khushboo-rancher (approved):

PR #673, opened 2022-12-23
Added in e2e tests for volume from URL and test skels

noahgildersleeve:
This adds 
- Create image from URL
   - params for that
- Create volume and image
   - This does not yet create a volume from an image.
   
   Still to do is add image id to volume spec in the data models


apiclient/harvester_api/managers.py
```diff
@@ -184,9 +184,9 @@ def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._get(path, raw=raw)
 
-    def create(self, name, volume_spec, namespace=DEFAULT_NAMESPACE, *, raw=False):
+    def create(self, name, volume_spec, namespace=DEFAULT_NAMESPACE, *, raw=False, image_id=None):
```
lanfon72, line 187: I thought we could let `image_id` be position argument would be more convenient:
```python
def create(self, name, volume_spec, namespace=DEFAULT_NAMESPACE, image_id=None, *, raw=False): ...
```

harvester_e2e_tests/apis/test_images.py
```diff
@@ -136,6 +137,56 @@ def test_create_with_reuse_display_name(self, api_client, unique_name, fake_imag
...
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to delete image {unique_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+    @pytest.mark.dependency(name="create_image_url")
+    def test_create_with_url(self, request, api_client, unique_name):
```
lanfon72, line 154: as the test case requires external resource, we should add it as integration test case.
to make the test steps more clear, we should prevent to use `request` besides other fixtures, we can make a fixture `opensuse_image` for it.

harvester_e2e_tests/conftest.py
```diff
@@ -269,6 +275,13 @@ def pytest_collection_modifyitems(config, items):
             if 'windows_vm' in item.keywords:
                 item.add_marker(skip_windows_vm)
 
+    if config.getoption('--opensuse-image-url') == '':
+        skip_create_image_url = pytest.mark.skip(reason=(
+            'openSUSE image not specified'))
+        for item in items:
+            if 'windows_vm' in item.keywords:
```
lanfon72, line 282: This line means any tests marks with `windows_vm` will be skipped if `opensuse-image-url` is empty, I think it is not reflect to the thing what we want to do.

harvester_e2e_tests/apis/test_volumes.py
```diff
@@ -119,3 +153,48 @@ def test_delete(self, api_client, unique_name, wait_timeout):
                 f"Failed to delete volume {unique_name} with {wait_timeout} timed out\n"
                 f"Still got {code} with {data}"
             )
+
+    def test_create_volume_backing_image(self, request, api_client, unique_name):
```
lanfon72, line 157: as this test case requires external resource, we should move it to `integration`

lanfon72 (changes requested):

harvester_e2e_tests/conftest.py
noahgildersleeve: Good catch. I'll fix that.

harvester_e2e_tests/apis/test_images.py
noahgildersleeve: removed external resource and changed to use opensuse_image fixture

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: removed request and just used the images fixture

noahgildersleeve:
Updated.

harvester_e2e_tests/apis/test_images.py
noahgildersleeve: Were you referring to the part? I removed that, but it just occurred to me that you might be referring to the external image (in this case opensuse)?

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: Were you referring to the part? I removed that, but it just occurred to me that you might be referring to the external image (in this case opensuse)?

harvester_e2e_tests/fixtures/images.py
```diff
@@ -11,7 +11,7 @@
                               "/Leap_15.3/images/openSUSE-Leap-15.3.x86_64-NoCloud.qcow2")
 
 
-@pytest.fixture(scope="session")
+@pytest.fixture(scope="class")
 def opensuse_image(request, api_client):
```
lanfon72, line 15: the fixture is shareable and will not be update in test session, so we can keep `session` scope.

harvester_e2e_tests/apis/test_volumes.py
```diff
@@ -119,3 +154,46 @@ def test_delete(self, api_client, unique_name, wait_timeout):
                 f"Failed to delete volume {unique_name} with {wait_timeout} timed out\n"
                 f"Still got {code} with {data}"
             )
+
+    def test_create_volume_backing_image(self, api_client, unique_name, opensuse_image):
```
lanfon72, line 158: let's move the test case into `integration/test_volumes.py`
the file contains some legacy test cases might need to be removed or rewrite.

lanfon72 (changes requested):

harvester_e2e_tests/conftest.py
noahgildersleeve: I removed that part.

apiclient/harvester_api/managers.py
noahgildersleeve: Fixed.

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: moved.

harvester_e2e_tests/fixtures/images.py
noahgildersleeve: Changed.

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: Changed

lanfon72 (changes requested):
Consider that `test_create_volume_backing_image` duplicating the image upload, then we could dependency mark for it:
```python
@pytest.mark.dependency(name="create_with_url")
def test_create_with_url(api_client, opensuse_image, unique_name, wait_timeout):
    # Only test image upload, verify the progress is 100, size is correct.

@pytest.mark.dependency(depends=["create_with_url"])
def test_create_volume_backing_image(api_client, opensuse_image, unique_name, wait_timeout):
    # `unique_name` is module scope, so we can easily know the image name uploaded in previous test case
    # then reuse it to create image volume
    # and check the volume state changed to *Active* after created.
```

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,47 @@ def _wait_for_volume_ready():
...
+
+    endtime = datetime.now() + timedelta(minutes=3)
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        image_conds = data.get('status', {}).get('conditions', [])
+        if len(image_conds) > 0:
+            break
+        sleep(3)
+
+    assert "Initialized" == image_conds[len(image_conds)-1].get("type")
```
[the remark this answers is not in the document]
irishgordo: and also something like:
```python
items.pop() == items[len(items)-1]
```
would work too :smile:

tox.ini
```diff
@@ -15,6 +15,7 @@ commands =
   {toxinidir}/terraform_test_artifacts/terraform_install.sh
   find . -type f -name "*.pyc" -delete
   pytest {posargs}
+allowlist_externals = {toxinidir}/*,bash,find
 whitelist_externals =
   bash
   find
```
irishgordo, line 21: Since we're using `allowlist_externals` - I believe we could now remove the `whitelist_externals`.
https://tox.wiki/en/latest/config.html#allowlist_externals
Since I think, `whitelist_externals` has been deprecated, but not 100% sure :sweat_smile: :smile: 
Since if we leave both, I'm not sure if one configuration overrides the other.

irishgordo (changes requested):

tox.ini
```diff
@@ -15,6 +15,7 @@ commands =
   {toxinidir}/terraform_test_artifacts/terraform_install.sh
   find . -type f -name "*.pyc" -delete
   pytest {posargs}
+allowlist_externals = {toxinidir}/*,bash,find
```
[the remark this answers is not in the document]
noahgildersleeve: Yeah. I think you might be right about the deprecation. I had to add the `allowlist_externals` to get it to run in my environment.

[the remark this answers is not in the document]
noahgildersleeve: Removed.

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,47 @@ def _wait_for_volume_ready():
...
+    code, image_data = api_client.images.create_by_url(unique_name, opensuse_image.url)
+    if code != 201:
+        raise AssertionError(
+            f"Failed to create image {unique_name} from URL got\n"
+            f"Creation got {code} with {image_data}"
+        )
+
+    assert 201 == code, (code, image_data)
+
+    endtime = datetime.now() + timedelta(minutes=3)
```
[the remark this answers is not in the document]
noahgildersleeve: Fixed.

harvester_e2e_tests/apis/test_images.py
```diff
@@ -136,6 +139,55 @@ def test_create_with_reuse_display_name(self, api_client, unique_name, fake_imag
...
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to delete image {unique_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
+    @pytest.mark.dependency(name="create_image_url")
+    def test_create_with_url(self, api_client, opensuse_image, unique_name):
```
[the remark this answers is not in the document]
noahgildersleeve: Moved.

harvester_e2e_tests/integration/test_volumes.py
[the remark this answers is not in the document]
noahgildersleeve: Fixed.

harvester_e2e_tests/apis/test_images.py
```diff
@@ -136,6 +139,55 @@ def test_create_with_reuse_display_name(self, api_client, unique_name, fake_imag
...
+
+        endtime = datetime.now() + timedelta(minutes=3)
+        while endtime > datetime.now():
+            code, data = api_client.images.get(unique_name)
+            image_conds = data.get('status', {}).get('conditions', [])
+            if len(image_conds) > 0:
+                break
+            sleep(3)
+
+        # assert len(image_conds) == 1, f"Got unexpected image conditions!\n{data}"
```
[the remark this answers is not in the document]
noahgildersleeve: Removed

harvester_e2e_tests/apis/test_images.py
```diff
@@ -136,6 +139,55 @@ def test_create_with_reuse_display_name(self, api_client, unique_name, fake_imag
...
+        assert "Initialized" == image_conds[len(image_conds)-1].get("type")
+        assert "True" == image_conds[len(image_conds)-1].get("status")
```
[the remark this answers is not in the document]
noahgildersleeve: Changed

tox.ini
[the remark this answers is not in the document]
noahgildersleeve: Removed

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,50 @@ def _wait_for_volume_ready():
...
+    code, data = api_client.volumes.create(unique_name, spec, image_id=image_id_temp)
+
+    # This is for checking that the volume was created
+    code, volume_data = api_client.volumes.get(unique_name)
+    assert 200 == code, (code, volume_data)
+    assert unique_name == volume_data['metadata']['name'], (code, volume_data)
+    assert image_id_temp == volume_data['id']
+
+    api_client.images.delete(unique_name)
+    api_client.volumes.delete(unique_name)
```
lanfon72, line 123: the order need to be reversed, Images can not be deleted if there have any volumes created from it.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: Excellent point. I'll change that.

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,50 @@ def _wait_for_volume_ready():
...
+    spec.size = "10Gi"
+    # spec.storage_cls = "longhorn-" + image_data["spec"]["displayName"]
+    image_id_temp = image_data["metadata"]["namespace"] + '/' + image_data["metadata"]["name"]
+    code, data = api_client.volumes.create(unique_name, spec, image_id=image_id_temp)
+
+    # This is for checking that the volume was created
+    code, volume_data = api_client.volumes.get(unique_name)
+    assert 200 == code, (code, volume_data)
+    assert unique_name == volume_data['metadata']['name'], (code, volume_data)
+    assert image_id_temp == volume_data['id']
```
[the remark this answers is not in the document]
noahgildersleeve: I looked into this and you can delete a volume that is created with a downloaded image if done through the API.  I'm going to look into it. If you do it through the UI you can't. Which probably means that I'm not doing it the way that it's being done via the UI.

noahgildersleeve:
I looked into why I could delete a volume before an image and you can delete a volume that is created with a downloaded image if done through the API.  I'm going to look into it. If you do it through the UI you can't. Which probably means that I'm not doing it the way that it's being done via the UI.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,49 @@
...
+
+    1. Create an image from URL.
+    2. Check for 201 response.
+    3. loop until the image has conditions.
+    4. Check if the image is intialzied and the status is true
+    5. Remove image
+    """
+    code, data = api_client.images.create_by_url(unique_name, opensuse_image.url)
+    if code != 201:
+        raise AssertionError(
```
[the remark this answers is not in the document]
noahgildersleeve: Changed.

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,50 @@ def _wait_for_volume_ready():
         timeout=request.config.getoption('--wait-timeout'))
     assert success, 'Timed out while waiting for volume to be ready.'
     return pvc_json
+
+
+@pytest.mark.volumes_p1
```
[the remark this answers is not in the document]
noahgildersleeve: Changed in all cases in the file.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: This was working but it shouldn't have been. I changed it and am looking into why it was still working.

noahgildersleeve:
I looked into it and  can't figure out why it's doing that. It's out of scope for this PR though.

harvester_e2e_tests/integration/test_volumes.py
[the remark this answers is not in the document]
noahgildersleeve: I looked into this and it was an issue with the storage_cls not being set properly in the volume spec. I fixed it and verified that it could be used to create a bootable VM

noahgildersleeve:
I looked more into this and it was an issue with the storage_cls not being set properly in the volume spec. I fixed it and verified that it could be used to create a bootable VM

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,49 @@
...
+            f"Failed to create image {unique_name} from URL got\n"
+            f"Creation got {code} with {data}"
+            )
+
+    assert 201 == code, (code, data)
+
+    endtime = datetime.now() + timedelta(minutes=3)
```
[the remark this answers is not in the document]
noahgildersleeve: Done

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,47 @@
...
+    'harvester_e2e_tests.fixtures.api_endpoints',
+    'harvester_e2e_tests.fixtures.api_client',
+    'harvester_e2e_tests.fixtures.volume',
+    'harvester_e2e_tests.fixtures.session',
+    'harvester_e2e_tests.fixtures.image',
+    'harvester_e2e_tests.fixtures.images',
+]
+
+
+@pytest.mark.images_p1
```
khushboo-rancher, line 15: Could you please add separate markers as in `p0` and `images` ?

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+
+    assert 201 == code, (code, image_data)
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        image_conds = data.get('status', {}).get('conditions', [])
+        if len(image_conds) > 0:
+            break
+        sleep(3)
```
khushboo-rancher, line 106: I would suggest to create a general variable for the wait time instead of hardcoded 3

khushboo-rancher (commented):
Thanks @noahgildersleeve !

In general this is looking good. Left minor comments

harvester_e2e_tests/integration/test_images.py
noahgildersleeve: Can do.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: That's an excellent point. I'll do that.

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+
+    # This is for checking that the volume was created
+    code, volume_data = api_client.volumes.get(unique_name)
+    code, image_data = api_client.images.get(unique_name)
+    assert 200 == code, (code, volume_data)
+    assert unique_name == volume_data['metadata']['name'], (code, volume_data)
+    assert volume_data['status']['phase'] == "Bound", (volume_data)
+    assert image_id_temp == volume_data['id'], (volume_data)
+    if not request.config.getoption('--do-not-cleanup'):
```
lanfon72, line 139: the configure option `do-not-cleanup` is deprecated and will be removed, so we should not use it.

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+            break
+        sleep(3)
+
+    assert "Initialized" == image_conds[-1].get("type")
+    assert "True" == image_conds[-1].get("status")
+    spec = api_client.volumes.Spec(1)
+    spec.size = "10Gi"
+    # This is required for the volume to be properly assigned to the image
+    spec.storage_cls = "longhorn-" + image_data["spec"]["displayName"]
+    image_id_temp = image_data["metadata"]["namespace"] + '/' + image_data["metadata"]["name"]
```
lanfon72, line 122: I think we can just use `image_id` instead, there is other variable use it.
and to concatenate strings, use f-string would be more readable:
```python
image_id_temp = image_data["metadata"]["namespace"] + '/' + image_data["metadata"]["name"]
# same as
image_id = f"{image_data['metadata']['namespace']}/{image_data['metadata']['name']}"

File: harvester_e2e_tests/integration/test_volumes.py [hunk 23]
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+        if image_conds[0]['reason'] == 'Imported':
+            break
+        sleep(3)
+
+    assert "Initialized" == image_conds[-1].get("type")
+    assert "True" == image_conds[-1].get("status")
+    spec = api_client.volumes.Spec(1)
+    spec.size = "10Gi"
+    # This is required for the volume to be properly assigned to the image
+    spec.storage_cls = "longhorn-" + image_data["spec"]["displayName"]
```
lanfon72, line 121: we can add `storage_class when creating the Spec like:
```python
spec = api_client.volumes.Spec(10, f"longhorn-{image_data['spec']['displayName']}")
```

harvester_e2e_tests/integration/test_volumes.py
lanfon72: @khushboo-rancher I think the while loop already explain the sleep in here is targeting to snooze a bit to do another check. :thinking: 

I think waiting loop will need variety snooze seconds which depends on its condition, so this should be fine to use plain integer inside.

In here, we should check the while loop exited without break, then the `image_conds` might be `[]`, the the code would becomes:
```python
    code, data = api_client.images.get(unique_name)
    image_conds = data.get('status', {}).get('conditions', [])
    if len(image_conds) > 0:
        break
    sleep(3)
assert image_conds > 0, "Failed to get image ..."
```
equals to
```python
    code, data = api_client.images.get(unique_name)
    image_conds = data.get('status', {}).get('conditions', [])
    if len(image_conds) > 0:
        break
    sleep(3)
else:
    raise AssertionError(f"Failed to get image, with wait {wait_timeout}s timed out")
```

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+    1. Create a new image from URL
+    2. Check that it is created succesffully.
+    3. Create a new volume with the image_id of the previous image
+    4. Check that the new image is created
+    5. Delete image and volume
+    """
+
+    code, image_data = api_client.images.create_by_url(unique_name, opensuse_image.url)
+    if code != 201:
+        raise AssertionError(
```
lanfon72, line 93: same as:
```python
assert 201 == code, "error messages"
```

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,47 @@
...
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        image_conds = data.get('status', {}).get('conditions', [])
+        if len(image_conds) > 0:
+            break
+        sleep(3)
+
+    assert "Initialized" == image_conds[-1].get("type")
+    assert "True" == image_conds[-1].get("status")
+    if not request.config.getoption('--do-not-cleanup'):
```
lanfon72, line 46: `do-not-cleanup` is deprecated.

harvester_e2e_tests/integration/test_images.py
```diff
@@ -0,0 +1,47 @@
...
+    4. Check if the image is intialzied and the status is true
+    5. Remove image
+    """
+    code, data = api_client.images.create_by_url(unique_name, opensuse_image.url)
+    assert 201 == code, (
+                f"Failed to create image {unique_name} from URL got\n"
+                f"Creation got {code} with {data}"
+    )
+    endtime = datetime.now() + timedelta(wait_timeout)
+    while endtime > datetime.now():
```
lanfon72, line 37: need to check the `image_conds` when while loop timed out.

lanfon72 (changes requested):

harvester_e2e_tests/apis/test_volumes.py
```diff
@@ -65,12 +81,21 @@ def test_create_without_name(self, api_client):
...
+        """
         spec = api_client.volumes.Spec(1)
         code, data = api_client.volumes.create(unique_name, spec)
 
         assert 201 == code, (code, data)
 
     def test_get(self, api_client, unique_name):
+        """
+        1. Runs a get command on the volume created in the previous test
+        note: this will fail if you run it by itslef
```
irishgordo, line 96:
```suggestion
        note: this will fail if you run it by itself
```

harvester_e2e_tests/apis/test_volumes.py
```diff
@@ -84,6 +109,11 @@ def test_get(self, api_client, unique_name):
         assert unique_name == data['metadata']['name'], (code, data)
 
     def test_update_size(self, api_client, unique_name, wait_timeout):
+        """
+        1. Updates the volume created in the previous test to 10GB
+        note: this will fail if you run it by itslef
```
irishgordo, line 114:
```suggestion
        note: this will fail if you run it by itself
```

harvester_e2e_tests/apis/test_volumes.py
```diff
@@ -104,6 +134,11 @@ def test_update_size(self, api_client, unique_name, wait_timeout):
         assert 200 == code, (f"Failed to update volume with error: {code}, {data}")
 
     def test_delete(self, api_client, unique_name, wait_timeout):
+        """
+        1. Deletes the volume created in the previous test
+        note: this will fail if you run it by itslef
```
irishgordo, line 139:
```suggestion
        note: this will fail if you run it by itself
```

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -72,3 +74,68 @@ def _wait_for_volume_ready():
...
+        image_conds = data.get('status', {}).get('conditions', [])
+        if len(image_conds) > 0:
+            break
+        sleep(3)
+    # This waits for the import to finish
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        image_conds = data.get('status', {}).get('conditions', [])
+        if image_conds[0]['reason'] == 'Imported':
```
irishgordo, line 112:
```suggestion
        if len(image_conds) != 0 and image_conds[0]['reason'] == 'Imported':
```
There could be a chance that if we're waiting in the while loop, since we set `image_conds` to ultimately be an empty list, that by the type the evaluation of:
```python
if image_conds[0]['reason'] == 'Imported':
```
It could fire back an error of:
```python
IndexError: list index out of range
```
Which would bubble up and fail the test since we're not currently catching the Exception.
When it might just mean that the test wouldn't fail, because we were just waiting for the conditions - and it just happens in the first iteration we didn't get any, so we set as an empty list - but maybe 3-4 iterations down the line we would get 'conditions' and we wouldn't end up setting it as an empty list.

I know earlier in the test we have that logic block of:
```python

        code, data = api_client.images.get(unique_name)
        image_conds = data.get('status', {}).get('conditions', [])
        if len(image_conds) > 0:
            break
        sleep(3)
```
So I guess it could also be written something like:
```suggestion
        if len(image_conds) > 0 and image_conds[0]['reason'] == 'Imported':
```
too :smile:

irishgordo (changes requested):

harvester_e2e_tests/integration/test_volumes.py
khushboo-rancher: I was just thinking to have some standard variable which is used across the tests. So, if we have any performance change in getting result, we'll get to know from our test.
I'm fine with hardcoded also, this was just a nit.

noahgildersleeve: I'm adding in a new fixture with a sleep timeout how we have a wait-timeout. It will pull from the config the same way.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: Good point. I'll change those to be more concise

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: Excellent point. I'll make the change. That's more concise and  less prone to timeouts.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: removed.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: changed.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: Changed.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: changed.

harvester_e2e_tests/integration/test_images.py
noahgildersleeve: removed.

harvester_e2e_tests/integration/test_images.py
noahgildersleeve: It shouldn't time out now. If it does I'll add in some more error checking.

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: Changed.

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: Changed.

harvester_e2e_tests/apis/test_volumes.py
noahgildersleeve: Changed.

harvester_e2e_tests/integration/test_volumes.py
noahgildersleeve: Changed.

irishgordo (approved):

lanfon72 (approved):
LGTM

PR #675, opened 2022-12-23
test: add get_processors.sh

FrankYang0529:


lanfon72:
@FrankYang0529 I think we don't need to add the script to get `processors` counts, they already in the response of `api.host.get()` :thinking:

lanfon72:
close as the script is not necessary.
we can leverage `host_shell` to execute shell commands on host.

PR #681, opened 2023-01-06
[e2e] add initial integration test cases for network

TachunLin:
#### Add initial network backend test including: 

- [x] 1. Test management network connectivity of VM

   - test_mgmt_network_connection

- [x] 2. Test vlan network connectivity of VM

   - test_vlan_network_connection

- [x] 3. Test VLAN network after reboot 

   - test_reboot_vlan_connection

- [x] 4. Edit vm network, from VLAN to management 

   - test_vlan_to_mgmt_connection

- [x] 5. Change management network to external VLAN

   - test_mgmt_to_vlan_connection

- [x] 6.  Delete single network from multiple networks vm 

    - test_delete_vlan_from_multiple

#### Prerequisite of environment
1. At least **two** network interface, one for management and another for external vlan network
2. The external vlan network have DHCP server configured to assign IP address to VM 

#### Configuration setup
Edit config.yaml and provide the following attribute (take bare machine for example)
1. `vlan-id` to create vlan network
2. `vlan-nic` to create cluster network and config
3. `vlan-cidr` to determine which IP subent range for vlan network test cases

  ```
  # VLAN ID, will invoke the tests depended on external networking if not set to -1.
  vlan-id: 1216
  
  # Physical NIC for VLAN. Default is "harvester-mgmt"
  vlan-nic: 'ens1f1'
  
  # VLAN CIDR, indicate the vlan network subnet range"
  vlan-cidr: '10.84.99.0/24'
  ```
  
#### Additional context 
To test it on our local ipxe-example cluster, we can the following config   
  ```
  vlan-id: 1
  vlan-nic: 'ens5'
  vlan-cidr: '192.168.122.0/24'
  
  ```


TachunLin:
* Add test case of: 

  - test_reboot_vlan_connection
  - test_mgmt_to_vlan_connection
  - test_vlan_to_mgmt_connection

* Add delete cluster network 
* Add vlan-cidr in config.yaml

TachunLin:
Add test case of:
* test_delete_vlan_from_multiple 
  - Delete single network from multiple networks vm

TachunLin:
Verified can pass all test suite on `dolphin` bare machine triggered from local terminal.

TachunLin:
Validate can run the TestBackendNetwork test suite on Jenkins runtest job #252
image

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+
+        # Ping vlan ip address
+        vlan_ip = ip_addresses[0]
+
+        command = ['/usr/bin/ping', '-c', '50', vlan_ip]
+
+        result = subprocess.check_output(command, shell=False, encoding="utf-8")
+
+        assert result.find("64 bytes from {0}".format(vlan_ip)) > 0, (
+            'Failed to ping VM management IP %s' % (vlan_ip))
```
lanfon72, line 521: I think the error message is not clear, it didn't point out what the type of NIC using by VM, and from where to do the ping.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+        client_ip = request.config.getoption('--endpoint').strip('https://')
+
+        # Check image exists
+        code, data = api_client.images.get(image_name)
+
+        if code == 404:
+            create_image_url(api_client, image_name, qcow2_image_url, wait_timeout)
+
+        # Update AllowTcpForwarding for ssh jumpstart
+        tcp = "sudo sed -i 's/AllowTcpForwarding no/AllowTcpForwarding yes/g' /etc/ssh/sshd_config"
```
lanfon72, line 183: The command expose the node to be a jumphost, it would better named `jumphost` rather than `tcp`.
And we should always revert the change when leaving.

Or you can simply use `host_shell` fixture in `api_client`.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+        code, data = api_client.vms.get(unique_name)
+        spec = spec.from_dict(data)
+
+        # Switch to vlan network
+        spec.mgmt_network = True
+
+        # Update VM spec
+        code, data = api_client.vms.update(unique_name, spec)
+
+        spec.delete_vlan_network(spec, vlan_network['id'])
```
lanfon72, line 594: steps in here is not align to test plan, we would expect the VM will have only 1 network.
so we should remove mgmt and add VLAN in the same time, it would be:
```python
        # Switch to vlan network
        spec.mgmt_network = True

        # Update VM spec   -------- don't update VM otherwise the VM will have 2 NICs
        # code, data = api_client.vms.update(unique_name, spec)

        spec.delete_vlan_network(spec, vlan_network['id'])
```

and it could be simplify as:
```python
spec.networks = []  # clear all NICs
spec.mgmt_network = True
```

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+
+        while endtime > datetime.now():
+            code, data = api_client.vms.vm_instance(unique_name)
+            if 'interfaces' in data['status']:
+                interfaces_data = data['status']['interfaces']
+                ip_addresses = []
+                if 'ipAddress' in data['status']['interfaces'][0]:
+                    ip_addresses.append(interfaces_data[0]['ipAddress'])
+
+                    if '10.52' in ip_addresses[0]:
```
lanfon72, line 643: I'm doubt whether we need to check this

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+
+        # Check until VM ip address exists
+        check_vm_ip_exists(api_client, unique_name, wait_timeout)
+
+        # get data from running VM and transfer to spec
+        code, data = api_client.vms.get(unique_name)
+        spec = spec.from_dict(data)
+
+        # Switch to vlan network
+        spec.add_network("nic-1", "default/" + vlan_network['id'])
```
lanfon72, line 721: I thought steps before here in not align to test plan, it described:
```
On a VM with both an external VLAN and a management VLAN delete the external VLAN via the web form
```
but not tell us to add network after the VM started.
I would suggest to add 2 NICs before create the VM.

And the comment looks weird, we are adding new NIC, not switch to.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+                f"Still got {code} with {data}"
+            )
+
+        # Ping management ip address
+        mgmt_ip = ip_addresses[0]
+
+        ping_command = "ping -c 50 {0}".format(mgmt_ip)
+
+        _stdout, _stderr = self.ssh_client(
+            client, client_ip, "rancher", "p@ssword", ping_command, wait_timeout)
```
lanfon72, line 804: It pretty weird, you already constant `node_user` and `node_password`, and still hard coded them in here?

lanfon72 (changes requested):
@TachunLin could you:
1. Update to use existing fixtures in `api_client` instead of use `request.config.getoption` in test cases
2. Update string formatting to f-string if it is possible

something like:
```python
assert result.find("64 bytes from {0}".format(vlan_ip)) > 0, (
            'Failed to ping VM management IP %s' % (vlan_ip))
```
should be simplify as:
```python3
assert result.find(f"64 bytes from {vlan_ip}") > 0, (
    f"Failed to ping VM management IP {vlan_ip}"
)
```

And please check all codes, I didn't add comments for duplicated line, so you would need to check things like `10 minutes`, `wait_timeout` and so on.

lanfon72 (changes requested):
@TachunLin could you:
1. Update to use existing fixtures in `api_client` instead of use `request.config.getoption` in test cases
2. Update string formatting to f-string if it is possible

something like:
```python
assert result.find("64 bytes from {0}".format(vlan_ip)) > 0, (
            'Failed to ping VM management IP %s' % (vlan_ip))
```
should be simplify as:
```python3
assert result.find(f"64 bytes from {vlan_ip}") > 0, (
    f"Failed to ping VM management IP {vlan_ip}"
)
```

And please check all codes, I didn't add comments for duplicated line, so you would need to check things like `10 minutes`, `wait_timeout` and so on.

lanfon72:
The parameter `name` in dependency is not required and we would better not assign it to point out the test case will not be rely on other test cases.

lanfon72:
Please also add two marks for those test cases:
```python
@pytest.mark.virtualmachines
@pytest.mark.networks
```

harvester_e2e_tests/integration/test_networks.py
bk201: @TachunLin Please help change to constant and also check if there are similar cases too. Thanks!

TachunLin: Appreciated for the review. 
I will update the hard coded at line 804 to the existing constant `node_user` and `node_password`. 
And also check no hard code added as the function parameter in test cases.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+import pytest
+import paramiko
+import shlex
+
+
+pytest_plugins = [
+    "harvester_e2e_tests.fixtures.api_client",
+    'harvester_e2e_tests.fixtures.network',
+    'harvester_e2e_tests.fixtures.vm',
+    "harvester_e2e_tests.fixtures.virtualmachines"
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review. I will remove the following legacy fixtures.
1. 'harvester_e2e_tests.fixtures.network' 
2. 'harvester_e2e_tests.fixtures.vm'

And verify the entire test suite again to ensure have no impact

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+    yield data
+
+    api_client.networks.delete(network_name)
+    api_client.clusternetworks.delete_config(vlan_nic)
+    api_client.clusternetworks.delete(vlan_nic)
+
+
+@pytest.fixture(scope='session')
+def qcow2_image_url():
+    base_url = "https://downloadcontent-us1.opensuse.org/repositories/" \
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review.
I will use the `image_info` fixtures in `fixtures/images.py` to parse the image_url for create image from url 
  ```
@pytest.fixture(params=["image_opensuse"], scope="module")
def image_info(request):
    return request.getfixturevalue(request.param)
  ```

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+
+
+        Steps:
+        1. Create a new VM
+        2. Make sure that the network is set to the management network with masquerade as the type
+        3. Wait until the VM boot in running state
+        4. Check can ping VM
+        5. Check can SSH to VM
+        """
+        wait_timeout = request.config.getoption('--wait-timeout')
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review and reminder. 
I have removed all the `wait_timeout = request.config.getoption('--wait-timeout')` in every test case and set wait_timeout fixture instead.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+        client_ip = request.config.getoption('--endpoint').strip('https://')
+
+        # Check image exists
+        code, data = api_client.images.get(image_name)
+
+        if code == 404:
+            create_image_url(api_client, image_name, qcow2_image_url, wait_timeout)
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review.
I will use the `image_info` fixtures in fixtures/images.py to parse the image_url for create image from url 

  ```
  create_image_url(api_client, image_name, image_url, wait_timeout)
  ```

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+        # Get VM interface ipAddresses
+        code, data = api_client.vms.vm_instance(unique_name)
+        interfaces_data = data['status']['interfaces']
+        for interface in interfaces_data:
+            ip_addresses = interface['ipAddresses']
+
+        vlan_ip = ip_addresses[0]
+
+        vlan_subnet_prefix = vlan_ip.rsplit('.', 1)[0]
+        vlan_subnet_prefix_store["value"] = vlan_subnet_prefix
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review. 
The purpose to store the vlan_subnet information is we need to use the vlan subnet information e.g `10.84.99` to identify when we switch from mgmt to vlan network in test case 5: `test_mgmt_to_vlan_connection`

This can ensure the exposed ip address 10.84.99.x did actually match the subnet information. 
Thus we could determine it actually switch from management subnet (10.52) to vlan subnet (10.84.99) accordingly. 

Without this information, we need extra code effort to access the the gitlab repository 
`harvester-baremetal-info/-/blob/master/pcloud001/dhcp.yml` to retrieve the `vm_subnet` field

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+    def test_vlan_network_connection(self, api_client, request, client, unique_name,
+                                     vlan_network, vlan_subnet_prefix_store):
+        """
+        Manual test plan reference:
+        https://harvester.github.io/tests/manual/network/validate-network-external-vlan/
+
+
+        Steps:
+        1. Create an external VLAN network
+        2. Create a new VM already with the vlan network
```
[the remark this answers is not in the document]
TachunLin: Appreciated for the review. 

Sorry for the misleading information on steps 2 in test skeleton: `Create a new VM already with the vlan network` 
In this case I actually create only one VM and add the external VLAN to it. 

The step2 should be updated to **Create a new VM and add the external vlan network** which related to the line of code 
```
spec.add_network("default", vlan_network['id'])
```
Thank you for the check.

harvester_e2e_tests/integration/test_networks.py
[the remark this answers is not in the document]
lanfon72: What if the mask number is not fully divided by eight?
When you use `"192.168.10.1".rsplit(".", 1)[0]` to get the **vlan_subnet_prefix**, which means the prefix is `192.168.10`, but do we guarantee that the CIDR is `192.168.10.0/24`?
What if CIDR is `192.168.10.0/16` or even `/15`, `/25`, `/26`?

And according to
https://github.com/harvester/tests/pull/681/files#diff-e30c67d488529228b4fa2f07596daadf626451f5114bc2a6c8622c4e25e87f04R504
when you use
```python
if vlan_subnet_prefix in ip_addresses[0]: ...
```
consider the case:
```python
vlan_subnet_prefix = '10.52.1'
ip_addresses = ["10.10.52.1"]
```
Is the case should be *True* in the if condition?


> Without this information, we need extra code effort to access the the gitlab repository
harvester-baremetal-info/-/blob/master/pcloud001/dhcp.yml to retrieve the vm_subnet field

We didn't declare the mgmt network should always be `10.52.x.x/y`, so why we have to check it by its ip address? (and the test plan didn't mentioned that?)
When you look into the manual test plan, **Expected Results** described clearly: Check the VM is able to access from external network, but nothing about its IP address.

If we are  going to check the **interface** comes from which NIC, the name field is more useful, consider the schema:
```python
 'interfaces': [{'infoSource': 'domain, guest-agent',
                 'interfaceName': 'enp1s0',
                 'ipAddress': '10.52.1.23',
                 'ipAddresses': ['10.52.1.23'],
                 'mac': '52:54:00:8d:1f:56',
                 'name': 'default'}]
```
the `name` field is reflect to the name you defined for the NIC, when we adding a NIC to a VM, we are always give it a name, so we can simply check interface's name to know it is from which NIC.

apiclient/harvester_api/managers.py
```diff
@@ -749,6 +749,10 @@ def softreboot(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
         params = dict(action="softreboot")
         return self._create(path, raw=raw, params=params)
 
+    def vm_instance(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
```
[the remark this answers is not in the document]
TachunLin: Thanks for the reminder and review. 

I try to replace every occurrence of `api_client.vms.vm_instance(unique_name)` to `code, data = api_client.vms.get_status(unique_name)` in the test suite. 

And run the entire network test suite on dolphin cluster from local machine. 
It works well and can pass all test cases
image

I will also remove the `vm.vm_instance` function in apiclient/managers.py

```
def vm_instance(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
        path = self.VMI_fmt.format(uid=f"/{name}", ns=namespace, VM_API=self.API_VERSION)
        return self._get(path, raw=raw, **kwargs)
```

harvester_e2e_tests/integration/test_networks.py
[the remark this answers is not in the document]
TachunLin: Verified again can pass entire suite on the dolphin cluster after removing the `network` and `vm` fixtures.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+    'harvester_e2e_tests.fixtures.network',
+    'harvester_e2e_tests.fixtures.vm',
+    "harvester_e2e_tests.fixtures.virtualmachines"
+]
+
+
+image_name = "suse"
+vm_credential = {"user": "opensuse", "password": "123456"}
+node_user = "rancher"
+node_password = "p@ssword"
```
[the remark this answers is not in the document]
TachunLin: I added the node password in `host-password` of config.yaml. 
Use the configure option to retrieve the password for all test cases
  ```
  node_password = request.config.getoption("--host-password")
  ```
And verified can pass all test cases on dolphin cluster 

Thanks for the reminder.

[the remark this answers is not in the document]
lanfon72: Could you please update the PR (push commits) before update comments?
Otherwise It would always confusing me that received the mail but code is not changed.

I would suggest you can just update PR and click *Re-request review*, then I can check the newer changed directly and mark it resolved or add more comments in further. (it would also reduce mails flooding my mail box)

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+
+        # SSH to management ip address and execute command
+        _stdout, _stderr = self.ssh_jumpstart(
+            client, mgmt_ip, client_ip, node_user, node_password,
+            vm_credential["user"], vm_credential["password"], "ls")
+
+        stdout = _stdout.read().decode('ascii').strip("\n")
+
+        assert stdout.find("bin") == 0, (
+            'Failed to ssh to VM management IP %s' % (mgmt_ip))
```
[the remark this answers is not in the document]
khushboo-rancher: As the management network is usually used for VMs whose traffic only flows inside the cluster. We should check both the cases that 
1. Vms with management network should be sshable within the cluster.
2. Vms with management network should be NOT be sshable externally.

@TachunLin Could you add the 2nd case as well.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,840 @@
...
+        mgmt_ip = ip_addresses[0]
+        ping_command = "ping -c 50 {0}".format(mgmt_ip)
+
+        _stdout, _stderr = self.ssh_client(
+            client, client_ip, node_user, node_password, ping_command, wait_timeout)
+
+        stdout = _stdout.read().decode('ascii').strip("\n")
+
+        assert stdout.find("64 bytes from {0}".format(mgmt_ip)) > 0, (
+            'Failed to ping VM management IP %s' % (mgmt_ip))
```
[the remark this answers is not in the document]
khushboo-rancher: As the management network is usually used for VMs whose traffic only flows inside the cluster. We should check both the cases that

Vms with management network should be sshable within the cluster.
Vms with management network should be NOT be sshable externally.

@TachunLin Could you add the 2nd case as well.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,895 @@
...
+
+        # Update VM spec
+        code, data = api_client.vms.update(unique_name, spec)
+
+        # Remove external vlan network from spec
+        net_uid = vlan_network['id']
+        spec.networks = [net for net in spec.networks
+                         if net_uid != net["network"].get('multus', {}).get('networkName')]
+
+        code, data = api_client.vms.update(unique_name, spec)
```
bk201, line 643: Does a non-2xx status raise an exception? If it doesn't, we should always check the response code.
Please check the rest of the changes as well; thanks!

TachunLin:
Validate the latest update can pass the TestBackendNetwork test suite on Jenkins runtest job 353

  image

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        if code == 404:
+            create_image_url(api_client, image_name, image_url, wait_timeout)
+
+        spec = api_client.vms.Spec(1, 2, mgmt_network=False)
+        spec.user_data += cloud_user_data.format(password=vm_credential["password"])
+
+        # Create VM
+        spec.add_image(image_name, "default/" + image_name)
+        spec.add_network("default", vlan_network['id'])
```
lanfon72, line 360: The **mgmt** network will use `default` as the name, this will make NIC name conflict.

and the network_id should contains its `namespace`.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+        # Check until VM ip address exists
+        check_vm_ip_exists(api_client, unique_name, wait_timeout)
+
+        # Get VM interface ipAddresses
+        code, data = api_client.vms.get_status(unique_name)
+        assert 200 == code, (f"Failed to get specific vm content: {code}, {data}")
+
+        interfaces_data = data['status']['interfaces']
+        for interface in interfaces_data:
+            ip_addresses = interface['ipAddresses']
```
lanfon72, line 377: so, when:
```python
interfaces_data = [dict(ipAddresses=["127.0.0.1"]), dict(ipAddresses=["192.168.0.1"])]
```
`ip_addresses` will not get all IPs, is what we want?

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        vlan_ip = ip_addresses[0]
```
lanfon72, line 379: You didn't check the IP is from the (VLAN) NIC, how can you say it is `vlan_ip`?

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        Steps:
+        1. Create an external VLAN network
+        2. Create a new VM already with the vlan network
+        3. Check can ping external VLAN IP
+        4. Reboot VM
+        5. Ping VM during reboot
+        6. Check can't ping VM during reboot
+        7. Check the VM should reboot
+        8. Ping VM during after reboot
```
lanfon72, line 342: Do you mean `ping VM after rebooted`?

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        assert result.find(f"64 bytes from {vlan_ip}") > 0, (
+            f"Failed to ping VM external vlan IP {vlan_ip} "
+            f"on vlan interface from external node")
+
+        # Restart VM
+        code, data = api_client.vms.restart(unique_name)
+        assert 204 == code, (f"Failed to reboot vm with error: {code}, {data}")
+
+        # Check VM start in Starting state
```
lanfon72, line 395: the following lines is not align to the test step.
what we want is:
```
4. Reboot VM
5. Ping VM during reboot
6. Check can't ping VM during reboot
7. Check the VM should reboot
8. Ping VM during after reboot
```

we should do `ping VM` after `restart` submitted (step 5),
then check the ping status (step 6),
then check VM is started and get IP (step 7)
....

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        # get data from running VM and transfer to spec
+        code, data = api_client.vms.get(unique_name)
+        assert 200 == code, (f"Failed to get specific vm content: {code}, {data}")
+
+        spec = spec.from_dict(data)
+
+        # Switch to vlan network
+        spec.mgmt_network = False
+        # spec.add_network("default", "default/" + vlan_network['id'])
```
lanfon72, line 491: remove the code if you don't use it.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+        # Get VM interface ipAddresses
+        code, data = api_client.vms.get_status(unique_name)
+        assert 200 == code, (f"Failed to get specific vm content: {code}, {data}")
+
+        interfaces_data = data['status']['interfaces']
+        for interface in interfaces_data:
+            ip_addresses = interface['ipAddresses']
+
+        # Check VM start in running state
```
lanfon72, line 516: is the step duplicated?

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+        # Switch to vlan network
+        spec.mgmt_network = True
+
+        # Update VM spec
+        # code, data = api_client.vms.update(unique_name, spec)
+        # assert 200 == code, (f"Failed to update specific vm with spec: {code}, {data}")
+
+        # Remove external vlan network from spec
+        net_uid = vlan_network['id']
+        spec.networks = [net for net in spec.networks
```
lanfon72, line 653: In here, we can simply do:
 ```python
spec.networks = []
spec.mgmt_network = True
```
rather than
```python
spec.mgmt_network = True
spec.networsk = [...]
```

https://github.com/harvester/tests/pull/681#discussion_r1188698247

harvester_e2e_tests/integration/test_networks.py
lanfon72: https://github.com/harvester/tests/pull/681#discussion_r1202701376

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+        assert stdout.find("bin") == 0, (
+            f"Failed to ssh to VM management IP {mgmt_ip} "
+            f"on management interface from Harvester node")
+
+        # Check should not SSH to management ip address from external host
+        command = ['/usr/bin/ssh', '-o', 'ConnectTimeout=5', mgmt_ip]
+        try:
+            result = subprocess.check_output(
+                command, stderr=subprocess.STDOUT, shell=False, encoding="utf-8")
+        except subprocess.CalledProcessError as e:
```
lanfon72, line 746: I think we should use `pytest.raises` in here.
https://docs.pytest.org/en/7.1.x/how-to/assert.html#assertions-about-expected-exceptions

And why we can't use ssh client to test the connection?

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,919 @@
...
+
+    @pytest.mark.p0
+    @pytest.mark.dependency(name="delete_vlan_connection")
+    def test_delete_vlan_from_multiple(self, api_client, request, client, unique_name,
+                                       image_info, vlan_network, wait_timeout):
+        """
+        Manual test plan reference:
+        https://harvester.github.io/tests/manual/network/delete-vlan-network-form/
+
+        Steps:
```
lanfon72, line 764: steps and codes are not align to manual test plan, could you check it again?

lanfon72 (changes requested):
Please also check not resolved comments in https://github.com/harvester/tests/pull/681#pullrequestreview-1418384757

and https://github.com/harvester/tests/pull/681#issuecomment-1540335258

TachunLin:
Thanks for the review. 

Verified the latest update can pass the `TestBackendNetwork` test suite on Jenkins runtest job **366**
image

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,954 @@
...
+            result = e.output
+
+        assert "connect to host {0} port 22: No route to host".format(mgmt_ip) in result, (
+            f"Failed: Should not be able to SSH to VM management IP {mgmt_ip}"
+            f"on management interface from Harvester node")
+
+        # cleanup VM
+        delete_vm(api_client, unique_name, wait_timeout)
+
+        # Revert AllowTcpForwarding for ssh jumpstart
```
lanfon72, line 246: The reversion should always be executed which means it should be placed in a `finally` clause.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,954 @@
...
+            result = e.output
+
+        assert "connect to host {0} port 22: No route to host".format(mgmt_ip) in result, (
+            f"Failed: Should not be able to SSH to VM management IP {mgmt_ip}"
+            f"on management interface from Harvester node")
+
+        # cleanup vm
+        delete_vm(api_client, unique_name, wait_timeout)
+
+        # Revert AllowTcpForwarding for ssh jumpstart
```
lanfon72, line 774: The reversion should always be executed which means it should be placed in a `finally` clause.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,954 @@
...
+        """
+        vip = request.config.getoption('--endpoint').strip('https://')
+
+        node_password = request.config.getoption("--host-password")
+
+        # Check image exists
+        code, data = api_client.images.get(image_opensuse.name)
+
+        if code == 404:
+            create_image_url(api_client, image_opensuse.name, image_opensuse.url, wait_timeout)
```
lanfon72, line 169: I think what I'm saying is, use a fixture like:

https://github.com/harvester/tests/blob/cd34f0e5006d466d467ab359ac3e26e5b6f0d7e4/harvester_e2e_tests/integration/test_vm_functions.py#L74-L99

to make the image be the part of setup.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,954 @@
...
+    def test_vlan_network_connection(self, api_client, request, client, unique_name,
+                                     image_opensuse, vlan_network, wait_timeout):
+        """
+        Manual test plan reference:
+        https://harvester.github.io/tests/manual/network/validate-network-external-vlan/
+
+
+        Steps:
+        1. Create an external VLAN network
+        2. Create a new VM and add the external vlan network
```
lanfon72, line 265: The test steps is not align to the manual test plan, what the manual test plan asking:
> Make sure that the network is set to the external VLAN with bridge as the type

which means there will have **only 1 NIC**, and it is external VLAN.

harvester_e2e_tests/integration/test_networks.py
```diff
@@ -0,0 +1,954 @@
...
+        assert stdout.find("bin") == 0, (
+            f"Failed to ssh to VM external vlan IP {vlan_ip}"
+            f"on vlan interface from external host")
+
+        # cleanup vm
+        delete_vm(api_client, unique_name, wait_timeout)
+
+    @pytest.mark.p0
+    # @pytest.mark.dependency(name="reboot_vlan_connection",
+    #                         depends=["vlan_network_connection"])
```
lanfon72, line 334: If it is not used, remove instead of comment them.

lanfon72 (changes requested):
https://github.com/harvester/tests/pull/681#pullrequestreview-1418384757
> 2. Update string formatting to f-string if it is possible
> 
> something like:
> ```python
> assert result.find("64 bytes from {0}".format(vlan_ip)) > 0, (
>             'Failed to ping VM management IP %s' % (vlan_ip))
> ```
```python
ping_command = "ping -c 50 {0}".format(mgmt_ip)
```
> should be simplify as:
> ```python3
> assert result.find(f"64 bytes from {vlan_ip}") > 0, (
>     f"Failed to ping VM management IP {vlan_ip}"
> )
> ```




https://github.com/harvester/tests/pull/681#issuecomment-1540335258
> Please also add two marks for those test cases:
> ```python
> @pytest.mark.virtualmachines
> @pytest.mark.networks
> ```

TachunLin:
Thanks for the review. 

Verified the latest update can pass the `TestBackendNetwork` test suite on Jenkins runtest job **387**
image

bk201:
@lanfon72 @khushboo-rancher I think David already addressed most comments and requests. Should we merge this PR to increase our test coverage first? We don't need the PR to be perfect.

If we see issues in daily tests, @TachunLin will continue improving it.

bk201 (approved):

lanfon72 (approved):

lanfon72:
Fix conflict then we can continue

PR #689, opened 2023-01-11
feat: Additional VM Tests

irishgordo:
* test bogus vlan vm creation
* test vm creation with outlandish resources
* test vm updating machine types

Resolves: feat/vm-testcases

### DIFF
--- a/harvester_e2e_tests/integration/test_vm_functions.py
+++ b/harvester_e2e_tests/integration/test_vm_functions.py
@@ -1,11 +1,12 @@
-import re
-from time import sleep
+from datetime import datetime, timedelta
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from datetime import datetime, timedelta
+from time import sleep
 
-import yaml
+import json
+import re
 import pytest
+import yaml
 from paramiko.ssh_exception import ChannelException
 
 pytest_plugins = [
@@ -13,6 +14,9 @@ pytest_plugins = [
     "harvester_e2e_tests.fixtures.images",
     "harvester_e2e_tests.fixtures.virtualmachines"
 ]
+
+# GLOBAL Vars:
+MAX = 999999
 
 
 @pytest.fixture(scope="session")
@@ -32,6 +36,60 @@ def kubeconfig_file(api_client):
         f.write(kubeconfig)
         f.seek(0)
         yield Path(f.name)
+
+
+@pytest.fixture(scope='module')
+def bogus_vlan_net(request, api_client):
+    """bogus vlan network fixture (no dhcp) on mgmt network
+
+    Args:
+        request (FixtureRequest): https://docs.pytest.org/en/7.1.x/_modules/_pytest/fixtures.html#FixtureRequest # noqa
+        api_client (HarvesterAPI): HarvesterAPI client
+
+    Yields:
+        dict: created bogus network attachment definition dictionary
+    """
+    original_vlan_id = request.config.getoption('--vlan-id')
+
+    existing_vm_net_code, existing_vm_net_data = api_client.networks.get()
+    assert existing_vm_net_code == 200, 'we should be able to fetch vm networks from harvester'
+    existing_vm_net_list = existing_vm_net_data.get('items', [])
+    vlans_to_exclude = set()
+    vlans_to_exclude.add(1)
+    for existing_vm_net in existing_vm_net_list:
+        existing_vm_net_config = existing_vm_net.get('spec', {}).get('config', '{}')
+        assert existing_vm_net_config != '{}', 'existing vm net should exist'
+        existing_vm_net_config_dict = json.loads(existing_vm_net_config)
+        assert existing_vm_net_config_dict.get(
+            'vlan', 0) != 0, 'we should be able to get the vlan off the config'
+        existing_vm_net_vlan = existing_vm_net_config_dict.get('vlan')
+        vlans_to_exclude.add(existing_vm_net_vlan)
+
+    if original_vlan_id != -1:
+        vlans_to_exclude.add(original_vlan_id)
+
+    vlan_ids = set(range(2, 4095))  # 4094 is the last, 1 should always be excluded.
+    code, data = api_client.networks.get()
+    for net in data['items']:
+        config = json.loads(net['spec'].get('config', '{}'))
+        if config.get('vlan'):
+            try:
+                # try to remove the key, but VLAN may be used in both 'mgmt'
+                # and other cluster network(s) so it might have already been removed
+                vlan_ids.remove(config['vlan'])
+            except KeyError:
+                print(f"key, {config['vlan']} was already removed by another cluster network")
+
+    vlan_id = vlan_ids.pop()  # Remove and return an arbitrary set element.
+    vm_network_name = f'bogus-net-{vlan_id}'
+    code, data = api_client.networks.create(vm_network_name, vlan_id)
+    assert 201 == code, (
+        f"Failed to create N.A.D. {vm_network_name} with error {code}, {data}"
+    )
+
+    yield data
+
+    api_client.networks.delete(vm_network_name)
 
 
 @pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
         )
 
 
+@pytest.mark.p0
+@pytest.mark.virtualmachines
```
lanfon72, line 143: this should be marked as negative test

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+                'not_running': None,
+                'unschedulable': None
+            }
+            for status_condition in data.get('status', {}).get('conditions', []):
+                if status_condition.get('reason') == 'GuestNotRunning':
+                    check_dict['not_running'] = True
+                if status_condition.get('reason') == 'Unschedulable':
+                    check_dict['unschedulable'] = True
+            assert check_dict["not_running"] is True, "the vm shouldn't be running"
+            assert check_dict["unschedulable"] is True, "the vm shouldn't be schedulable"
```
lanfon72, line 216: we can just check that the vmi is entering 'Unschedulable' condition as:
```python
    code, data = api_client.vms.get_status(unique_name_for_vm)
    conds = data['status'].get('conditions', [{}])
    if conds[-1].get('reason') == 'Unschedulable':
        break
    sleep(3)
else:
    raise AssertionError(
        f"VM {unique_name_for_vm} entered unexpected condition\n"
        f"Phase: {data['status'].get('phase')}\n"
        f"VM resources: {overall_vm_obj}\n"
        f"conditions: {conds}\n"
        f"API Status({code}): {data}"
)
```
and delete vm should not be included in the same while loop.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
         )
 
 
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+@pytest.mark.parametrize("resource", [("cpu", 999999), ("mem", 999999), ("disk", 999999),
+                                      ("mem_and_cpu", 999999), ("mem_cpu_and_disk", 999999)],
+                         ids=['cpu', 'mem', 'disk', 'mem_and_cpu', 'mem_cpu_and_disk'])
+def test_create_vm_no_available_resources(resource, api_client, image_id, wait_timeout):
+    """Creates a VM with outlandish resources for varying elements (purposefully negative test)
```
lanfon72, line 148: docstring should add refer to manual test case

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+            f"Phase: {data.get('status', {}).get('phase','')}\t"
+            f"Status: {data.get('status', {})}\n"
+            f"API Status({code}): {data}"
+        )
+
+
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+def test_vm_with_bogus_vlan(api_client, image_id, unique_vm_name, wait_timeout, bogus_vlan_net):
+    """test building a VM with a VM (VLAN) Network has a bogus VLAN ID (no DHCP)
```
lanfon72, line 425: docstring should add refer to manual test case

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+            f"Failed to create VM({cpu} core, {mem} RAM) with errors:\n"
+            f"Phase: {data.get('status', {}).get('phase','')}\t"
+            f"Status: {data.get('status', {})}\n"
+            f"API Status({code}): {data}"
+        )
+
+
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+def test_vm_with_bogus_vlan(api_client, image_id, unique_vm_name, wait_timeout, bogus_vlan_net):
```
lanfon72, line 424: this should be marked as negative test

lanfon72 (changes requested):
the `wait_timeout` is inputted by configuration and could be very small, so we should make every waiting while not nested in others.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+        overall_vm_obj['disk'] = resource_val
+    if resource_name == 'mem_and_cpu':
+        overall_vm_obj["mem"] = resource_val
+        overall_vm_obj["cpu"] = resource_val
+    if resource_name == 'mem_cpu_and_disk':
+        overall_vm_obj["mem"] = resource_val
+        overall_vm_obj["cpu"] = resource_val
+        overall_vm_obj["disk"] = resource_val
+
+    vm = api_client.vms.Spec(overall_vm_obj['cpu'], overall_vm_obj['mem'])
```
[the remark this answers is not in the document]
irishgordo: that's a great suggestion, I'll introduce that global variable :smile: !

apiclient/harvester_api/models.py
```diff
@@ -70,7 +70,7 @@ class VMSpec:
     # SSH Key
 
     def __init__(self, cpu_cores, memory, description="", reserved_mem="", os_type="",
-                 mgmt_network=True, guest_agent=True):
+                 mgmt_network=True, guest_agent=True, machine_type=None, networks=None):
```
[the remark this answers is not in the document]
irishgordo: If we were to add a network, it wouldn't be able to be the `default` network - is there a way to tell the VM via configuration to "only" use that network and not the other `default` network?

I was thinking that if we just had the ability to set the Networks on the VMSpec directly during initalization of the object we wouldn't need to worry about the juggling of a VM with two NICs.

Since the test was intending to be negative, if the VM only had one NIC and that NIC was tied directly to a VLAN that purposefully didn't have DHCP access.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -30,6 +31,36 @@ def kubeconfig_file(api_client):
...
+    Yields:
+        dict: created bogus network attachment definition dictionary
+    """
+    original_vlan_id = request.config.getoption('--vlan-id')
+    # don't create network if VLAN is not correctly specified
+    if original_vlan_id == -1:
+        return
+    # just increase the original vlan by one,
+    # optimistic attempt that it isn't another valid vlan
+    vlan_id = original_vlan_id + 1
```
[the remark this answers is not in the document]
irishgordo: I was thinking maybe the fixture's focus would be tied to just the `mgmt` network.
I did add some additional logic to do a `set()` operation to ensure that we wouldn't be creating a bogus-vlan with a VLAN that was already in use.  Instead it now picks something at random with some exclusions.
Something like:
```python
    original_vlan_id = request.config.getoption('--vlan-id')

    existing_vm_net_code, existing_vm_net_data = api_client.networks.get()
    assert existing_vm_net_code == 200, 'we should be able to fetch vm networks from harvester'
    existing_vm_net_list = existing_vm_net_data.get('items', [])
    vlans_to_exclude = set()
    for existing_vm_net in existing_vm_net_list:
        existing_vm_net_config = existing_vm_net.get('spec', {}).get('config', '{}')
        assert existing_vm_net_config != '{}', 'existing vm net should exist'
        existing_vm_net_config_dict = json.loads(existing_vm_net_config)
        assert existing_vm_net_config_dict.get(
            'vlan', 0) != 0, 'we should be able to get the vlan off the config'
        existing_vm_net_vlan = existing_vm_net_config_dict.get('vlan')
        vlans_to_exclude.add(existing_vm_net_vlan)

    if original_vlan_id != -1:
        vlans_to_exclude.add(original_vlan_id)

    # create a random vlan that 'shouldnt' have dhcp
    # create based on excluding some values and from within a range as a set operation
    vlan_id = choice(list(set([vlan for vlan in range(1, 4094)]) - vlans_to_exclude))
```

With that logic, is there still a potential for collision? 
Is it fine if the fixture is directly tied to only working with creating VMNetworks on the `mgmt` network?

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+    3. check for conditions of guest not running and vm being unschedulable
+    4. delete vm and volumes
+
+    Expected Result:
+    - building vm with outlandish resource requests to be successful
+    - asserting that the status condition of the vm that is built to not be running
+    - asserting that the status condition of the vm that is built to be unschedulable
+    - assert deleting vm and volumes to be successful
+    """
+    unique_name_for_vm = _adjusted_unique_name()
```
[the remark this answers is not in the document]
irishgordo: great call out :smile: :+1:  - updated

harvester_e2e_tests/integration/test_vm_functions.py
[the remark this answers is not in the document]
irishgordo: updated :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: I did forget that marker, great catch :smile: - updated :+1:

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+            f"API Status({code}): {data}"
+        )
+
+
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+@pytest.mark.parametrize("machine_types", [("pc", "q35"), ("q35", "pc")],
+                         ids=['pc_to_q35', 'q35_to_pc'])
+def test_update_vm_machine_type(api_client, image_id, unique_vm_name, wait_timeout, machine_types):
+    """Create a VM with machine type then update to another
```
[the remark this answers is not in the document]
irishgordo: updated to include those elements in the docstring :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: updated to include those elements in the docstring :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: updated to include those elements in the docstring :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: I did forget that marker, great catch smile - updated +1

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -99,6 +139,406 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+        "network": dict(name='default')
+    }
+    net['iface']["bridge"] = dict()
+    net['network']['multus'] = dict(
+        networkName=bogus_vlan_net['metadata']['namespace'] +
+        '/'+bogus_vlan_net['metadata']['name']
+    )
+    network_list = []
+    network_list.append(net)
+    vm = api_client.vms.Spec(cpu, mem, networks=network_list)
```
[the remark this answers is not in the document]
irishgordo: updated :smile:

apiclient/harvester_api/models.py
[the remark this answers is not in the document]
lanfon72: we can update `networks` after the Spec object initialized, the code snippet:
```python
>>> vm = api.vms.Spec(1, 2)
>>> vm.networks
[{'iface': {'model': 'virtio', 'name': 'default', 'masquerade': {}}, 'network': {'name': 'default', 'pod': {}}}]
>>> 
>>> vm.add_network('nic-1', 'the-net-uid')
{'iface': {'model': 'virtio', 'name': 'nic-1', 'bridge': {}}, 'network': {'name': 'nic-1', 'multus': {'networkName': 'the-net-uid'}}}
>>> vm.networks
[{'iface': {'model': 'virtio', 'name': 'default', 'masquerade': {}}, 'network': {'name': 'default', 'pod': {}}}, {'iface': {'model': 'virtio', 'name': 'nic-1', 'bridge': {}}, 'network': {'name': 'nic-1', 'multus': {'networkName': 'the-net-uid'}}}]
>>> vm.mgmt_network = False
>>> vm.networks
[{'iface': {'model': 'virtio', 'name': 'nic-1', 'bridge': {}}, 'network': {'name': 'nic-1', 'multus': {'networkName': 'the-net-uid'}}}]
```

if you want to let the `mgmt-network` not being the first NIC, you can use:
```python
>>> vm = api.vms.Spec(1, 2, mgmt_network=False)
>>> vm.networks
[]
>>> vm.add_network('nic-1', 'the-net-uid')
{'iface': {'model': 'virtio', 'name': 'nic-1', 'bridge': {}}, 'network': {'name': 'nic-1', 'multus': {'networkName': 'the-net-uid'}}}
>>> vm.mgmt_network = True
>>> vm.networks
[{'iface': {'model': 'virtio', 'name': 'nic-1', 'bridge': {}}, 'network': {'name': 'nic-1', 'multus': {'networkName': 'the-net-uid'}}}, {'iface': {'model': 'virtio', 'name': 'default', 'masquerade': {}}, 'network': {'name': 'default', 'pod': {}}}]
```

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
 
 @pytest.mark.p0
 @pytest.mark.virtualmachines
-@pytest.mark.dependency(depends=["minimal_vm"])
+@pytest.mark.negative
+@pytest.mark.parametrize("resource", [dict(cpu=MAX), dict(mem=MAX), dict(disk=MAX),
+                                      dict(mem=MAX, cpu=MAX), dict(mem=MAX, cpu=MAX, disk=MAX)],
+                         ids=['cpu', 'mem', 'disk', 'mem-and-cpu', 'mem-cpu-and-disk'])
+def test_create_vm_no_available_resources(request, resource, api_client, image_id,
```
lanfon72, line 160: fixture `request` not used.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+    vm = api_client.vms.Spec(overall_vm_obj['cpu'], overall_vm_obj['mem'])
+    vm.add_image("disk-0", image_id, size=overall_vm_obj.get('disk'))
+    code, data = api_client.vms.create(unique_name_for_vm, vm)
+    assert 201 == code, (code, data)
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.vms.get_status(unique_name_for_vm)
+        if 200 == code and len(data.get('status', {}).get('conditions', [])) > 1:
+            check_dict = {
```
lanfon72, line 207: consider this with type hint:
```python
check_dict : dict = { ... }
```

usually, we will not add any type name when defining variables, as a ducking type language, we would more consider it's behavior like:
```python
checks = dict(...)  # with type hint: checks : dict = dict(...)
# `s` means it is iterable
checks['something'] = 'good'
# slice with string usually means it would fit Mapping protocol
```

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+                'not_running': None,
+                'unschedulable': None
+            }
+            for status_condition in data.get('status', {}).get('conditions', []):
+                if status_condition.get('reason') == 'GuestNotRunning':
+                    check_dict['not_running'] = True
+                if status_condition.get('reason') == 'Unschedulable':
+                    check_dict['unschedulable'] = True
+            assert check_dict["not_running"] is True, "the vm shouldn't be running"
+            assert check_dict["unschedulable"] is True, "the vm shouldn't be schedulable"
```
lanfon72, line 217: We should check all conditions, so:
```python
assert all(check_dict.values()), (
    "the VM miss condition:\n"
    " and".join(k for k, v in check_dict.items() if not v)
)
```

then we can optimized to readability:
```python
checks = dict(GuestNotRunning=False, Unschedulable=False)
for condition in data['status']['conditions']:
    if condition.get('reason') in checks:
        checks[condition['reason']] = True

assert all(checks.values()), (
    "The VM miss condition:\n"
    " and ".join(k for k, v in check_dict.items() if not v)
)
```

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+                    fails.append(
+                        (vol_name, f"Failed to delete\nStatus({code}): {data}"))
+
+            assert not fails, (
+                f"Failed to delete VM({unique_vm_name})'s volumes with errors:\n"
+                "\n".join(f"Volume({n}): {r}" for n, r in fails)
+            )
+            break
+        sleep(3)
+    else:
```
lanfon72, line 264: The `else` scope should be placed in Line 218.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+            f"{overall_vm_obj.get('mem')} RAM, \n"
+            f"{overall_vm_obj.get('disk')} DISK) with errors:\n"
+            f"Phase: {data.get('status', {}).get('phase')}\t"
+            f"Status: {data.get('status')}\n"
+            f"API Status({code}): {data}"
+        )
+
+
+@ pytest.mark.p0
+@ pytest.mark.virtualmachines
```
lanfon72, line 276: space is not necessary.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+                if status_condition.get('reason') == 'Unschedulable':
+                    check_dict['unschedulable'] = True
+            assert check_dict["not_running"] is True, "the vm shouldn't be running"
+            assert check_dict["unschedulable"] is True, "the vm shouldn't be schedulable"
+            code, data = api_client.vms.delete(unique_name_for_vm)
+            assert 200 == code, (code, data)
+
+            spec = api_client.vms.Spec.from_dict(data)
+
+            while endtime > datetime.now():
```
lanfon72, line 223: following codes (includes lines start from 218) are the part of tear down, we should not let them be include in the first while loop.

Otherwise we will not know how long the while duration is.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+def test_update_vm_machine_type(api_client, image_id, unique_vm_name, wait_timeout, machine_types):
+    """Create a VM with machine type then update to another
+
+    Prerequisite:
+        Setting opensuse-image-url set to a valid URL for
+        an opensuse image.
+
+    Manual Test Doc(s):
+        - https://harvester.github.io/tests/manual/virtual-machines/create-new-vm-with-a-machine-type-pc/ # noqa
+        - https://harvester.github.io/tests/manual/virtual-machines/create-new-vm-with-a-machine-type-pc/ # noqa
```
lanfon72, line 288: duplicate line, it looks we should reference another document that machine type is q35.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+        raise AssertionError(
+            f"Failed to create VM({cpu} core, {mem} RAM) with errors:\n"
+            f"Phase: {data.get('status', {}).get('phase','')}\t"
+            f"Status: {data.get('status', {})}\n"
+            f"API Status({code}): {data}"
+        )
+
+
+@ pytest.mark.p0
+@ pytest.mark.negative
```
lanfon72, line 428: space is not necessary

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+    net = {
+        "iface": dict(model='virtio', name='default'),
+        "network": dict(name='default')
+    }
+    net['iface']["bridge"] = dict()
+    net['network']['multus'] = dict(
+        networkName=f"{bvn['metadata']['namespace']}/{bvn['metadata']['name']}"
+    )
+    network_list = []
+    network_list.append(net)
```
lanfon72, line 472: `net` and `network_list` is not used.

lanfon72 (changes requested):
Please remember to remove the space after the `@` symbol.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -30,6 +37,51 @@ def kubeconfig_file(api_client):
...
+    # create based on excluding some values and from within a range as a set operation
+    vlan_id = choice(list(set([vlan for vlan in range(1, 4094)]) - vlans_to_exclude))
+
+    vm_network_name = f'bogus-net-{vlan_id}'
+    code, data = api_client.networks.create(vm_network_name, vlan_id)
+    assert 201 == code, (
+        f"Failed to create N.A.D. {vm_network_name} with error {code}, {data}"
+    )
+
+    yield data
```
lanfon72, line 80: if we will just need `data['metadata']['namespace']` and `data['metadata']['name']`, we can only yield them out.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_vm_functions.py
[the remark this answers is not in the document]
lanfon72: reply in https://github.com/harvester/tests/pull/689#discussion_r1098885556

apiclient/harvester_api/models.py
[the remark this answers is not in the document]
irishgordo: That's a good mention :+1: :smile: 
I've updated it, since I think we can still check for a bogus vlan if it has the mgmt-network being the first vlan and just adding an additional NIC that doesn't have DHCP resolution.

apiclient/harvester_api/models.py
```diff
@@ -70,7 +70,7 @@ class VMSpec:
     # SSH Key
 
     def __init__(self, cpu_cores, memory, description="", reserved_mem="", os_type="",
-                 mgmt_network=True, guest_agent=True):
+                 mgmt_network=True, guest_agent=True, machine_type=None):
```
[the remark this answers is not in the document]
irishgordo: That's a good point, I've gone ahead and simplified it :+1: 
Thanks for the link to the builder pattern as well, that's a great article :smile:

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -101,14 +153,414 @@ def test_minimal_vm(api_client, image_id, unique_vm_name, wait_timeout):
...
+def test_create_vm_no_available_resources(request, resource, api_client, image_id,
+                                          wait_timeout, unique_vm_name):
+    """Creates a VM with outlandish resources for varying elements (purposefully negative test)
+
+    Prerequisite:
+        Setting opensuse-image-url set to a valid URL for
+        an opensuse image.
+
+    Manual Test Doc(s):
+        - N/A
```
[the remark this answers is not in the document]
irishgordo: Thank you for the links :smile:  - I think I was having trouble finding those documents :sweat_smile: - I've gone ahead and updated that

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: great catch :smile:  - I've removed the unused pytest request fixture :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: That's a great point :smile:  - I think perhaps I had named it that to keep track in scope while scaffolding the test.

It might be great to bring in type hinting with some tests at some point.
Something like:
```python
def headline(text: str, centered: bool = False) -> str:
    if not centered:
        return f"{text.title()}\n{'-' * len(text)}"
    else:
        return f" {text.title()} ".center(50, "o")
```
Having a quick way to know at a function level what the args types are can be pretty useful :smile: 

I've gone ahead and renamed it from an explicit data type to just a simple checks variable :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: I like the simplicity with that, I've gone ahead and refactored that portion :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: I've pulled that tear down outside the scope of the first while loop - great catch :+1: :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: ahh, yes, duplicate indeed :sweat_smile:  - good catch :smile:  - I've gone ahead and updated the secondary link to point to the appropriate document for q35 :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: updated to remove the space - might have been a hiccup in my linting/formatter :sweat_smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: I've cleaned those variables up.  Seem to have been hanging around from a previous iteration on the test function.  Good catch :smile: :+1:

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -1,15 +1,22 @@
...
-from datetime import datetime, timedelta
+from time import sleep
 
 import pytest
 
 pytest_plugins = [
     "harvester_e2e_tests.fixtures.api_client",
-    "harvester_e2e_tests.fixtures.images"
+    "harvester_e2e_tests.fixtures.images",
+    'harvester_e2e_tests.fixtures.network'
```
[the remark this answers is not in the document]
irishgordo: good catch, :+1:  - I've gone ahead and update this :smile:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: It might be nice to yield back the whole complete object from that fixture - if in the future more checks need to utilize other properties.

We could possibly build another dict with those elements and yield that to trim things back, like:
```python
vlan_info = {
  "namespace": data['metadata']['namespace'],
  "name": data['metadata']['name'],
}
yield vlan_info
```

I was thinking it might just be easier to yield back `data` entirely. 
So I left it for now - but I'm open to change that too :smile:

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -30,6 +37,51 @@ def kubeconfig_file(api_client):
...
+            'vlan', 0) != 0, 'we should be able to get the vlan off the config'
+        existing_vm_net_vlan = existing_vm_net_config_dict.get('vlan')
+        vlans_to_exclude.add(existing_vm_net_vlan)
+
+    if original_vlan_id != -1:
+        vlans_to_exclude.add(original_vlan_id)
+
+    # create a random vlan that 'shouldnt' have dhcp
+    # create based on excluding some values and from within a range as a set operation
+    vlan_id = choice(list(set([vlan for vlan in range(1, 4094)]) - vlans_to_exclude))
```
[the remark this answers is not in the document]
irishgordo: `set` operations can be pretty neat but I do see that it is more readable with this approach.  So I've updated that :smile: . 
Additionally, I added a `try/except` section, since collecting all VLANs over multiple cluster-networks, there's a chance a cluster-network other than 'mgmt' also is using a VLAN that 'mgmt' just happens to use.  If that's the case then when it loops over to try to remove it, since it's seen the VLAN more than once, we would hit a `KeyError`.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -1,15 +1,20 @@
-from time import sleep
+# noqa: E501
+import json
+from datetime import datetime, timedelta
 from pathlib import Path
 from tempfile import NamedTemporaryFile
-from datetime import datetime, timedelta
+from time import sleep
```
lanfon72, line 6: It looks we Import json but modify other import's order, I think we can just do `import json` in the first or the last would be enough.

And I have no idea why we should skip `E501` in here.

lanfon72 (changes requested):
@irishgordo it looks there still have many changes not be updated, did you forget to do `git push`? :thinking:

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: yeah, that's a great catch, I've gone ahead and removed the ignoring of `E501` - I'm not sure if that was there for some reason prior, perhaps I added it when testing something :sweat_smile:  - I've also updated the placement of the `import json` within the file :smile:

irishgordo:
I did on runtests-186 notice:
```
Failed | harvester_e2e_tests/integration/test_vm_functions.py::test_create_vm_no_available_resources[disk] | 902.86 |  
-- | -- | -- | --
Failed | harvester_e2e_tests/integration/test_vm_functions.py::test_update_vm_machine_type[pc_to_q35] | 0.10 |  
Failed | harvester_e2e_tests/integration/test_vm_functions.py::test_update_vm_machine_type[q35_to_pc] | 0.09 |  
Failed | harvester_e2e_tests/integration/test_vm_functions.py::test_vm_with_bogus_vlan | 0.12
```

irishgordo:
re-running with runtests-201

lanfon72 (changes requested):
Please rebase main before update it.
There still have several nested while loop need to be update.

irishgordo:
Noticing that via Jenkins runs where the underlying infrastructure is a spinning-disk, it seems that the 'disk' based param run seems to always fail... 

I believe that could be something to be iterated on, because everytime the test with that param, runs on an SSD or NVMe backed storage cluster either Virtualized or Bare-Metal it seems to pass, yet on the infrastructure that powers Jenkins, it's a spinning disk and seems to possibly be related to it never passing, that spinning disk on Jenkins underlying infra also doesn't meet the minimum requirements for IOPS I believe...

Screenshot from 2023-04-11 10-43-05
Screenshot from 2023-04-11 10-40-53

irishgordo:
harvester-runtests:#217 - does pass, after removing the 'disk' test that breaks on the non-ssd/non-nvme, low iops 
Screenshot from 2023-04-11 11-48-52

per, [etcd-docs-on-disks](https://etcd.io/docs/v3.5/op-guide/hardware/#disks):
> When possible, back etcd’s storage with a SSD. A SSD usually provides lower write latencies and with less variance than a spinning disk, thus improving the stability and reliability of etcd. 

I believe it's a RAID 5 setup with, [Hewlet-Packer-Enterprise-based-10k-SAS-HDD](https://www.amazon.com/HP-EG0600FCVBK-600GB-10K-RPM/dp/B0153IN0YQ) on the underlying infrastructure.

we could omit the `disk` based test for now since it will fail on the underlying infrastructure and simply re-introduce that python test param when we have dedicated non-hdd based storage to support the test.

wdyt @khushboo-rancher , @lanfon72  ?

khushboo-rancher:
@irishgordo Let's not omit the disk based test and let it fail on infra which doesn't have Harvester minimum resource requirement. 
If we run these tests on local ssd or in future we move to ssd they will work.

irishgordo:
@khushboo-rancher thanks for mentioning that - I agree, that sounds good to me too :smile:  - :+1:  - I will re-introduce that param for the test case and push up those changes.
Should I add a comment above the case, to mention that we "expect" that test to fail?

khushboo-rancher:
Sure

lanfon72 (approved):
In general LGTM, there have 2 comments commented last time, but it will not affect test results.

lanfon72:
I'm thinking we might need to adjust the order of test cases, consider `TestVMOperations` is focusing on basic VM functions, other test cases would better be placed after it.

irishgordo:
> I'm thinking we might need to adjust the order of test cases, consider `TestVMOperations` is focusing on basic VM functions, other test cases would better be placed after it.

I'll go ahead and re-arrange the structure so that `TestVMOperations` takes place all of the other test cases introduced in this PR :smile: :+1:

khushboo-rancher (approved):

khushboo-rancher:
@irishgordo Can you please fix the Flake8 check?

PR #693, opened 2023-01-12
test: add host integration

FrankYang0529:
Run on `harvester-runtests`. Jod ID is `136`.

### DIFF
--- a/harvester_e2e_tests/integration/test_hosts.py
+++ b/harvester_e2e_tests/integration/test_hosts.py
@@ -8,9 +8,119 @@ pytest_plugins = [
 ]
 
 
+@pytest.fixture(scope="session")
+def focal_image_url():
+    return "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img"  # noqa
+
+
+@pytest.fixture(scope="class")
+def focal_image(api_client, unique_name, focal_image_url, wait_timeout):
+    code, data = api_client.images.create_by_url(unique_name, focal_image_url)
+    assert 201 == code, (
+        f"Failed to upload focal image with error: {code}, {data}"
+    )
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        if 'status' in data and 'progress' in data['status'] and \
+                data['status']['progress'] == 100:
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Image {unique_name} can't be ready with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    namespace = data['metadata']['namespace']
+    name = data['metadata']['name']
+
+    yield dict(ssh_user="ubuntu", id=f"{namespace}/{name}")
+
+    api_client.images.delete(name, namespace)
+
+
+@pytest.fixture(scope="class")
+def focal_vm(api_client, unique_name, focal_image, wait_timeout):
+    vm_spec = api_client.vms.Spec(1, 1)
+    vm_spec.add_image('disk-0', focal_image['id'])
+    code, data = api_client.vms.create(unique_name, vm_spec)
+    assert 201 == code, (
+        f"Failed to create VM {unique_name} with error: {code}, {data}"
+    )
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.vms.get(unique_name)
+        if data.get('status', {}).get('ready', False):
+            code, data = api_client.vms.get_status(unique_name)
+            if data['status']['conditions'][-1]['status'] == 'True':
+                break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Can't find VM {unique_name} with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    data['name'] = data['metadata']['name']
+    data['namespace'] = data['metadata']['namespace']
+    yield data
+
+    volume_name = ""
+    for volume in data['spec']['volumes']:
+        if volume['name'] == 'disk-0':
+            volume_name = volume['persistentVolumeClaim']['claimName']
+    api_client.vms.delete(unique_name)
+    api_client.volumes.delete(volume_name)
+
+
+@pytest.fixture(scope="class")
+def available_node_names(api_client):
+    status_code, nodes_info = api_client.hosts.get()
+    assert status_code == 200, f"Failed to list nodes with error: {nodes_info}"
+
+    node_names = []
+    for node_info in nodes_info.get('data', []):
+        is_ready = False
+        for condition in node_info.get('status', {}).get('conditions', []):
+            if condition.get('type', "") == "Ready" and \
+                    condition.get('status', "") == "True":
+                is_ready = True
+                break
+
+        if is_ready and not node_info.get('spec', {}).get('unschedulable', False):
+            node_names.append(node_info['metadata']['name'])
+
+    return node_names
+
+
+@pytest.fixture(scope="class")
+def vm_force_reset_policy(api_client):
+    # update vm-force-reset-policy to 10s, so we don't need to wait 300s to force remove VM
+    code, data = api_client.settings.get("vm-force-reset-policy")
+    assert 200 == code, (
+        f"Failed to get vm-force-reset-policy setting with error: {code}, {data}")
+    original_value = data.get("value", data['default'])
+
+    updates = {"value": '{"enable":true,"period":10}'}
+    code, data = api_client.settings.update("vm-force-reset-policy", updates)
+    assert 200 == code, (
+        f"Failed to update vm-force-reset-policy setting with error: {code}, {data}"
+    )
+
+    yield data
+
+    # teardown
+    updates = {"value": original_value}
+    api_client.settings.update("vm-force-reset-policy", updates)
+
+
 @pytest.mark.hosts
 @pytest.mark.p1
 @pytest.mark.host_management
+@pytest.mark.dependency(name="host_poweroff")
 def test_host_poweroff_state(api_client, host_state, wait_timeout):
     """
     Test the hosts are the nodes which make the cluster
@@ -44,6 +154,7 @@ def test_host_poweroff_state(api_client, host_state, wait_timeout):
 @pytest.mark.hosts
 @pytest.mark.p1
 @pytest.mark.host_management
+@pytest.mark.dependency(name="host_poweron")
 def test_host_poweron_state(api_client, host_state, wait_timeout):
     _, nodes_info = api_client.hosts.get()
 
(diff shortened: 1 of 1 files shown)


apiclient/harvester_api/managers.py
```diff
@@ -730,3 +730,11 @@ def softreboot(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         params = dict(action="softreboot")
         return self._create(path, raw=raw, params=params)
+
+
+class VirtualMachineInstanceManager(BaseManager):
+    PATH_fmt = "v1/harvester/kubevirt.io.virtualmachineinstances/{ns}{uid}"
+
+    def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
```
lanfon72, line 738: this method has been implemented as `api.vms.get_status`, consider that we will not let user to manipulate VMI directly, so we will not add VMI Manager.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -8,6 +8,66 @@
...
+        if 'status' in data and 'progress' in data['status'] and \
+                data['status']['progress'] == 100:
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Image {unique_name} can't be ready with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+    sleep(10)
```
lanfon72, line 35: do we have any reason to delay in here? :thinking:

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -8,6 +8,66 @@
...
+
+    yield dict(ssh_user="ubuntu", id=f"{namespace}/{name}")
+
+    api_client.images.delete(name, namespace)
+
+
+@pytest.fixture(scope="class")
+def focal_vm(api_client, unique_name, focal_image, wait_timeout):
+    vmSpec = api_client.vms.Spec(1, 1)
+    vmSpec.add_image('disk-0', focal_image['id'])
```
lanfon72, line 48: snake case

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+
+@pytest.mark.hosts
+@pytest.mark.p0
+def test_verify_host_info(api_client):
+    status_code, nodes_info = api_client.hosts.get()
+
+    assert status_code == 200, f"Failed to list nodes with error: {nodes_info}"
+    assert len(nodes_info['data']) >= 1, "Incorrect hosts amount"
+
+    for node in nodes_info.get('data', []):
```
lanfon72, line 140: we already check the `node_info['data']` is not empty, so we don't need the default value in here.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+
+        cpus = node.get('status', {}).get('capacity', {}).get('cpu', 0)
+        assert 0 != cpus, (
+            f"Node {node_name} cpu should not be zero"
+        )
+
+        mems = node.get('status', {}).get('capacity', {}).get('memory', "0Ki")
+        mems = "".join(c for c in mems if c.isdigit())
+        mems = mems or "0"
+        assert 0 != int(mems, 10), (
```
lanfon72, line 152: We need to check every nodes with every resources, so we can not assert in the for-loop directly.
In here, we should check them in another way, such like:
https://github.com/harvester/tests/blob/241d35b577e4a4d322a669a1c2f9c7e21ad57e22/harvester_e2e_tests/integration/test_vm_functions.py#L442-L468

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+class TestHostTriggerVmMigrate:
+    @pytest.mark.dependency()
+    def test_prerequisite(self, api_client):
+        status_code, nodes_info = api_client.hosts.get()
+
+        assert status_code == 200, f"Failed to list nodes with error: {nodes_info}"
+        assert len(nodes_info['data']) > 1, "Please use multiple hosts to test vm migration"
+
+        for node_info in nodes_info.get('data', []):
+            for condition in node_info.get('status', {}).get('conditions', []):
```
lanfon72, line 168: prerequisite is not a independent test item, so we would not make it as a test. (you can let it be a fixture)

for test trigger VM migration, we will just need `2` nodes to be ready, so we don't need to check all nodes.
I would suggest to change this method into fixture and return a list contains name of ready nodes.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+                    )
+            assert False is node_info.get('spec', {}).get('unschedulable', False), (
+                f"Node {node_info['metadata']['name']} is unschedulable"
+            )
+
+    @pytest.mark.dependency(depends=["TestHostTriggerVmMigrate::test_prerequisite"])
+    def test_maintenance_mode(self, api_client, focal_vm, wait_timeout):
+        vm_name = focal_vm['metadata']['name']
+        vm_namespace = focal_vm['metadata']['namespace']
+        code, data = api_client.vmis.get(name=vm_name, namespace=vm_namespace)
```
lanfon72, line 181: If we will just need VM's name and namespace, we could update fixture to return them directly, then we can make the code looks like:
```python
code, data = api_client.vms.get_status(focal_vm['name'], focal_vm['namespace'])
```
or in further:
```python
code, data = api_client.vms.get_status(**focal_vm)
```

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+            if data.get('status', {}).get('phase', "") == "Running":
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"VM {vm_namespace}/{vm_name} can't be Running with {wait_timeout} timed out\n"
+                f"Got error: {code}, {data}"
+            )
+
+        # teardown
```
lanfon72, line 204: We need to check that the VM's host is updated and not be hosted on `src_host`

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+        vm_name = focal_vm['metadata']['name']
+        vm_namespace = focal_vm['metadata']['namespace']
+        code, data = api_client.vmis.get(name=vm_name, namespace=vm_namespace)
+        assert 200 == code, (
+            f"Can't get VMI {vm_namespace}/{vm_name} with error: {code}, {data}"
+        )
+        src_host = data['status']['nodeName']
+
+        # update vm-force-reset-policy to 10s, so we don't need to wait 300s to force remove VM
+        code, data = api_client.settings.get("vm-force-reset-policy")
```
lanfon72, line 221: This would mix the test a bit up with the setting, to make sure the setting will be tear down gracefully, we should use fixture for it.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +127,156 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+                f"Got error: {code}, {data}"
+            )
+
+        # teardown
+        code, data = api_client.hosts.maintenance_mode(src_host, enable=False)
+        assert 200 == code, (
+            f"Failed to disable maintenance mode on node {src_host} with error: {code}, {data}",
+        )
+
+    @pytest.mark.dependency(depends=["TestHostTriggerVmMigrate::test_maintenance_mode"])
```
lanfon72, line 210: As the test will not use `maintenance_mode`, so we don't need to mark this depends on it.
But we will need make this depends on `power_off_state` and `power_on_state` to let the test be skipped if scripts is not work in previous two tests.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -8,6 +8,117 @@
...
+        for condition in node_info.get('status', {}).get('conditions', []):
+            if condition.get('type', "") == "Ready" and \
+                    condition.get('status', "") == "True":
+                is_ready = True
+                break
+
+        if is_ready and not node_info.get('spec', {}).get('unschedulable', False):
+            node_names.append(node_info['metadata']['name'])
+
+    assert 2 <= len(node_names), (
```
lanfon72, line 95: In minimal, we would requires 2 nodes for it? 🤔

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -8,6 +8,117 @@
...
+        f"The cluster only have {len(node_names)} available node. It's not enough."
+    )
+    yield node_names
```
lanfon72, line 98: If no tear down, we can do `return` directly

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,150 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
     assert "active" == node["metadata"]["state"]["name"]
+
+
+@pytest.mark.hosts
+@pytest.mark.p0
+def test_verify_host_info(api_client):
+    status_code, nodes_info = api_client.hosts.get()
+
+    assert status_code == 200, f"Failed to list nodes with error: {nodes_info}"
+    assert len(nodes_info['data']) >= 1, "Incorrect hosts amount"
```
lanfon72, line 189: I think this test case should able to work with single node

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,150 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+        mems = mems or "0"
+        if int(mems, 10) == 0:
+            fail.append("Node {node_name} memory should not be zero")
+
+        if fail:
+            fails.append(fail)
+
+    assert not fails, (
+        "Failed to get node information with errors:\n",
+        "\n".join(fails)
```
lanfon72, line 213: I thought here might raise another exception: `TypeError: sequence item 0: expected str instance, list found`

If node name be included in error message, I think we can just append into `fails`.
And we should also consider the case that `node_name` is empty. (to fall back to use `data['id']` like `f"ID: {data['id']}`"?)

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -8,6 +8,117 @@
...
+        raise AssertionError(
+            f"Can't find VM {unique_name} with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    data['name'] = data['metadata']['name']
+    data['namespace'] = data['metadata']['namespace']
+    yield data
+
+    api_client.vms.delete(unique_name)
```
lanfon72, line 75: the API will not delete volumes attached to VM, we should also handle it.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,150 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+    # teardown
+    code, data = api_client.hosts.maintenance_mode(src_host, enable=False)
+    assert 200 == code, (
+        f"Failed to disable maintenance mode on node {src_host} with error: {code}, {data}",
+    )
+
+
+@pytest.mark.hosts
+@pytest.mark.p0
+@pytest.mark.dependency(depends=["test_host_poweroff_state, test_host_poweron_state"])
```
lanfon72, line 265: we need to mark those two test cases before add dependency.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,150 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"The maintain-status of node {src_host} can't be completed \
+                with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    code, data = api_client.vms.get_status(focal_vm['name'], focal_vm['namespace'])
```
lanfon72, line 247: we should also check the VM is `running` as next test does.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,159 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+
+
+@ pytest.mark.hosts
+@ pytest.mark.p0
+def test_maintenance_mode_trigger_vm_migrate(api_client, focal_vm, wait_timeout,
+                                             available_node_names):
+    assert 2 <= len(available_node_names), (
+        f"The cluster only have {len(available_node_names)} available node. \
+            It's not enough for migration test."
+    )
```
khushboo-rancher, line 219: Good check!

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,159 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+        f"Failed to migrate VM {focal_vm['namespace']}/{focal_vm['name']} \
+            from {src_host} to another node"
+    )
+    assert 'Running' == data['status']['phase'], (
+        f"Failed to migrate VM {focal_vm['namespace']}/{focal_vm['name']}, \
+            it's not running after migration"
+    )
+
+    # teardown
+    code, data = api_client.hosts.maintenance_mode(src_host, enable=False)
```
khushboo-rancher, line 261: If migration fails, the host will remain in the maintenance mode, which will lead the subsequent tests failure.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,159 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"VM {focal_vm['namespace']}/{focal_vm['name']} can't be Running \
+                with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    # teardown
+    rc, out, err = host_state.power(node['id'], node_ip, on=True)
```
khushboo-rancher, line 322: Same concern, if the test fails, the host will remain power off.

khushboo-rancher (commented):
In general LGTM, left few comments.

Can we use something like `finally` if we have mandatory tear off in the test.

harvester_e2e_tests/integration/test_hosts.py
lanfon72: I think this should be fine, we will need to check whether the fail is caused by node or VM.

harvester_e2e_tests/integration/test_hosts.py
lanfon72: same as above, we already check that cluster will have at least 2 nodes, so even 1 node is powered off, following test cases (not requires multiple nodes) will still able to be tested.

khushboo-rancher: I'm not only concerned about the next subsequent test but other tests in the entire test suite which will run after this test case. We don't want to have a powered off node when other tests are running assuming that all the nodes are up.

Do we have any fixture or are we resetting the env before starting the other module test cases?

harvester_e2e_tests/integration/test_hosts.py
khushboo-rancher: Yeah, but we don't want to have the host in maintenance mode if this test fails. It might cause problems for other tests which runs after this test.

harvester_e2e_tests/integration/test_hosts.py
lanfon72: Yeah we can make a module scoped fixture to do it, but if we do that, which means the failed environment is not kept.

there were a issue that VM will not migrate to another host automatically when the node powered off.  But the VM will try migrating again after power on the node (and the migration will success).

for other test suits, we should always check them if nodes’ status is the prerequisite of testing

lanfon72: I also have concerns about the *vm-force-reset-policy*, I think might need to discuss the test case further to make sure steps and criteria is suitable (and we should update manual tests too)

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,159 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+            fails.append((node['id'], "memory should not be zero"))
+
+    assert not fails, (
+        "Failed to get node information with errors:\n",
+        "\n".join(f"Node {n}: {r}" for n, r in fails)
+    )
+
+
+@ pytest.mark.hosts
+@ pytest.mark.p0
```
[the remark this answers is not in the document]
FrankYang0529: Fixed it. Thank you.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -67,3 +178,159 @@ def test_host_poweron_state(api_client, host_state, wait_timeout):
...
+    # teardown
+    code, data = api_client.hosts.maintenance_mode(src_host, enable=False)
+    assert 200 == code, (
+        f"Failed to disable maintenance mode on node {src_host} with error: {code}, {data}",
+    )
+
+
+@ pytest.mark.hosts
+@ pytest.mark.p0
+@ pytest.mark.dependency(depends=["host_poweroff, host_poweron"])
```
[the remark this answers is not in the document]
FrankYang0529: Fixed it. Thank you.

khushboo-rancher:
Merging this for now and we can watch the results to see our concerns surface out in the automation run.

PR #695, opened 2023-01-12
[e2e] add more VM test cases

lanfon72:
# Add Test Cases
- [x] test_create_stopped_vm
- [x] test_clone_running_vm
- [x] test_clone_stopped_vm
- [x] test_create_with_two_volumes
- [x] test_create_with_existing_volume

# Misc
- improve and fix fixture bug
- add marker `virtualmachines`
- improve `VMSpec` in harvester_api

### DIFF
--- a/harvester_e2e_tests/integration/test_vm_functions.py
+++ b/harvester_e2e_tests/integration/test_vm_functions.py
@@ -128,7 +128,6 @@ def test_minimal_vm(api_client, image, unique_vm_name, wait_timeout):
     else:
         raise AssertionError(
             f"Failed to create Minimal VM({cpu} core, {mem} RAM) with errors:\n"
-            f"Phase: {data.get('status', {}).get('phase')}\t"
             f"Status: {data.get('status')}\n"
             f"API Status({code}): {data}"
         )
@@ -260,7 +259,6 @@ class TestVMOperations:
         else:
             raise AssertionError(
                 f"Failed to Start VM({unique_vm_name}) with errors:\n"
-                f"Phase: {data.get('status', {}).get('phase')}\t"
                 f"Status: {data.get('status')}\n"
                 f"API Status({code}): {data}"
             )
@@ -308,7 +306,6 @@ class TestVMOperations:
         else:
             raise AssertionError(
                 f"Failed to Restart VM({unique_vm_name}) with errors:\n"
-                f"Phase: {data.get('status', {}).get('phase')}\t"
                 f"Status: {data.get('status')}\n"
                 f"API Status({code}): {data}"
             )
@@ -522,8 +519,9 @@ def test_create_stopped_vm(api_client, stopped_vm, wait_timeout):
     endtime = datetime.now() + timedelta(seconds=wait_timeout)
     while endtime > datetime.now():
         code, data = api_client.vms.get(unique_vm_name)
-        if ('Stopped' == data['status']['printableStatus']
-                and 'Halted' == data['spec']['runStrategy']):
+        if (code == 200
+                and 'Halted' == data['spec']['runStrategy']
+                and 'Stopped' == data.get('status', {}).get('printableStatus')):
             break
         sleep(3)
     else:
@@ -575,7 +573,6 @@ class TestVMClone:
         else:
             raise AssertionError(
                 f"Failed to Start VM({unique_vm_name}) with errors:\n"
-                f"Phase: {data.get('status', {}).get('phase')}\t"
                 f"Status: {data.get('status')}\n"
                 f"API Status({code}): {data}"
             )
@@ -636,7 +633,6 @@ class TestVMClone:
         else:
             raise AssertionError(
                 f"Failed to Start VM({cloned_name}) with errors:\n"
-                f"Phase: {data.get('status', {}).get('phase')}\t"
                 f"Status: {data.get('status')}\n"
                 f"API Status({code}): {data}"
             )
@@ -737,7 +733,6 @@ class TestVMClone:
         else:
             raise AssertionError(
                 f"Failed to Start VM({unique_vm_name}) with errors:\n"
-                f"Phase: {data.get('status', {}).get('phase')}\t"
                 f"Status: {data.get('status')}\n"
                 f"API Status({code}): {data}"
             )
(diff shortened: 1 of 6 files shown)


lanfon72:
Tested on Jenkins `harvester-runtests#195`, some (harvester) bugs found and I will file issues for them.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -61,10 +66,49 @@ def unique_vm_name(unique_name):
...
+    userdata['ssh_authorized_keys'] = [pub_key]
+    vm_spec.user_data = yaml.dump(userdata)
+
+    code, data = api_client.vms.create(unique_vm_name, vm_spec)
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.vms.get(unique_vm_name)
+        if data.get('status'):
+            break
+        sleep(1)
```
irishgordo, line 88: Do we want to implement the `sleep_timeout` fixture for this?

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -466,3 +507,642 @@ def test_delete(self, api_client, unique_vm_name, wait_timeout):
...
+                f"API Status({code}): {data}"
+            )
+        vm_ip = next(iface['ipAddress'] for iface in data['status']['interfaces']
+                     if iface['name'] == 'default')
+        code, data = api_client.hosts.get(data['status']['nodeName'])
+        host_ip = next(addr['address'] for addr in data['status']['addresses']
+                       if addr['type'] == 'InternalIP')
+
+        # Log into VM to make some data
+        with host_shell.login(host_ip, jumphost=True) as h:
```
irishgordo, line 757: when we open the context manager maybe we could name it something other than `h`?  I wonder if tox pep8 would argue about the line length if we named it something like `hstshell`?

irishgordo (changes requested):
curious if for all our `sleep()` calls if we want to leverage the timeout provided at the `config.yml` portion?

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: Consider some cases:
1. Clone VM having large Volume
2. Get latest resource of created stopped VM
3. Download Image
4. Check Backup status

For 1st case, we would expect that it will few minutes to complete, then we can snooze more seconds to reduce API calls, same as case 3 and 4.
But for 2nd case, Create a stopped VM would be completed in a short time, and we could call API more frequently to make sure the VM is created.

So as I comment in other PR, we should not use fixture to make the seconds fixed, but let it be able to adjust on demend.

lanfon72:
fix some timing bug and re-run on Jenkins `harvester-runtests#200`.

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: I think short variable name is one of conventions in Python, like the example[^1] of `open`:
```python
import os
dir_fd = os.open('somedir', os.O_RDONLY)
def opener(path, flags):
    return os.open(path, flags, dir_fd=dir_fd)

with open('spamspam.txt', 'w', opener=opener) as f:
    print('This will be written to somedir/spamspam.txt', file=f)

os.close(dir_fd)  # don't leak a file descriptor
```
Usually we will let variable name short as its living time (not scope), so sometimes you would able find some code snippets like:
```python
# https://github.com/aio-libs/aiohttp/blob/3ff5fde3051f552a0aa2ac3678ab7f5fbdc007e1/aiohttp/web.py#L328
for h in host:
    sites.append(TCPSite(...))
```
and emphasize the variable by renaming it:
https://github.com/harvester/tests/blob/4baf4bc1531c49ff561bdedf63f1b7785f8c61dc/harvester_e2e_tests/integration/test_vm_functions.py#L601-L607

It would be a bit conflict with the suggestion: meaningful variable name (one of the topic in _clean code_, targeting for Java), but yeah, it should be different, because Java don't have any JEP like PEP8.

[^1]: https://docs.python.org/3/library/functions.html#open

irishgordo: That's a great mention :smile:  - it sounds good to leave it short.
I've seen some context managers as:
```python
with open('spamspam.txt', 'w', opener=opener) as file:
    print('context manager open... with file....')
```
Where `file` was used instead of `f` but they both mean the same thing and the shorter works great :smile: .

harvester_e2e_tests/integration/test_vm_functions.py
irishgordo: Would that adjustment be that we have something like:
```python
sleep(sleep_timeout + 3)
```
If we wanted to sleep for 3 more seconds than the "fixed" base?
Or 
```python
sleep(sleep_timeout - 1)
```
If we wanted to sleep for 1 second less than the "fixed" base?

I'm thinking I could update my open PR to handle that if that's something we wanted to do.

I suppose it does get a little confusing with something like:
```python
def test_abc(request, api_client, wait_timeout, sleep_timeout):
    ...
    sleep(sleep_timout +2)
    ...
``` 
as reading through `test_abc()` we'd have to navigate over to the `config.yml` in order to find what the "base"/"fixed" sleep timeout was, then remember that and do additional math while reading through the lines in the function of `test_abc()`.
Whereas, when we do `sleep(1)` it's just very clear and apparent what the amount of sleep is before the next iteration.
So I like that as well :smile: 

Maybe if the need surfaced we could have an exponential backoff sleep, maybe some utility method that then returns a number based around the last iteration...perhaps?

irishgordo (approved):
runtests 200 passed, this in general looks good :smile: LGTM

khushboo-rancher (approved):

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: We should notice that config options could be updated by the test executor, so when we use:

```
sleep(sleep_timeout - 3)
```

The `sleep_timeout` might be negative if config option be updated to 2 or less.

PR #699, opened 2023-01-13
test: add vm migration

FrankYang0529:
* Job ID `152` for `test_migrate_vm_with_user_data`.
* Job ID `153` for `test_multiple_migrations`.
* Job ID `155` for `test_migrate_vm_with_multiple_volumes`.

### DIFF
--- a/harvester_e2e_tests/integration/test_vm.py
+++ b/harvester_e2e_tests/integration/test_vm.py
@@ -1,0 +1,319 @@
+import os
+
+from datetime import datetime, timedelta
+from time import sleep
+
+import pytest
+
+pytest_plugins = [
+    "harvester_e2e_tests.fixtures.api_client"
+]
+
+
+@pytest.fixture(scope="session")
+def focal_image_url(request):
+    base_url = request.config.getoption(
+        '--image-cache-url',
+        'https://cloud-images.ubuntu.com/focal/current/')
+    return os.path.join(base_url, "focal-server-cloudimg-amd64.img")
+
+
+@pytest.fixture(scope="class")
+def focal_image(api_client, unique_name, focal_image_url, wait_timeout):
+    code, data = api_client.images.create_by_url(unique_name, focal_image_url)
+    assert 201 == code, (
+        f"Failed to upload focal image with error: {code}, {data}"
+    )
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        code, data = api_client.images.get(unique_name)
+        if 'status' in data and 'progress' in data['status'] and \
+                data['status']['progress'] == 100:
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Image {unique_name} can't be ready with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    namespace = data['metadata']['namespace']
+    name = data['metadata']['name']
+
+    yield dict(ssh_user="ubuntu", id=f"{namespace}/{name}")
+
+    api_client.images.delete(name, namespace)
+
+
+@pytest.fixture(scope="class")
+def available_node_names(api_client):
+    status_code, nodes_info = api_client.hosts.get()
+    assert status_code == 200, f"Failed to list nodes with error: {nodes_info}"
+
+    node_names = []
+    for node_info in nodes_info['data']:
+        is_ready = False
+        for condition in node_info.get('status', {}).get('conditions', []):
+            if condition.get('type', "") == "Ready" and \
+                    condition.get('status', "") == "True":
+                is_ready = True
+                break
+
+        if is_ready and not node_info.get('spec', {}).get('unschedulable', False):
+            node_names.append(node_info['metadata']['name'])
+
+    assert 2 <= len(node_names), (
+        f"The cluster only have {len(node_names)} available node. It's not enough."
+    )
+    yield node_names
+
+
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+def test_multiple_migrations(api_client, unique_name, focal_image, wait_timeout,
+                             available_node_names):
+    vm_names = [f"migrate-1-{unique_name}", f"migrate-2-{unique_name}"]
+
+    vm_spec = api_client.vms.Spec(1, 1)
+    vm_spec.add_image('disk-0', focal_image['id'])
+    vm_data = []
+    for vm_name in vm_names:
+        code, data = api_client.vms.create(vm_name, vm_spec)
+        assert 201 == code, (
+            f"Failed to create VM {vm_name} with error: {code}, {data}"
+        )
+        vm_data.append(data)
+
+    vmi_data = []
+    vm_name, code, data = None, None, None
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        all_vm_ready = True
+        vmi_data.clear()
+        for vm_name in vm_names:
+            code, data = api_client.vms.get(vm_name)
+            if data.get('status', {}).get('ready', False):
+                code, data = api_client.vms.get_status(vm_name)
+                if code != 200 or data['status']['conditions'][-1]['status'] != 'True':
+                    all_vm_ready = False
+                    break
+                vmi_data.append(data)
+        if all_vm_ready:
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Can't find VM {vm_name} with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    vm_src_dst_hosts = {}  # {vm_name: [src_host, dst_host]}
+    for vm_name, vm_datum in zip(vm_names, vmi_data):
+        src_host = vm_datum['status']['nodeName']
+        dst_host = next(n for n in available_node_names if n != src_host)
+        vm_src_dst_hosts[vm_name] = [src_host, dst_host]
+
+        code, data = api_client.vms.migrate(vm_name, dst_host)
+        assert code == 204, (
+            f"Can't migrate VM {vm_name} from host {src_host} to {dst_host}",
+            f"Got error: {code}, {data}"
+        )
+
+    code, data = None, None
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        all_vm_migrated = True
+        for vm_name, (src, dst) in vm_src_dst_hosts.items():
+            code, data = api_client.vms.get_status(vm_name)
+            if code == 200 and \
+                    data.get('status', {}).get('migrationState', {}).get('completed', False):
+                assert dst == data['status']['nodeName'], (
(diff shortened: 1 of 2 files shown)


harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+    code, second_vm_data = api_client.vms.get_status(second_vm_name)
+    assert 200 == code, (
+        f"Failed to get VM {second_vm_name} with error: {code}, {second_vm_data}"
+    )
+
+    first_vm_src_host = first_vm_data['status']['nodeName']
+    first_vm_dst_host = ""
+    for node_name in available_node_names:
+        if node_name != first_vm_src_host:
+            first_vm_dst_host = node_name
```
lanfon72, line 120: this could be reduced as:
```python
first_vm_dst_host = next(n for n in available_node_names if n != first_vm_src_host)
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+            f"Got error: {code}, {data}"
+        )
+
+
+@pytest.mark.multivmtest
+@pytest.mark.virtual_machines_p1
+def test_multiple_migrations(api_client, unique_name, focal_image, wait_timeout,
+                             available_node_names):
+    first_vm_name = f"{unique_name}-1"
+    second_vm_name = f"{unique_name}-2"
```
lanfon72, line 89: I will suggest use
```python
vms = [f"migrate-1-{unique_name}", f"migrate-2-{unique_name}"]
```

The unique_name would be very long, so add prefix would better than suffix (when the name is meaningful)

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+    )
+    second_vm_spec = api_client.vms.Spec(1, 1)
+    second_vm_spec.add_image('disk-0', focal_image['id'])
+    code, data = api_client.vms.create(second_vm_name, second_vm_spec)
+    assert 201 == code, (
+        f"Failed to create VM {second_vm_name} with error: {code}, {data}"
+    )
+
+    wait_vm_agent_connected(api_client, first_vm_name, wait_timeout)
+    wait_vm_agent_connected(api_client, second_vm_name, wait_timeout)
```
lanfon72, line 105: we should check VMs in the same loop to let the error message like:
```python
assert len(vms) == len(connected_vms), (
    f"{len(vms) - len(connected_vms)} VM(s) is not able to boot into OS\n"
    "\n".join(f"{vm} is not connected" for vm in set(vms).difference(connected_vms))
)
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+        code, data = api_client.vms.get_status(second_vm_name)
+        if data.get('status', {}).get('migrationState', {}).get('completed', False):
+            assert second_vm_dst_host == data['status']['nodeName'], (
+                f"Failed to migrate VM {second_vm_name} \
+                    from {second_vm_src_host} to {second_vm_dst_host}"
+            )
+            break
+        sleep(5)
+    else:
+        raise AssertionError(
```
lanfon72, line 162: we should check them in the same while loop, so the error would like:
```python
assert len(vms) == len(connected_vms), (
    f"{len(vms) - len(migrated_vms)} VM(s) is not able to boot into OS\n"
    "\n".join(f"{vm} is not connected" for vm in set(vms).difference(migrated_vms))
)
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+            f"Got error: {code}, {data}"
+        )
+
+    # teardown
+    api_client.vms.delete(first_vm_name)
+    volume_name = ""
+    for volume in first_vm_data['spec']['volumes']:
+        if volume['name'] == 'disk-0':
+            volume_name = volume['persistentVolumeClaim']['claimName']
+    api_client.volumes.delete(volume_name)
```
lanfon72, line 173: when VM is still terminating, Volumes will not able to delete, we should check whether the VM is removed first.
and we can leverage Spec to remove all volumes like:
```python
for vol in api.vms.Spec.from_dict(data).volumes:
    api.volumes.delete(vol['volume']['persistentVolumeClaim']['claimName'])
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+    vm_spec = api_client.vms.Spec(1, 1)
+    vm_spec.add_image('disk-0', focal_image['id'])
+    user_data = vm_spec.user_data
+    user_data += (
+        "password: test\n"
+        "chpasswd:\n"
+        "  expire: false\n"
+        "ssh_pwauth: true\n"
+    )
+    vm_spec.user_data = user_data
```
lanfon72, line 194: it is the same as:
```python
vm_spec.user_data += (...)
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,295 @@
...
+    code, data = api_client.vms.create(unique_name, vm_spec)
+    assert code == 201, (
+        f"Failed to create VM {unique_name} with error: {code}, {data}"
+    )
+
+    wait_vm_agent_connected(api_client, unique_name, wait_timeout)
```
lanfon72, line 200: If we really need helper function, we should use assertion in test cases, not in helper functions.

lanfon72 (changes requested):
I'm not add duplicated comments in the similar code lines, please check all similar codes.

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+        code, data = api_client.vms.migrate(vm_name, dst_host)
+        assert code == 204, (
+            f"Can't migrate VM {vm_name} from host {src_host} to {dst_host}",
+            f"Got error: {code}, {data}"
+        )
+
+    for vm_name, src_dst_hosts in vm_src_dst_hosts.items():
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        code, data = None, None
+        while endtime > datetime.now():
```
lanfon72, line 134: same as above, we would need use `while-for`, if timed out, we would know all VMs' status. (when raising exception)

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+                f"Got error: {code}, {data}"
+            )
+
+    # teardown
+    for vm_name in vm_names:
+        api_client.vms.delete(vm_name)
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        code, data = None, None
+        while endtime > datetime.now():
+            code, data = api_client.vms.get(vm_name)
```
lanfon72, line 155: I had encountered that VM definition has been removed but VMI still exists, so I would suggest to use `get_status` to make sure VMI been removed. (to avoid delete volumes being deny.)

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+            if data['status']['conditions'][-1]['status'] == 'True':
+                break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Can't find VM {unique_name} with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    code, vmi_data = api_client.vms.get_status(unique_name)
```
lanfon72, line 203: we can get vmi info inside the while loop (before `break`)

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+    assert 200 == code, (
+        f"Failed to get VM {unique_name} with error: {code}, {vmi_data}"
+    )
+    src_host = vmi_data['status']['nodeName']
+    dst_host = ""
+    for node_name in available_node_names:
```
lanfon72, line 209:
```python3
dst_host = next(n for n in available_node_names if n != src_host)
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+            if data['status']['conditions'][-1]['status'] == 'True':
+                break
+        sleep(5)
+    else:
+        raise AssertionError(
+            f"Can't find VM {unique_name} with {wait_timeout} timed out\n"
+            f"Got error: {code}, {data}"
+        )
+
+    code, vmi_data = api_client.vms.get_status(unique_name)
```
lanfon72, line 280: we can get vmi info inside the while loop (before `break`)

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,327 @@
...
+    assert 200 == code, (
+        f"Failed to get VM {unique_name} with error: {code}, {vmi_data}"
+    )
+    src_host = vmi_data['status']['nodeName']
+    dst_host = ""
+    for node_name in available_node_names:
```
lanfon72, line 286:
```python3
dst_host = next(n for n in available_node_names if n != src_host)
```

lanfon72 (changes requested):
Please rebase main before update changes.

to prevent commit time be refreshed, you can use `git rebase --ignote-date main`

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,313 @@
...
+        assert code == 204, (
+            f"Can't migrate VM {vm_name} from host {src_host} to {dst_host}",
+            f"Got error: {code}, {data}"
+        )
+
+    code, data = None, None
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        all_vm_migrated = True
+        for vm_name, src_dst_hosts in vm_src_dst_hosts.items():
```
lanfon72, line 127: It cloud also be written as:
```python
for vm_name, (src, dst) in vm_src_dst_hosts.items():
    ....
```

harvester_e2e_tests/integration/test_vm.py
```diff
@@ -0,0 +1,313 @@
...
+            f"The migration of VM {vm_name} is not completed with {wait_timeout} timed out"
+            f"Got error: {code}, {data}"
+        )
+
+    # teardown
+    for vm_name in vm_names:
+        api_client.vms.delete(vm_name)
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        code, data = None, None
+        while endtime > datetime.now():
```
lanfon72, line 152: I would suggest to use `while-for` instead, in `for-while` case, the 2nd VM will never handled if 1st VM timed out.

lanfon72 (approved):
In general LGTM

lanfon72 (approved):

khushboo-rancher (approved):

khushboo-rancher:
Thanks @FrankYang0529 & @lanfon72

PR #721, opened 2023-02-16
RestoreSpec fix.

masteryyh:
A little fix up when calling `to_dict` in `RestoreSpec`.


lanfon72 (changes requested):
Hi @masteryyh, please help to fix this bug in `BackupManager.create`, we would need to keep models fetch minimal data as possible.  thank you 

https://github.com/harvester/tests/blob/4eaae360b37ff233e852efa31eb5ab9ce7ec28ae/apiclient/harvester_api/managers.py#L357-L362

```python
# fix in L359
old_vm = data['spec']['source']['name']
```

lanfon72 (approved):
Thanks!

PR #738, opened 2023-03-08
[e2e] fix e2e testing errors for v1.1-head

lanfon72:
## Changes (Based on `harvester-install-and-test/build#673`)
- [apiclient] add new parameter `force=False` for `hosts.maintenance_mode`
- Update `apis/test_hosts.py` to wait node entering maintenance mode for its enhancement https://github.com/harvester/harvester/issues/3363
- Update `apis/test_images.py` to include condition of retry which introduced in https://github.com/harvester/harvester/pull/3148
- Update `apis/test_networks.py` and fixture `skip_version_after` to skip test case correctly
- Add new test case to fulfill skipped test cases in `apis/test_networks.py`
- Update `integration/test_volumes.py` to fix fixture reference

## To fix
- [x] apis/test_hosts.py::test_maintenance_mode
- [x] apis/test_images.py::test_create_with_invalid_url
- [x] apis/test_networks.py::TestNetworksNegative::test_create_with_invalid_id[0]
- [x] apis/test_networks.py::TestNetworksNegative::test_create_with_invalid_id[4095] 
- [x] apis/test_networks.py::TestNetworks::test_create 
- [x] integration/test_volumes.py::test_create_volume_backing_image

## Need more monitoring
- [ ] integration/test_rancher_integration.py::TestRKE::test_create_rke1
- [ ] integration/test_backup_target.py::test_invalid_backup_target_nfs
- [x] integration/test_hosts.py::test_host_poweron_state

### Potential bug
- [ ] integration/test_hosts.py::test_maintenance_mode_trigger_vm_migrate
### Legacy code
- [ ] scenarios/test_create_vm.py::test_create_vm_on_available_cpu_node
- [ ] scenarios/test_create_vm.py::test_update_vm_on_available_cpu_node
- [ ] scenarios/test_vm_actions.py::test_restore_backup_vm_on

### DIFF
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -105,10 +105,10 @@ class HostManager(BaseManager):
     def get_metrics(self, name="", *, raw=False):
         return self._get(self.METRIC_fmt.format(uid=name), raw=raw)
 
-    def maintenance_mode(self, name, enable=True):
+    def maintenance_mode(self, name, enable=True, force=False):
         action = "enable" if enable else "disable"
         params = dict(action=f"{action}MaintenanceMode")
-        self._create(self.PATH_fmt.format(uid=name), params=params)
+        self._create(self.PATH_fmt.format(uid=name), params=params, data=dict(force=force))
         return self.get(name)
 
 
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -108,8 +108,8 @@ class HostManager(BaseManager):
     def maintenance_mode(self, name, enable=True, force=False):
         action = "enable" if enable else "disable"
         params = dict(action=f"{action}MaintenanceMode")
-        self._create(self.PATH_fmt.format(uid=name), params=params, data=dict(force=force))
-        return self.get(name)
+        payload = dict(force=str(force).lower())
+        return self._create(self.PATH_fmt.format(uid=name), params=params, json=payload)
 
 
 class ImageManager(BaseManager):
--- a/harvester_e2e_tests/apis/test_networks.py
+++ b/harvester_e2e_tests/apis/test_networks.py
@@ -58,14 +58,28 @@ class TestNetworksNegative:
         assert 422 == code, (code, data)
         assert "Invalid" == data.get("reason"), (code, data)
 
+    @pytest.mark.parametrize("vlan_id", [4095])
+    @pytest.mark.skip_version_before("v1.1.0")  # ref to harvester/issues/3151
+    def test_create_with_invalid_id_110(self, api_client, unique_name, vlan_id):
+        code, data = api_client.networks.create(unique_name, vlan_id)
+
+        assert 422 == code, (vlan_id, code, data)
+        assert "Invalid" == data.get("reason"), (vlan_id, code, data)
+
 
 @pytest.mark.p0
 @pytest.mark.networks
 class TestNetworks:
 
     @pytest.mark.dependency(name="create_network")
+    @pytest.mark.skip_version_after("v1.0.3")
     def test_create(self, api_client, unique_name):
         code, data = api_client.networks.create(unique_name, VLAN_ID)
+        assert 201 == code, (code, data)
+
+    @pytest.mark.skip_version_before("v1.1.0")
+    def test_create_110(self, api_client, unique_name):
+        code, data = api_client.networks.create(unique_name, VLAN_ID, cluster_network='mgmt')
         assert 201 == code, (code, data)
 
     @pytest.mark.dependency(depends=["create_network"])
--- a/harvester_e2e_tests/apis/test_networks.py
+++ b/harvester_e2e_tests/apis/test_networks.py
@@ -46,7 +46,7 @@ class TestNetworksNegative:
 
     @pytest.mark.parametrize("vlan_id", [0, 4095])
     @pytest.mark.skip_version_after("v1.0.3")  # ref to harvester/issues/3151
-    def test_create_with_invalid_id(self, api_client, unique_name, vlan_id):
+    def test_create_with_invalid_id_103(self, api_client, unique_name, vlan_id):
         code, data = api_client.networks.create(unique_name, vlan_id)
 
         assert 422 == code, (vlan_id, code, data)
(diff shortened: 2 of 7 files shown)


lanfon72:
tested on `harvester-runtests/build#210`

apiclient/harvester_api/managers.py
```diff
@@ -105,11 +105,11 @@ def delete(self, name, *, raw=False):
     def get_metrics(self, name="", *, raw=False):
         return self._get(self.METRIC_fmt.format(uid=name), raw=raw)
 
-    def maintenance_mode(self, name, enable=True):
+    def maintenance_mode(self, name, enable=True, force=False):
         action = "enable" if enable else "disable"
         params = dict(action=f"{action}MaintenanceMode")
-        self._create(self.PATH_fmt.format(uid=name), params=params)
-        return self.get(name)
+        payload = dict(force=str(force).lower())
```
irishgordo, line 111: Do we need to do a `.lower()` there?
It doesn't seem like it matters to the API backend.
With for instance:
```
curl --location --request POST 'https://192.168.1.186/v1/harvester/nodes/v11-head-top?action=enableMaintenanceMode' \
--header 'Authorization: Bearer token-wz48k:4c8rwf6lfqcvb2q6c688vj5s6ltvqx2kclkfs9cvmcg9gfrg2h6vz2' \
--header 'Content-Type: application/json' \
--data-raw '{
    "force": "TrUe"
}'
```
That returns a 204 and does put the node into maintenance mode, forcefully as far as I can tell.
I also wasn't able to find this documented in the OpenAPI swagger.json spec, I wonder if this could be added to show the users that this can be an API action:
https://github.com/harvester/harvester/blob/master/api/openapi-spec/swagger.json

irishgordo (changes requested):

apiclient/harvester_api/managers.py
lanfon72: The parameter be introduced in https://github.com/harvester/harvester/issues/2723, and you can see we will use `true` to identify whether it is **forced** in the switch case.[^1]

The PR didn't update swagger API spec, so it isn't listed in the API.

[^1]: https://github.com/harvester/harvester/blob/9f440f9fc299bb1fe5404aac88ee5ac57ca1522d/pkg/api/node/formatter.go#L98

irishgordo: @lanfon72  - Thanks for linking that :+1: !

I guess, I'm surprised by the request of:
```
curl --location --request POST 'https://192.168.1.186/v1/harvester/nodes/v11-head-top?action=enableMaintenanceMode' \
--header 'Authorization: Bearer token-wz48k:4c8rwf6lfqcvb2q6c688vj5s6ltvqx2kclkfs9cvmcg9gfrg2h6vz2' \
--header 'Content-Type: application/json' \
--data-raw '{
    "force": "TrUe"
}'
```
Getting a `204` back.
This sounds like an API question I suppose, since the server I'm sure understands the request, but the JSON payload is incorrect. 
On the backend, it *must* have a case-sensitive nature, as in:
https://github.com/harvester/harvester/blob/master/pkg/api/node/formatter.go#L126-L143
( enabling maintenance mode )
in the source it relies on an exact match of "true" ( with the casing being respected )
Same with:
https://github.com/harvester/harvester/blob/master/pkg/api/node/formatter.go#L92-L101

I'm just surprised because I think the end user would also like to know that:
```
{
   "force": "TrUe"
}
```
the value of `TrUe` is invalid and shouldn't be sending back a `204` perhaps send back instead a `422`.

Also curious as to why the design was to send a `string` of "true" vs something like:
```json
{
    "force": true
}
```
boolean values are valid JSON... :thinking: 

I'm just confused as to how the end user would know to pass a "case sensitive" value when the feedback from the api is a `204`.  Like how they would know "they" made a mistake in the payload value with the `204`.  Like the node will be put into maintenance mode regardless but if they wanted to force the node into maintenance mode and shipped the wrong JSON payload, they wouldn't know from the response that it was inadequate.

irishgordo (approved):

apiclient/harvester_api/managers.py
lanfon72: > the value of `TrUe` is invalid and shouldn't be sending back a `204` perhaps send back instead a `422`.

It returns `204` means the `maintenance mode` been applied, the `force` option is the additional way to **forced** drain nodes, if it is not matched, maintenance mode will still applied without **forced** drain, so it should not return `422` but `204`.

> I'm just confused as to how the end user would know to pass a "case sensitive" value when the feedback from the api is a `204`. 

We don't have any API library yet, and API document not include this, so end user will only able to use the function by dashboard or the `harvester_api` (apiclient) in this repo.

To use boolean or not, It is design decision, I thought you can discuss with @ibrokethecloud about the details.

irishgordo: that's a great point :smile: 
yeah, since it would still put it into maintenance mode regardless of the `force` option being passed in `422` doesn't make sense.
I looked at those code snippets again this morning and it makes more sense :+1: :smile: 

I guess I'm still curious surrounding the case sensitivity but if a user is typically going to only ever trigger this via the dashboard then I suppose it doesn't matter that much :smile:

harvester_e2e_tests/apis/test_networks.py
```diff
@@ -58,16 +58,30 @@ def test_create_without_name(self, api_client):
         assert 422 == code, (code, data)
         assert "Invalid" == data.get("reason"), (code, data)
 
+    @pytest.mark.parametrize("vlan_id", [4095])
+    @pytest.mark.skip_version_before("v1.1.0")  # ref to harvester/issues/3151
+    def test_create_with_invalid_id_110(self, api_client, unique_name, vlan_id):
```
khushboo-rancher, line 63: Let's keep the name as `test_create_with_invalid_id` which will run after v1.1.0 and `test_create_with_invalid_id_103` for the test which run with version 1.0.3.

harvester_e2e_tests/apis/test_networks.py
```diff
@@ -58,16 +58,30 @@ def test_create_without_name(self, api_client):
...
 class TestNetworks:
 
     @pytest.mark.dependency(name="create_network")
+    @pytest.mark.skip_version_after("v1.0.3")
     def test_create(self, api_client, unique_name):
         code, data = api_client.networks.create(unique_name, VLAN_ID)
         assert 201 == code, (code, data)
 
+    @pytest.mark.skip_version_before("v1.1.0")
+    def test_create_110(self, api_client, unique_name):
```
khushboo-rancher, line 81: Same as above comment, let's keep the name as `test_create` of the test which will run after v1.1.0 and `test_create_103` for v1.0.3

khushboo-rancher (commented):
LGTM, just one minor comment

lanfon72:
tested on `harvester-install-and-test/build#683` with `v1.1.2-rc3`

harvester_e2e_tests/apis/test_networks.py
lanfon72: updated

khushboo-rancher (approved):

PR #763, opened 2023-04-10
[e2e] fix bugs in api and integration test cases

lanfon72:
## Changes
- **API**
    - Fix resource condition bugs
    - Fix dependencies in `test_network`
- **Integration**
    - Fix checker in `test_host_poweron_state`
    - Add checker for fixture `stoppped_vm` (to fix conditional bug in several test cases)
    - Fix checker for `unpase` and `restart`

### DIFF
--- a/harvester_e2e_tests/apis/test_hosts.py
+++ b/harvester_e2e_tests/apis/test_hosts.py
@@ -45,6 +45,94 @@ def test_get_host(api_client):
     assert node_id == node['id'], f"Responsed unexpected Node {node['id']}"
 
 
+@pytest.mark.dependency(depends=["get_host"])
+@pytest.mark.hosts
+@pytest.mark.p0
+def test_update_node(api_client):
+    """
+    Test the hosts are the nodes which make the cluster
+    Covers:
+        hosts-04-Edit Config Add description and other details
+    """
+
+    fields = ('field.cattle.io/description', 'harvesterhci.io/host-custom-name')
+
+    _, nodes_info = api_client.hosts.get()
+
+    # There is no different to update any nodes, so we use the last.
+    node = nodes_info['data'][-1]
+    original_annotations = {k: node['metadata']['annotations'].get(k, "")
+                            for k in fields}
+
+    test_annotations = {k: f"Test_update_{k}_{datetime.now()}" for k in fields}
+    test_data = dict(metadata=dict(annotations=test_annotations))
+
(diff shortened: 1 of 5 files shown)


harvester_e2e_tests/apis/test_hosts.py
```diff
@@ -131,8 +80,11 @@ def test_update_node(api_client):
     )
 
     # For teardown
-    _ = api_client.hosts.update(node['id'],
-                                dict(metadata=dict(annotations=original_annotations)))
+    sleep(1)  # to prevent update too fast cause code 409 conflict: 'object has been modified'
```
khushboo-rancher, line 83: Do we have better way to avoid 409 conflict, like `get` after `update` to make sure the object is update?

harvester_e2e_tests/apis/test_hosts.py
```diff
@@ -174,8 +126,62 @@ def test_update_using_yaml(api_client):
...
```
khushboo-rancher, line 129: Same as above

harvester_e2e_tests/apis/test_hosts.py
```diff
@@ -174,8 +126,62 @@ def test_update_using_yaml(api_client):
...
+    code, data = api_client.hosts.update(node['id'],
+                                         dict(metadata=dict(annotations=original_annotations)))
+
+    assert 200 == code, (code, data)
+
+
+@pytest.mark.dependency(depends=["get_host"], name="enable_maintenance_mode")
+@pytest.mark.hosts
+@pytest.mark.p2
+def test_maintenance_mode(api_client, wait_timeout):
```
khushboo-rancher, line 139: Just to confirm, no change in this test just the placement?

khushboo-rancher (commented):
LGTM in general, few clarifications needed.

harvester_e2e_tests/apis/test_hosts.py
lanfon72: I don't have any better way to prevent resource conflict for now, this more like a temporarily fix because it will not always happen in the test case.

The problem of conflict is, we update host's information and do few check in million seconds, then revert (update) it to the original information.  when calling `hosts.update`, API will do `hosts.get` internally to make sure the resource is latest, but as the result, it will not always work (then hit resource conflict).

khushboo-rancher (approved):

PR #793, opened 2023-05-08
fix RKE1 test case

lanfon72:
## Changes
- **ADD** inject `qemu-agent` for **NodeTemplate**
- **UPDATE** use **latest** `engineInstallURL` instead of constant
- **UPDATE** fixture `focal_image_url` to use cache if possible
- **ADD** fixture `rke1_k8s_version` to convert rke2 version to fit rke1 format, when the rke1 version is not support, use latest instead.

### DIFF
--- a/harvester_e2e_tests/integration/test_rancher_integration.py
+++ b/harvester_e2e_tests/integration/test_rancher_integration.py
@@ -15,6 +15,9 @@
 # To contact SUSE about this file by physical or electronic mail,
 # you may find current contact information at www.suse.com
 
+import re
+import os
+import warnings
 from time import sleep
 from datetime import datetime, timedelta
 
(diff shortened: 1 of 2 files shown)


harvester_e2e_tests/integration/test_rancher_integration.py
```diff
@@ -47,6 +50,32 @@ def rke2_cluster_name(unique_name):
     return f"{unique_name}-rke2"
 
 
+@pytest.fixture(scope='module')
+def rke1_k8s_version(k8s_version, rancher_api_client):
+    # `v1.24.11+rke2r1` -> `v1.24.11-rancher2-1`
```
irishgordo, line 55: I appreciate the comment in the code :smile: :+1:  as it helps make sense of the Lambda and RegEx that's taking place

FrankYang0529 (approved):
LGTM. Thanks.

irishgordo (approved):

PR #802, opened 2023-05-10
+ [apiclient] Implement VirtualMachineSnapshotManager

lanfon72:
## Changes
- **ADD** `VirtualMachineSnapshotManager` to manipulate snapshot API calls (get, create, delete, restore)
- **ADD** `vm.snapshot` to create snapshot (alias to `vm_snapshots.create)
- **UPDATE** `backups.create` change to create backup (alias to `vm.backup`) rather than **restore** a backup.

### DIFF
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -361,16 +361,33 @@ class TemplateManager(BaseManager):
 
 
 class BackupManager(BaseManager):
-    BACKUP_fmt = "v1/harvester/harvesterhci.io.virtualmachinebackups/{ns}/{uid}"
+    BACKUP_fmt = "v1/harvester/harvesterhci.io.virtualmachinebackups/{ns}{uid}"
     RESTORE_fmt = "v1/harvester/harvesterhci.io.virtualmachinerestores/{ns}"
 
     RestoreSpec = RestoreSpec
 
     def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
-        path = self.BACKUP_fmt.format(uid=name, ns=namespace)
-        return self._get(path, raw=raw, **kwargs)
+        path = self.BACKUP_fmt.format(uid=f"/{name}", ns=namespace)
+        resp = self._get(path, raw=raw, **kwargs)
+        try:
+            code, data = resp[0], resp[1]
+            if name and "backup" != data['spec']['type']:
+                return 404, dict(type='error', status=404, message=f'Backup {name!r} not found')
 
-    def create(self, name, restore_spec, namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
+            data['data'] = [d for d in data['data'] if "backup" == d.get('spec', {}).get('type')]
+            return code, data
+        except TypeError:
+            # raw=True
+            return resp
+        except KeyError:
+            # !data.spec || !data.data
+            return code, data
+
+    def create(self, *args, **kwargs):
+        # Delegate to vm.backups
+        return self.api.vms.backup(*args, **kwargs)
+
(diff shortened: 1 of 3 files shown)


lanfon72:
Code snpipets for the new API:
```python
from harvester_api import API
import urllib3
urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
api = API.login('https://VIP', 'admin', 'YOUR PASSWORD', ssl_verify=False)

vm_name = "exist vm in default namespace"
snapshot_name = "snapshot-name"

# CREATE a snapshot for `vm_name`
code, data = api.vm_snapshots.create(vm_name, )
# equals to
code, data = api.vms.snapshot(vm_name, f"{snapshot_name}-2")

# GET all snapshots in `default` namespace
code, data = api.vm_snapshots.get()
# GET specific snapshot
code, data = api.vm_snapshots.get(snapshot_name)

# RESTORE a snapshot to new VM
spec = api.vm_snapshots.RestoreSpec.for_new("new-vm-name")
code, data = api.vm_snapshots.restore(snapshot_name, spec)
# RESTORE a snapshot to replace existing VM
spec = api.vm_snapshots.RestoreSpec.for_existing()
# VM need to be stopped before restore
api.vms.stop(vm_name)
code, data = api.vm_snapshots.restore(snapshot_name, spec)

# DELETE a snapshot
code, data = api.vm_snapshots.delete(snapshot_name)
```

cc @connorkuehl

apiclient/harvester_api/managers.py
```diff
@@ -361,27 +362,95 @@ def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
...
+        # Delegate to vm.backups
+        return self.api.vms.backup(*args, **kwargs)
+
+    def restore(self, name, restore_spec, namespace=DEFAULT_NAMESPACE, *, raw=False, **kwargs):
+        code, data = self.get(name, namespace)
+        try:
+            old_vm = data['spec']['source']['name']
+            spec = restore_spec.to_dict(name, namespace, old_vm)
+            path = self.RESTORE_fmt.format(ns=restore_spec.namespace or namespace)
+            return self._create(path, json=spec, raw=raw, **kwargs)
```
khushboo-rancher, line 397: Should this be `_create` as it is creating vm's backup?

khushboo-rancher (changes requested):

apiclient/harvester_api/managers.py
```diff
@@ -361,27 +362,95 @@ def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
...
-        path = self.BACKUP_fmt.format(uid=name, ns=namespace)
+        path = self.BACKUP_fmt.format(uid=f"/{name}", ns=namespace)
         return self._delete(path, raw=raw, **kwargs)
 
-    restore = create  # alias
+
+class VirtualMachineSnapshotManager(BackupManager):
+    RestoreSpec = SnapshotRestoreSpec
+
+    def create_data(self, vm_uid, vm_name, snapshot_name, namespace):
```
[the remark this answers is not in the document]
lanfon72: we use `create_data` in other managers, and consider the case:
```python
spec = api.vm_snapshots.RestoreSpec(...)  # will returns an object to manipulate

data = api.vm_snapshots.create_data(...)   # will returns a dict as a data set
spec = api.vm_snapshots.create_spec(...)  # will returns a dict instead of object
```

we already let `spec` point to an object to manipulate and propagate to API calls, so I thought in here, we use `create_data` will self-explain its type.

apiclient/harvester_api/managers.py
lanfon72: do you mean `vm._create`?
In this method, we will create a VM from backup's API endpoint, I would be a bit confusing when you use the calls like `api.backups._create`, but as `_create` is a protected method (as python's convention), user will not use it or they should know how it works.

khushboo-rancher (approved):

PR #808, opened 2023-05-17
[e2e] add  keyword for test selection

lanfon72:
## Changes
- **ADD** `__init__.py` file to make `package` scope for **integration**
- **add** `and depends` keyword for test selection when using `-k` to select test cases

for example, when we want to test `delete_rke2`, it would also need select the test `create_rke2`, otherwise the test will pass with test skipped:
image
image

and now we can use `and depends` to select the test and its depended test cases:
image

To be aware that this PR will not recursively add all depended test cases[^1] (it's the known issue):
image

[^1]: for `test_start` in `test_vm_functions.py`, it will depends on `test_stop` which depends on `minimal_vm`, so in this case, we should select `test_minimal_vm`, `test_stop` then `test_start`

### DIFF
--- a/harvester_e2e_tests/conftest.py
+++ b/harvester_e2e_tests/conftest.py
@@ -418,39 +418,46 @@ def pytest_collection_modifyitems(session, config, items):
         "class": pytest.Class
     }
     # ref: https://github.com/RKrahl/pytest-dependency/blob/0.5.1/pytest_dependency.py#L77
-    # named_items : dict['dep-scope', dict['dep-name', list['test-item']]]
-    deselected, named_items = list(), dict()
-    for item in all_items:
+    # deselected : list['test-item']
+    # named_items : dict[('dep-scope', 'dep-name'), list[(idx, 'test-item')]]
+    # picked : list[(idx, 'test-item')]
+    deselected, named_items, picked = list(), dict(), list()
+    for idx, item in enumerate(all_items):
         if item in items:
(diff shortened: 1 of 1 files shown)


khushboo-rancher:
cc: @harvester/qa

lanfon72:
Update to enable diamond dependencies, for existing examples:
```
// In test_vm_functions.py
test_start -> test_stop -> minimal_vm
```
which means when we select `test_start`, both `test_stop` and `minimal_vm` should be selected (as they be depended)
This PR will enable the selection by keyword `and depends` to propagate to pytest's keyword argument (`-k`)

The more complex example, consider we have a test suite like:
```python
import pytest


@pytest.mark.dependency()
def test_1():
    pass


@pytest.mark.dependency()
def test_2():
    pass


@pytest.mark.dependency()
def test_4():
    pass


class Test3:
    @pytest.mark.dependency(name='test_a', depends=['test_2', 'test_4'])
    def test_a(self):
        pass

    @pytest.mark.dependency(name='test_b', depends=['test_2'])
    def test_b(self):
        pass

    @pytest.mark.dependency(depends=['test_a', 'test_b'])
    def test_c(self):
        pass
```
When we want to select `test_c` to do testing, we can use the keyword for pytest like `pytest -k "test_c and depends"`, then all depended test cases would be selected together.
image

khushboo-rancher (approved):

PR #811, opened 2023-05-18
[e2e] fix bug when testing between NFS/S3

lanfon72:
## Changes
- **ADD** `backup_config` fixture to trigger VM recreation
- **ADD** clear up all backups not belongs to current `backup-target`

### DIFF
--- a/harvester_e2e_tests/integration/test_backup_restore.py
+++ b/harvester_e2e_tests/integration/test_backup_restore.py
@@ -106,12 +106,20 @@ def config_backup_target(api_client, backup_config):
 
     yield spec
 
+    # restore to original backup-target and remove backups not belong to it
     code, data = api_client.settings.update('backup-target', origin_spec)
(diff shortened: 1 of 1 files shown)


lanfon72:
tested on `harvester-runtests#272`
image

lanfon72:
The logic of fixture setup:
- **backup_config**, be used to return the configuration by the parameter propagated from parameterize marker
- **config_backup_target**, be used to setup `backup-target`, so this will request **backup_config**

For test class (or functions), will need add both `backup_config` and `config_backup_target` to trigger `backup-target` be setup and tear down.

For the fixture `base_vm_with_data`, actually we don't need `backup_config` because it should not depends on backup target configuration, but we add it to trigger the fixture re-create to keep the VM as clean as possible, otherwise it will be used for `TestBackupRestore[NFS]` and `TestBackupRestore[S3]`, with data wrote in from previous tested test cases.

PR #817, opened 2023-05-19
Add integration tests for VM snapshot support

connorkuehl:
Closes: https://github.com/harvester/tests/issues/363

There's an additional commit that adds an (incomplete) VolumeSnapshotManager. This is because one of the test cases asserts that the volume snapshot is automatically cleaned up after the volume it relates to has been deleted, so the test case asserts that it exists, then after the volume is deleted, it is used to assert that it no longer exists. I can split this commit out to a separate PR to implement it more completely, if that is preferred.

### DIFF
--- a/apiclient/harvester_api/managers.py
+++ b/apiclient/harvester_api/managers.py
@@ -221,6 +221,18 @@ class VolumeManager(BaseManager):
         return self._create(path, params=params, json=export_spec, raw=raw)
 
 
+class VolumeSnapshotManager(BaseManager):
+    PATH_fmt = "v1/harvester/snapshot.storage.k8s.io.volumesnapshots/{ns}/{uid}"
+
+    def delete(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
+        path = self.PATH_fmt.format(uid=name, ns=namespace)
+        return self._delete(path, raw=raw)
+
+    def get(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False):
+        path = self.PATH_fmt.format(uid=name, ns=namespace)
+        return self._get(path, raw=raw)
+
+
 class TemplateManager(BaseManager):
     # get, create, delete
     PATH_fmt = "apis/{{API_VERSION}}/namespaces/{ns}/virtualmachinetemplates/{uid}"
--- a/harvester_e2e_tests/integration/test_vm_snapshot.py
+++ b/harvester_e2e_tests/integration/test_vm_snapshot.py
@@ -1,0 +1,577 @@
+import yaml
+from datetime import datetime, timedelta
+from time import sleep
+
+from paramiko.ssh_exception import ChannelException
+import pytest
+
+
+pytest_plugins = [
+    "harvester_e2e_tests.fixtures.api_client",
+    "harvester_e2e_tests.fixtures.images",
+    "harvester_e2e_tests.fixtures.virtualmachines"
+]
+
+
+def unique_name(name):
+    return f"{datetime.now().strftime('%m%S%f')}-{name}"
+
+
+@pytest.fixture(scope="class")
+def sourcevm_name():
+    return unique_name("source-vm")
+
+
+@pytest.fixture(scope="class")
+def restored_from_snapshot_name():
+    return unique_name("vm-from-snapshot")
+
+
+@pytest.fixture(scope="class")
+def restored_from_snapshot_name_2():
+    return unique_name("vm-from-snapshot-2")
+
+
+@pytest.fixture(scope="class")
+def vm_snapshot_name():
+    return "vm-snapshot"
+
+
+@pytest.fixture(scope="class")
+def vm_snapshot_2_name():
+    return "vm-snapshot-2"
+
+
+@pytest.fixture(scope="module")
+def image(api_client, image_opensuse, wait_timeout):
+    unique_image_id = unique_name("image")
+    display_name = f"{unique_image_id}-{image_opensuse.name}"
+    code, data = api_client.images.create_by_url(unique_image_id,
+                                                 image_opensuse.url,
+                                                 display_name=display_name)
+
+    assert 201 == code, (code, data)
+
+    deadline = datetime.now() + timedelta(seconds=wait_timeout)
+    while deadline > datetime.now():
+        code, data = api_client.images.get(unique_image_id)
+        if 100 == data.get('status', {}).get('progress', 0):
+            break
+        sleep(3)
+    else:
+        raise AssertionError(
+            "Failed to create Image with error:\n"
+            f"Status({code}): {data}"
+        )
+
+    yield dict(id=f"{data['metadata']['namespace']}/{unique_image_id}",
+               user=image_opensuse.ssh_user)
+
+    code, data = api_client.images.delete(unique_image_id)
+
+
+def create_vm(name, api_client, ssh_keypair, image, timeout_secs):
+    cpu = 1
+    mem = 2
+    pubkey, _ = ssh_keypair
+
+    vm_spec = api_client.vms.Spec(cpu, mem)
+    vm_spec.add_image("disk-0", image["id"])
+
+    userdata = yaml.safe_load(vm_spec.user_data)
+    userdata["ssh_authorized_keys"] = [pubkey]
+    vm_spec.user_data = yaml.dump(userdata)
+
+    _, data = api_client.vms.create(name, vm_spec)
+    deadline = datetime.now() + timedelta(seconds=timeout_secs)
+    while deadline > datetime.now():
+        _, data = api_client.vms.get(name)
+        if "Running" == data.get("status", {}).get("printableStatus"):
+            break
+        sleep(1)
+    else:
+        raise AssertionError(f"timed out waiting for {name} to transition to Running")
+
+    return name, image["user"]
+
+
+def delete_vm(name, api_client, timeout_secs):
+    code, data = api_client.vms.get(name)
+    if code == 404:
+        return
+
+    vm_spec = api_client.vms.Spec.from_dict(data)
+
+    api_client.vms.delete(name)
+    deadline = datetime.now() + timedelta(seconds=timeout_secs)
+    while deadline > datetime.now():
+        code, data = api_client.vms.get_status(name)
+        if 404 == code:
+            break
+        sleep(3)
+
+    for vol in vm_spec.volumes:
+        vol_name = vol["volume"]["persistentVolumeClaim"]["claimName"]
+        api_client.volumes.delete(vol_name)
+
+
+def start_vm(name, api_client, timeout_secs):
+    _, data = api_client.vms.start(name)
+    deadline = datetime.now() + timedelta(seconds=timeout_secs)
+    while deadline > datetime.now():
+        _, data = api_client.vms.get(name)
+        status = data.get("status", {}).get("printableStatus")
+        if "Running" == status:
+            return
+        sleep(1)
+
+    raise AssertionError(f"timed out trying to start {name}")
+
+
+def stop_vm(name, api_client, timeout_secs):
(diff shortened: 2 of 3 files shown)


connorkuehl:
I was originally waiting to plug this PR into the automated testing infrastructure but I'm still trying to get access to that, but it's taking a long time to get access to that infrastructure. I think this is reviewable in the meantime and the tests run to completion on my local machine anyway, so.

connorkuehl:
Latest revision removes the redundant backup test cases as requested :slightly_smiling_face:

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,609 @@
...
+
+
+@pytest.fixture(scope="class")
+def backup_source_vm(newvm_name, api_client, ssh_keypair, image, wait_timeout):
+    yield create_vm(newvm_name, api_client, ssh_keypair, image, wait_timeout)
+    delete_vm(newvm_name, api_client, wait_timeout)
+
+
+@pytest.fixture(scope="class")
+def vm_backup(vm_backup_name, backup_source_vm,
```
khushboo-rancher, line 229: Are these backup related fixtures required?

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,609 @@
...
+                else:
+                    break
+            else:
+                raise AssertionError(f"Unable to login to {name}")
+
+            with vm_sh as sh:
+                action(sh)
+
+
+@pytest.mark.p1
```
khushboo-rancher, line 353: Can we mark this P0 test case?

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,609 @@
...
+        3. A file called `test.txt` containing data: `123`.
+        4. A snapshot called `vm-snapshot`.
+
+        Since all of the heavy lifting is done in the fixture,
+        assert the snapshot was created successfully for this test.
+        """
+        code, data = api_client.vm_snapshots.get(vm_snapshot)
+
+        assert 200 == code
+        assert data.get("status", {}).get("readyToUse") is True
```
khushboo-rancher, line 379: These checks are already there in the fixture. :) 
Instead of having the logic of creating snapshot and checking the status etc, can we move those steps in the test. We can use the fixtures to create VM etc. The reason for the asked changes are as below:
1. By having the steps in the test will make it more readable.
2. We will isolate the test. As fixture can be used at many places and making a change to only this test will become difficult in future.

khushboo-rancher (changes requested):
Thanks @connorkuehl for the PR.

I feel we can make some changes to make test cases isolated and more readable by having the logic of test steps in the test instead of writing them in fixtures. We should definitely use fixture to write the prerequisite of tests and cleaning up them. I've tried to explain my point in one of the below comment. Let me know we can sync up to discuss this.

harvester_e2e_tests/integration/test_vm_snapshot.py
connorkuehl: No, I've removed them from the latest revision.

apiclient/harvester_api/managers.py
```diff
@@ -221,6 +221,14 @@ def export(self, name, image_name, storage_class, namespace=DEFAULT_NAMESPACE, *
         return self._create(path, params=params, json=export_spec, raw=raw)
 
 
+class VolumeSnapshotManager(BaseManager):
```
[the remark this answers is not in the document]
connorkuehl: The latest revision has `delete`, but implementing the spec for VolumeSnapshot will take a little more time. I've pushed the latest revision so we can get the ball rolling on reviewing the rest of the PR in the meantime.

harvester_e2e_tests/integration/test_vm_snapshot.py
connorkuehl: Updated this in the latest revision

harvester_e2e_tests/integration/test_vm_snapshot.py
connorkuehl: Incorporated this in the latest revision

connorkuehl:
@khushboo-rancher Thanks for the review! 🙂  I've incorporated most of the feedback with the exception of the create method for the VolumeSnapshotManager (see response in other thread), I'll keep working on it, but hopefully the rest of the PR is ready for another pass

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,517 @@
+from datetime import datetime, timedelta
+from time import sleep
+from paramiko.ssh_exception import ChannelException
+
+
+import pytest
+import yaml
```
lanfon72, line 7: based on [PEP8](https://peps.python.org/pep-0008/#imports)
import should be grouped as:
```python
import yaml
from time import sleep

from paramiko.ssh_exception import ChannelException
import pytest
```

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,517 @@
...
+        machine results in a virtual machine with the expected
+        well-known file (`test.txt`) with the expected contents
+        (`123`).
+
+        Prerequisites:
+        1. The source VM from the first test case and its
+           snapshot (`vm-snapshot`).
+        2. A VM restored from `vm-snapshot`.
+        """
+
```
lanfon72, line 292: I think this implement will not hit criteria, the VM created and took snapshot, but we didn't modify it after snapshot, so we can not guarantee the VM really restored from the snapshot or not.

and while restoring to new VM, we will not stop the source VM.

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,517 @@
...
+                                         ssh_keypair, host_shell,
+                                         vm_shell, wait_timeout):
+        name, ssh_user = source_vm
+
+        stop_vm(name, api_client, wait_timeout)
+
+        """
+        Test that the Harvester API rejects a `replace`
+        VirtualMachineRestore where the deletePolicy is
+        not `retain`.
```
lanfon72, line 316: we should separate this into another test case.

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,517 @@
...
+        assert wantmsg in reason
+        assert 422 == code
+
+        """
+        Test that the original virtual machine can be replaced
+        from its original snapshot (`vm-snapshot`) and that
+        the snapshot's data contains the well-known file (`test.txt`)
+        and its expected contents (`123`).
+
+        Prerequisites:
```
lanfon72, line 339: same as previous `restore to new`, we didn't modify the VM after snapshot, so we can't guarantee the VM really restored from the snapshot or not.

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_vm_snapshot.py
connorkuehl: 🤔 I understand this logic for the test_replace_vm_with_vm_snapshot, but I'm not sure I follow for this test case. Should I add more assertions to the test proving that this is accessing a completely separate/new VM?

lanfon72: In manual testing, we will test this as following steps:
1. Create the VM `vm1`
2. Write data into `vm1` (`vm1-state-1`)
3. Take snapshot for `vm1`
4. Write or modify exisiting data into `vm1` (`vm1-state-2`)
5. Restore snapshot into `vm1` or `vm1-new`
6. Data in restored VM should keep in `vm1-state1`, not `vm1-state-2`

khushboo-rancher:
@lanfon72 This is ready for review, could you please review it whenever get a chance?

harvester_e2e_tests/integration/test_vm_snapshot.py
```diff
@@ -0,0 +1,556 @@
...
+        `vm-snapshot` VM snapshot exists.
+        """
+        name, ssh_user = source_vm
+
+        stop_vm(name, api_client, wait_timeout)
```
lanfon72, line 380: I think the implementation is still not fit the test criteria, data in the `source_vm` is the same as the snapshot taken, so we can't say that the restore is success or not.

what we test a snapshot is like:
1. Create `vm1`
2. Write data into `vm1`, naming its state as `a`
3. Take snapshot `vm1-s` for `vm1`, so the snapshot is in same state `a` of the `vm1`
4. Write or delete data into `vm1`, naming its state as `b`
5. Restore snapshot `vm1-s` for `vm1`
6. `vm1` should back to state `a`

Steps 4 is the criteria for the test case, then we can guarantee that restore the snapshot is succeed because the state is changed.

For this implementation, Step 4 is missing, so we can't guarantee that the snapshot is restored or not.

lanfon72 (changes requested):
All related test cases should change (write or delete) data in VM before restore, to make sure the restore will recover into previous state.

harvester_e2e_tests/integration/test_vm_snapshot.py
lanfon72: We will not make the side effect in the test case `test_vm_snapshot_create`, the test case is purely require that create a snapshot, modify it after snapshot is not included.

Without the side effect, when we restore the VM into **new** one, we can simply compare files in both VMs to make sure it restored correctly.

And If we are going to verify restore existing VM, we would update the VM before restore to make sure what we changed will be reverted.

Based on this, we will not care about what we wrote into, so there will not have some constant assertion like `assert "123" in out`, because we will compare files in two VMs or check the modification is reverted.

connorkuehl: > We will not make the side effect in the test case test_vm_snapshot_create,

The code currently _does_ make this side-effect. Just to make sure I'm parsing your comment correctly, are you saying that you want me to remove that side-effect from the test case and then duplicate it across all of the other relevant test cases? I put it there initially because I didn't want every test case to have a connect, write, wait 2-seconds loop, which would extend the runtime of the test suite.

lanfon72: yes it is necessary.
When we are just execute `test_vm_snapshot_create`, what we will expects is, the VM having the same state as the snapshot been taken, modify the VM is not required and not expected.

For other test cases, we will not depends on the _side effect_ of data what we wrote, in **restore new**, we would compare two VMs having same files and content, not compare the constants from the `test_vm_snapshot_create`; in **restore existing**, we will just check what we wrote/deleted before restore, will be reverted, not compare the constants from the `test_vm_snapshot_create`.

In current implementation, we would need make the filename comes from `source_vm`, then the implementation would be more stable to extend when we are going to change the data what we are going to wrote, we would just need to store file names and passing them in test cases.

connorkuehl: Sorry for the delay! 🙂 

I think I've addressed this in the latest revision. The new test fixtures will overwrite contents prior to restore and assert that the restored or replaced VM have the original contents prior to the overwrite.

lanfon72 (approved):
LGTM, thx !

khushboo-rancher:
Thanks @connorkuehl and @lanfon72.
@connorkuehl could you please fix the `GH-Pages` error before we merge this?
```
Error: testcases/virtualmachines/scheduling.spec.ts:17:46 - error TS7006: Parameter 'host' implicitly has an 'any' type.

1[7](https://github.com/harvester/tests/actions/runs/6090066516/job/16524173011?pr=817#step:9:8)     const hostNames: string[] = hostList.map(host => host.name);
```

khushboo-rancher (approved):

PR #840, opened 2023-05-31
Enhance apis test images

albinsun:
Propose some enhancement during verify harvester#3889 [[ENHANCEMENT] Harvester Single Node Cluster](https://github.com/harvester/harvester/issues/3889)

## Changes
1. Put test generated cache & report to .gitignore, includes:
   1. .tox
   1. test_result.json
   1. .report.json

1. Enhance TC error due to delete an image which is Initialized but still **Importing** 
   * Test Suite: `apis/test_images.py::TestImages`
   * Problem
     1. `api_client.images.get` reply 200 as long as the image status is `Initialized`. But it's possible that Longhorn still `Importing` it (`Initialized` but not `Imported` yet)
        <details>
          <summary>Imported v.s. Importing status in images.get response body</summary>
        
          importing_and_imported
        </details>
     1. Due to this, `test_get` is passed since it consider upload (`test_create`) was succeeded, thus immediately trigger `test_delete` on the image which Longhorn is still importing.
     1. This causes image record is deleted on Harvester but still existed on Longhorn, and causes successive TC `test_create_with_reuse_display_name` fail (504) due to it use the same image name `unique_name` which is still processing).
        <details>
          <summary>Fail example on harvester-runtests#277</summary>

          test_images-harvester-runtests#277
        </details>
        
        <details>
          <summary>Image is deleted on Harvester but Longhorn is still processing.</summary>
        
          Screenshot_20230531_184652
          Screenshot_20230531_152020
        </details>

   * Solution
     * Confirm image DO `Imported` after `api_client.images.create_by_file`
     * Tested on `harvester-runtests#278`
     test_images-harvester-runtests#278

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,6 @@ terraform_test_artifacts/bin
 node_modules
 scripts/**/*.sh
 .vscode/
+.tox
+test_result.json
+.report.json
--- a/harvester_e2e_tests/apis/test_hosts.py
+++ b/harvester_e2e_tests/apis/test_hosts.py
@@ -153,6 +153,7 @@ def test_maintenance_mode(api_client, wait_timeout):
     status_code, node_stats = api_client.hosts.maintenance_mode(node_id, enable=True)
     assert 204 == status_code, (status_code, node_stats)
 
+    maintain_stat = ""
     endtime = datetime.now() + timedelta(seconds=wait_timeout)
     while endtime > datetime.now():
         _, stats = api_client.hosts.get(node_id)
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -63,6 +63,21 @@ class TestImagesNegative:
 @pytest.mark.p0
 @pytest.mark.images
 class TestImages:
+    def image_do_created(self, api_client, unique_name, wait_timeout):
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        while endtime > datetime.now():
+            code, data = api_client.images.get(unique_name)
+            name = data['metadata']['name']
+            reason = data['status']['conditions'][0]['reason']
+            if all([code == 200, name == unique_name, reason == "Imported"]):
+                break
+            sleep(5)
+        else:
+            raise AssertionError(
+                f"Failed to get image {unique_name} with {wait_timeout} timed out\n"
+                f"Still got {code} with {data}"
+            )
+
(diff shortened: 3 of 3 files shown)


khushboo-rancher (changes requested):
I'm fine with a temporary solution to avoid the failures. Just two things.
1. Could you please change the `image_do_created` name to something like `verify_image_created`.
2. Add a note that this is a temporary fix and being done to fix the #bugid where ever we are introducing `image_do_created` check.

albinsun:
> I'm fine with a temporary solution to avoid the failures. Just two things.
> 
> 1. Could you please change the `image_do_created` name to something like `verify_image_created`.
> 2. Add a note that this is a temporary fix and being done to fix the #bugid where ever we are introducing `image_do_created` check.

Refactored and tested on harvester-runtests#288
image

lanfon72:
I have the opposing opinion.

From the perspective of automation,
- We want discover bugs, not chasing **PASS rate**
- An automation test case reflect to a manual test case
- Those failures are **real** bug, not automation's bug.

From the perspective of test case
- A existing test case having specific scenario (from manual test case)
- We implement the automation test case to fit the scenario, **not** change the scenario to fit test case.


So, do we really need to add the change for those test cases?

image

When you look into the daily test failures, there have **8** failures, to exclude `scenarios`, others are similar and showing it is about **recreate images with same name**, and those test cases are **PASS** in the beginning of `v1.2.0`.

So, are we going to update the scenario? In my opinion, we shouldn't.  this failures reflect that we have a bug about **recreate images with same name**, and we should aware this, not change the automation's steps to let it **PASS**. (otherwise, the test case would be changed to another scenario)


> But how about `create_invalid_url`?  the scenario is not about **recreate images with same name**.

Yes, this is automation's bug, It failed because the side effect of `unique_name`, we reuse the name to create **invalid image**, so we should fix the side effect to let the test case to not affected by the (discovered) bug.

albinsun:
FYI, developer already puts the related issue to v1.2.0 scope.
https://github.com/harvester/harvester/issues/4027
We can also watching it first

khushboo-rancher:
Yeah, the [issue](https://github.com/harvester/harvester/issues/4027) is now scoped for 1.2.0. So, we'll wait for the fix.

@lanfon72 My point was that In general, if any low severity bug is impacting multiple test cases and not going to get fixed soon, we should consider to have a temporary fix so that we'll not miss running important test cases.

Issue #829, opened 2023-05-24
[BUG] Backupstore connection test fails on our CI


khushboo-rancher:
## What's the test to develop? Please describe

The test `test_backup_restore.py::TestBackupRestore::test_connection` is failing for both NFS and S3

```
AssertionError: Failed to update backup target to NFS with {'endpoint': 'nfs://10.84.102.1:/srv/nfsshare'}
E         API Status(400): {'kind': 'Status', 'apiVersion': 'v1', 'metadata': {}, 'status': 'Failure', 'message': 'admission webhook "validator.harvesterhci.io" denied the request: There is VMBackup in creating or deleting progress', 'reason': 'BadRequest', 'code': 400}
E       assert 200 == 400
E         +200
E         -400
```

khushboo-rancher:
This is still failing sometime.

khushboo-rancher:
It should be good if we fix this before 1.2.0 final rc, so that we know that backup restore is working fine.

khushboo-rancher:
The fix is merged, let's watch the result on CI for few days and then close it.

PR #850, opened 2023-06-07
* [e2e] fix timing issue while switching backup-target

lanfon72:
## Changes
- **ADD** retries while update `backup-target` settings

the root cause of #829 is, switching from S3 to another backup-target, cluster will need more time to handle the change, then automation will hit resource conflict error.

Consider that the resource conflict is not a bug in Harvester, and retries will not affect tests' scenario, so this change is not a temporarily workaround.

### DIFF
--- a/harvester_e2e_tests/integration/test_backup_restore.py
+++ b/harvester_e2e_tests/integration/test_backup_restore.py
@@ -29,6 +29,12 @@ pytest_plugins = [
 
 
 @pytest.fixture(scope="module")
+def conflict_retries():
+    # This might be able to moved to config options in need.
+    return 5
+
+
+@pytest.fixture(scope="module")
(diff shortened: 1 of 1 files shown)


lanfon72:
Tested on `harvester-runtests#303` with `v1.2-27934bc3-head`

#829 will not happen in `v1.2.0-rc1`, but newer versions of `v1.2-head`, but considering we didn't have any changes related to `backup-target` settings after `v1.2.0-rc1`, and the **resource conflict** error will only happen on the scenario of switch S3 to NFS, so it might not be Harvester bug. (but it still have a chance be the racing condition bug in Harvester)

Considering that retry in manual testing is very general and acceptable while encountering resource conflict,thus, this PR will add retry mechanism while changing `backup-target` setting.

khushboo-rancher (approved):
Good Analysis, Thanks @lanfon72

PR #857, opened 2023-06-11
add test cases of volume hot-plug for VM

lanfon72:
## Changes
- **ADD** [apiclient] `vm.add_volume` to hot plug volumes
- **ADD** [apiclient] `vm.remove_volume` to hot unplug volumes
- **ADD** `TestHotPlugVolume` for volume hot plug/unplug
    - new test cases: `test_add`, `test_remove`
    - new fixtures: `small_volume`
- **DEL** unused fixtures: `virtctl`, `kubeconfig_file`

### DIFF
--- a/harvester_e2e_tests/integration/test_vm_functions.py
+++ b/harvester_e2e_tests/integration/test_vm_functions.py
@@ -1,6 +1,4 @@
 from datetime import datetime, timedelta
-from pathlib import Path
-from tempfile import NamedTemporaryFile
 from time import sleep
 
 import json
@@ -17,25 +15,6 @@ pytest_plugins = [
 
 # GLOBAL Vars:
 MAX = 999999
-
-
-@pytest.fixture(scope="session")
-def virtctl(api_client):
-    code, ctx = api_client.vms.download_virtctl()
-
-    with NamedTemporaryFile("wb") as f:
-        f.write(ctx)
-        f.seek(0)
-        yield Path(f.name)
-
-
-@pytest.fixture(scope="session")
-def kubeconfig_file(api_client):
-    kubeconfig = api_client.generate_kubeconfig()
-    with NamedTemporaryFile("w") as f:
-        f.write(kubeconfig)
-        f.seek(0)
-        yield Path(f.name)
 
 
 @pytest.fixture(scope='module')
@@ -122,6 +101,19 @@ def image(api_client, image_opensuse, unique_name, wait_timeout):
 @pytest.fixture(scope="module")
 def unique_vm_name(unique_name):
     return f"vm-{unique_name}"
+
+
+@pytest.fixture(scope="class")
+def small_volume(api_client, unique_name):
+    vol_name, size = f"sv-{unique_name}", 3
+    vol_spec = api_client.volumes.Spec(size)
+    code, data = api_client.volumes.create(vol_name, vol_spec)
+
+    assert 201 == code, (code, data)
+
+    yield vol_name, size
+
+    code, data = api_client.volumes.delete(vol_name)
 
 
 @pytest.fixture(scope="class")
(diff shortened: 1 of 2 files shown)


lanfon72:
Tested on `harvester-runtests#320`

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -124,6 +103,19 @@ def unique_vm_name(unique_name):
...
+def small_volume(api_client, unique_name):
+    vol_name, size = f"sv-{unique_name}", 3
+    vol_spec = api_client.volumes.Spec(size)
+    code, data = api_client.volumes.create(vol_name, vol_spec)
+
+    assert 201 == code, (code, data)
+
+    yield vol_name, size
+
+    code, data = api_client.volumes.delete(vol_name)
```
albinsun, line 116: Do we need to check delete done to avoid affect TC later?

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -1591,3 +1583,188 @@ def test_vm_with_bogus_vlan(api_client, image, unique_vm_name,
...
+                   and "AgentConnected" == conds[-1].get('type')
+                   and data['status'].get('interfaces')):
+                    break
+            sleep(3)
+        else:
+            raise AssertionError(
+                f"Failed to Start VM({unique_vm_name}) with errors:\n"
+                f"Status: {data.get('status')}\n"
+                f"API Status({code}): {data}"
+            )
```
khushboo-rancher, line 1634: I feel, we should have a fixture `running_vm` instead of writing `starting vm` code in the test. As in future, many test cases will require running vm to start with.
Could you help to implement that?

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -1591,3 +1583,188 @@ def test_vm_with_bogus_vlan(api_client, image, unique_vm_name,
...
+            while endtime > datetime.now():
+                try:
+                    vm_sh.connect(vm_ip, jumphost=h.client)
+                except ChannelException as e:
+                    login_ex = e
+                    sleep(3)
+                else:
+                    break
+            else:
+                raise AssertionError(f"Unable to login to VM {unique_vm_name}") from login_ex
```
khushboo-rancher, line 1654: Same with checking the login to VM, we should have a common method. So that we can use is at multiple places and keep the test simple.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -1591,3 +1583,188 @@ def test_vm_with_bogus_vlan(api_client, image, unique_vm_name,
...
```
khushboo-rancher, line 1688: As mentioned above, repeated code for Login, we can replace this with common method.

khushboo-rancher (changes requested):
In general looks good. Some improvements.

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: we need to start the VM because we leverage existing `stopped_vm` which already defined, stopped_vm would be more useful when we are going to update the VM or especially **cloud-config**.

Of course we can add another `running_vm` fixture, but as you mentioned in next comment, it will add more lines rather than check the VM is started.

It is a trade off, in here, I would prefer reuse existing fixture and start the VM to make sure the VM just started and accessible.

khushboo-rancher: 1. Can you elaborate on how the `running_vm` will add more lines? I'm assuming a running_vm fixture will have the checks that the vm is started successfully.

2. Running_vm could be useful, quick and convenient to use for other test cases.

harvester_e2e_tests/integration/test_vm_functions.py
khushboo-rancher: It's not necessary that you create a common file and put the common methods there but you can create common methods required within a module at least.

For example, In the test `test_add` I see there are two attempts to login to a VM, we can avoid writing same lines of code again and have a method `login_to_vm`. Writing repeated code is hard to maintain in future, if something changes you need to modify same things at multiple places.

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: 2 lines for function naming and decorator, 1 lines for return, and lines same as the codes, is it the minimal lines to create the fixture and depends on stopped_vm

harvester_e2e_tests/integration/test_vm_functions.py
lanfon72: I think it is not necessary to emphasize that a test case will frequently read to addressing the fail is a real bug or  need a implement improvement, otherwise it will not be modified, so not  couple with others function should be the highest priority than code lines.

but yeah, it is just my opinion like #840 , so let’s follow your design.

khushboo-rancher: @lanfon72 It's not about your or my opinion but I recommend let's follow what is best for Harvester test. The goal of these tests is to have good coverage and nicely written cases which are easy to read and maintain.
I think the discussion here is not working. Let's discuss this offline.

lanfon72: I think we have different concept about automation testing. what is care about the test case is:
1. Test isolation
2. readability, including test reports
3. accurate and stability

I had explained when I migrating legacy code to current implementation, one of problems is, legacy code calling sharing functions which make tests not isolated.  And I [said](https://github.com/harvester/tests/pull/857#discussion_r1230854703) I don't have another better way to reduce repeated code for now.

In my opinion, in automation testing, repeated code is not harmful, sharing functions does.
I will not say something like _good_, _easy_, or _nicely_, because it will not explain anything.

> I think the discussion here is not working. Let's discuss this offline.

If I can't persuade to not do that, I will just follow your instructions.
I don't think discuss anything offline will solve the problem, same as #840 is not closed, in my experienced, it is not necessary and should be closed, and we discussed offline; same as #363, I had explained in the issue, but yeah, let's follow your instructions.

khushboo-rancher (approved):
LGTM

PR #859, opened 2023-06-12
Enhance integration for single node cluster

albinsun:
Propose 2 enhancements during verifying [[ENHANCEMENT] Harvester Single Node Cluster#3889](https://github.com/harvester/harvester/issues/3889)

1. **integration/test_backup_restore.py**
   * Add check after deleting VM backup to avoid config_backup_target too early.
   * Symptom
     * Sometimes config_backup_target fail due to "VMBackup still creating or deleting".
     * Example (harvester-runtests#316)
       image
   * tested on harvester-runtests#318
     image

2. **integration/test_hosts.py**
   * Fail fast on `test_host_poweroff_state` when there is no or only 1 node in cluster.
     > **Note**
     > `pytest.xfail()` is also an option, but seems not current convention.
     > Ref. https://docs.pytest.org/en/7.0.x/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail
   * Symptom
     * In single node profile, TC will power-off the only node.
   * Tested on harvester-runtests#319
     image

### DIFF
--- a/harvester_e2e_tests/integration/test_backup_restore.py
+++ b/harvester_e2e_tests/integration/test_backup_restore.py
@@ -98,7 +98,7 @@ def backup_config(request):
 
 
 @pytest.fixture(scope="class")
-def config_backup_target(api_client, conflict_retries, backup_config):
+def config_backup_target(api_client, conflict_retries, backup_config, wait_timeout):
     backup_type, config = backup_config
     code, data = api_client.settings.get('backup-target')
     origin_spec = api_client.settings.BackupTargetSpec.from_dict(data)
@@ -129,10 +129,27 @@ def config_backup_target(api_client, conflict_retries, backup_config):
     code, data = api_client.backups.get()
     assert 200 == code, "Failed to list backups"
 
+    check_names = []
     for backup in data['data']:
         endpoint = backup['status']['backupTarget'].get('endpoint')
         if endpoint != origin_spec.value.get('endpoint'):
             api_client.backups.delete(backup['metadata']['name'])
+            check_names.append(backup['metadata']['name'])
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        for name in check_names[:]:
+            code, data = api_client.backups.get(name)
+            if 404 == code:
+                check_names.remove(name)
+        if not check_names:
+            break
+        sleep(3)
+    else:
+        raise AssertionError(
+            f"Failed to delete backups: {check_names}\n"
+            f"Last API Status({code}): {data}"
+            )
 
 
 @pytest.fixture(scope="class")
@@ -216,10 +233,28 @@ def base_vm_with_data(
 
     # remove backups link to the VM and is ready
     code, data = api_client.backups.get()
+
+    check_names = []
     for backup in data['data']:
         if (backup['status'].get('readyToUse') and
                 unique_vm_name == backup['spec']['source']['name']):
             api_client.backups.delete(backup['metadata']['name'])
+            check_names.append(backup['metadata']['name'])
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        for name in check_names[:]:
+            code, data = api_client.backups.get(name)
+            if 404 == code:
+                check_names.remove(name)
+        if not check_names:
+            break
+        sleep(3)
+    else:
+        raise AssertionError(
+            f"Failed to delete backups: {check_names}\n"
+            f"Last API Status({code}): {data}"
+            )
 
     # remove created VM
     code, data = api_client.vms.get(unique_vm_name)
--- a/harvester_e2e_tests/integration/test_hosts.py
+++ b/harvester_e2e_tests/integration/test_hosts.py
@@ -1,5 +1,6 @@
 from time import sleep
 from datetime import datetime, timedelta
+from urllib.parse import urljoin
 
 import pytest
 
@@ -9,8 +10,10 @@ pytest_plugins = [
 
 
 @pytest.fixture(scope="session")
-def focal_image_url():
-    return "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img"  # noqa
+def focal_image_url(request):
+    base_url = request.config.getoption("--image-cache-url").strip()
+    base_url = base_url or "http://cloud-images.ubuntu.com/releases/focal/release"
+    return urljoin(f"{base_url}/", "ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img")
 
 
 @pytest.fixture(scope="module")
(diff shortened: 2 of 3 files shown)


harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -9,8 +10,12 @@
...
 
 @pytest.fixture(scope="session")
-def focal_image_url():
-    return "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img"  # noqa
+def focal_image_url(request):
+    base_url = request.config.getoption(
+        '--image-cache-url',
+        'http://cloud-images.ubuntu.com/releases/focal/release'
+        )
+    return urljoin(base_url, "openSUSE-Tumbleweed-NET-x86_64-Current.iso")  # noqa
```
lanfon72, line 18: I think this will not work in some case?
Ubuntu will not contains SUSE's tumbleweed.

And as this fixture is targeting for `focal_image_url`, the return value should not be other images.

harvester_e2e_tests/integration/test_backup_restore.py
```diff
@@ -132,7 +132,19 @@ def config_backup_target(api_client, conflict_retries, backup_config):
     for backup in data['data']:
         endpoint = backup['status']['backupTarget'].get('endpoint')
         if endpoint != origin_spec.value.get('endpoint'):
-            api_client.backups.delete(backup['metadata']['name'])
+            target_backup = backup['metadata']['name']
+            api_client.backups.delete(target_backup)
+            endtime = datetime.now() + timedelta(seconds=wait_timeout)
+            while endtime > datetime.now():
```
lanfon72, line 138: the check condition should move out of for-loop.
otherwise the maximum will multiply with wait_timeout as `backups * wait_timeout`

so the expected change would like:
- delete all backups
- check deleted backups is unavailable

Not:
- delete all backups
   - delete backup and check it is unavailable

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_hosts.py
albinsun: Oops, that's a typo, will fix it. thx!

albinsun:
Patch submitted
- tested on harvester-runtests#321
- confirm no data (backup... etc) are left after test done

image

albinsun (commented):
Patch submitted
- tested on harvester-runtests#321
- confirm no data (backup... etc) are left after test done

image

harvester_e2e_tests/integration/test_backup_restore.py
```diff
@@ -313,6 +351,7 @@ def test_restore_with_new_vm(
         with host_shell.login(host_ip, jumphost=True) as h:
             vm_sh = vm_shell(base_vm_with_data['ssh_user'], pkey=pri_key)
             endtime = datetime.now() + timedelta(seconds=wait_timeout)
+            login_ex = None
```
lanfon72, line 354: Could you explain the case about `login_ex` is not defined?
I think it is not possible as we will encountered at least once before
```python
raise AssertionError(...) from login_ex
```

otherwise we will never hit the timeout of while loop :thinking:

lanfon72 (changes requested):

harvester_e2e_tests/integration/test_backup_restore.py
lanfon72: I think it is lint's bug, when `ChannelException` not hit, we will `break` or encountered another exception, then the while loop will never execute `else` closure.

I guess the bug of lint might be, they will not detect the `else` closure of try-except, thus it will looks like the variable is unbound.

lanfon72: FYI: https://github.com/microsoft/pylance-release/issues/2507#issuecomment-1077662362

harvester_e2e_tests/integration/test_backup_restore.py
```diff
@@ -129,10 +130,27 @@ def config_backup_target(api_client, conflict_retries, backup_config):
...
     assert 200 == code, "Failed to list backups"
 
+    target_backup_names = []
     for backup in data['data']:
-        endpoint = backup['status']['backupTarget'].get('endpoint')
-        if endpoint != origin_spec.value.get('endpoint'):
-            api_client.backups.delete(backup['metadata']['name'])
+        backup_endpoint = backup['status']['backupTarget'].get('endpoint')
+        if backup_endpoint != origin_endpoint:
+            target_backup_names.append(backup['metadata']['name'])
```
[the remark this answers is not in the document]
albinsun: Fixed in latest commit.

harvester_e2e_tests/integration/test_backup_restore.py
```diff
@@ -129,10 +130,27 @@ def config_backup_target(api_client, conflict_retries, backup_config):
...
+
+    for backup_name in target_backup_names:
+        api_client.backups.delete(backup_name)
+
+    for backup_name in target_backup_names:
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        while endtime > datetime.now():
```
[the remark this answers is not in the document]
albinsun: Fixed in latest commit.

harvester_e2e_tests/integration/test_backup_restore.py
albinsun: Reverted warning fix in latest commit.

albinsun:
Fix committed, tested on harvester-runtests#323.

harvester_e2e_tests/integration/test_hosts.py
```diff
@@ -9,8 +10,12 @@
 
 
 @pytest.fixture(scope="session")
-def focal_image_url():
-    return "http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img"  # noqa
+def focal_image_url(request):
+    base_url = request.config.getoption(
+        '--image-cache-url',
+        'http://cloud-images.ubuntu.com/releases/focal/release'
+        )
```
lanfon72, line 17: I just look over again, there will have a potential bug, the default value of `--image-cache-url` is `''`, then `base_url = ''`
And the design of `urljoin`:
```python
>>> urljoin('http://abc/a', '123')
'http://abc/123'
>>> urljoin('http://abc/a', '/123')
'http://abc/123'
>>> urljoin('http://abc/a', '123')
'http://abc/123'
>>> urljoin('https://abc/a/', '123')
'https://abc/a/123'
>>> urljoin('https://abc/a//', '/123')
'https://abc/123'
>>> urljoin('https://abc/a/', '/123')
'https://abc/123'

```

ref: https://github.com/harvester/tests/blob/main/harvester_e2e_tests/fixtures/images.py#L18

lanfon72 (approved):
overall LGTM

harvester_e2e_tests/integration/test_hosts.py
albinsun: Nice comment, patch submitted referring to `images.py` and tested on harvester-runtests#326.

A quick manual check:
```
>>> def focal_image_url(base_url):
...    base_url = base_url.strip()
...    base_url = base_url or "http://cloud-images.ubuntu.com/releases/focal/release"
...    return urljoin(f"{base_url}/", "ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img")
... 
>>> focal_image_url("http://abc")
'http://abc/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img'
>>> focal_image_url("http://abc") == focal_image_url("http://abc/")
True

>>> focal_image_url("http://abc/a")
'http://abc/a/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img'
>>> focal_image_url("http://abc/a") == focal_image_url("http://abc/a/")
True

>>> focal_image_url("")
'http://cloud-images.ubuntu.com/releases/focal/release/ubuntu-20.04-server-cloudimg-amd64-disk-kvm.img'
>>> focal_image_url("") == focal_image_url(" ")
True
```

noahgildersleeve (approved):
LGTM. @albinsun  also be sure to remember to squash your commits before merging your PR.

khushboo-rancher (changes requested):
LGTM, Just one change as @noahgildersleeve suggested, we need to have either 1 or 2 commits.
1. 1 commit for test_backup_restore.py fix
2. 1 commit for test_host fix

Also, it is better to create a ticket for the bug in our test repo and put a PR against it.

albinsun:
Hi @khushboo-rancher and @noahgildersleeve,
Already squash commits into one.
Will pay attention then, thanks for your reminder.

khushboo-rancher (approved):

PR #864, opened 2023-06-15
Support on-demand unique_name

### LINKED ISSUE #861: [BUG]  Test 'apis/test_images.py::test_create_with_invalid_url' fails on CI

The test `apis/test_images.py::test_create_with_invalid_url` as we are using same image name which has been used by other test.

Solution:
Use a unique name for this test

albinsun:
## Changes
1. Add `gen_unique_name` fixture for TCs needs more thank 1 unique name.
2. Fix #861 by `gen_unique_name`.
   * **`apis/test_images.py`**, tested on harvester-runtests#336
     <details>
     
       image
     </details>
   * **`integration/test_images.py`**, tested on harvester-runtests#337
     <details>

       image
     </details>

### DIFF
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -162,7 +162,8 @@ class TestImages:
 @pytest.mark.p0
 @pytest.mark.negative
 @pytest.mark.images
-def test_create_with_invalid_url(api_client, unique_name, wait_timeout):
+def test_create_with_invalid_url(api_client, gen_unique_name, wait_timeout):
+    unique_name = gen_unique_name()
     code, data = api_client.images.create_by_url(unique_name, f"https://{unique_name}.img")
 
     assert 201 == code, (code, data)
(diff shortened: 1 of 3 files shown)


lanfon72 (approved):
In my opinion, `unique_name` is already a generated unique name for each module (test file), so I would prefer just update the `unique_name` in test case to make it not collected with other test cases like:
```python
def test_create_with_invalid_url(api_client, unique_name, wait_timeout):
    unique_name = f"invalid-url-{unique_name}"
# ...
```

but the implementation also fine to me.

albinsun:
> In my opinion, `unique_name` is already a generated unique name for each module (test file), so I would prefer just update the `unique_name` in test case to make it not collected with other test cases like:

Thank you!
Yes, I was plan to just update `unique_name` in TC, but I found it's possible to accidentally use the same name with some TC else, for example:
``` python
class TestBackendImages:
    def test_create_image_from_volume(...):
        ...
        image_name = f"image-{unique_name}"

    ...
    def test_create_invalid_file(...):
       unique_name = f"image-{unique_name}"
```
To avoid possible side effects, so the final proposal is to add `gen_unique_name` directly.

lanfon72:
> 

I will not say that naming collide is a side effect , you can either control the naming to use it more close to the test case to make it independently, or more pytest way, use function scope fixture for it, but yeah, return a generate function is also a way to do it, but it would be a bit leaving the purpose that we are using the unique name to address it happened with which test case in a specific time.

Issue #794, opened 2023-05-09
[TEST] Need configurations on Jenkins to run RKE1 & RKE2 integration test with particular versions


khushboo-rancher:
## What's the test to develop? Please describe

Need to have below configurations to select the rancher integration test version

- [ ] Run RKE1 integration test
      Version  - 
- [ ] Run RKE2 integration test
     Version -

lanfon72:
Could you explain this request in further, is asking to make Jenkins' jobs able to propagate versions into `config.yml`, or just add particular versions for specific Harvester version when run tests?

khushboo-rancher:
This has been done.

PR #877, opened 2023-07-03
[e2e] add configure options for RKE1 and RKE2

lanfon72:
## Changes
- **ADD** configure options: `RKE1-version`, `RKE2-version`
- **UPDATE** fixtures to use new configure options if available

ref to #794

### DIFF
--- a/harvester_e2e_tests/conftest.py
+++ b/harvester_e2e_tests/conftest.py
@@ -210,19 +210,19 @@ def pytest_addoption(parser):
     parser.addoption(
         '--kubernetes-version',
         action='store',
-        default=config_data.get('kubernetes-version', 'v1.24.11+rke2r1'),
+        default=config_data.get('kubernetes-version'),
         help='Kubernetes version to use for Rancher integration tests'
     )
     parser.addoption(
         '--RKE2-version',
         action='store',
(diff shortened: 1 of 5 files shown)


albinsun (approved):
LGTM

harvester_e2e_tests/conftest.py
```diff
@@ -210,9 +210,21 @@ def pytest_addoption(parser):
     parser.addoption(
         '--kubernetes-version',
         action='store',
-        default=config_data.get('kubernetes-version', 'v1.21.6+rke2r1'),
+        default=config_data.get('kubernetes-version', 'v1.24.11+rke2r1'),
```
khushboo-rancher, line 213: Can we just use the value from config.yml and not set a fallback value here? Reason being as below:
1. Kubernetes version keeps updating frequently, we'll have overhead of updating the versions at 2 places.
2. It is okay to let the test fail if versions are not defined properly in config.yml or we have some problem in parsing config.yml.
3. Let these values in the file be consistent with others as we are not putting fallback values for other variables.

lanfon72: updated.

khushboo-rancher (approved):

PR #899, opened 2023-07-24
* [e2e] fix incorrect function signature

lanfon72:


irishgordo (approved):
this looks good :smile: :+1:  - thanks for the fix @lanfon72

PR #910, opened 2023-08-07
[Cy] Remove cypress.env.json in .dockerignore

n313893254:
Remove cypress.env.json in .dockerignore as this file is rendered by pipeline, we need to docker build into image.


WuJun2016 (approved):
lgmt

PR #913, opened 2023-08-09
[e2e] fix fixture bugs and exclude tainted node

lanfon72:
## Changes
- **Update** test cases about migration to exclude tainted nodes while picking schedulable nodes
- **Update** fixture `cluster_network` in _test_0_storage_network_ to fix generate bug. (at least yield once before return)
- **Update** _test_hosts_ to fix reference bug. (while accessing data inside)

### DIFF
--- a/harvester_e2e_tests/integration/test_0_storage_network.py
+++ b/harvester_e2e_tests/integration/test_0_storage_network.py
@@ -32,7 +32,8 @@ def cluster_network(request, api_client, unique_name):
     all_nodes = set(n['id'] for n in data['data'])
     try:
         # vlad_nic configured on specific cluster network, reuse it
-        return next(cnet for cnet, nodes in cnet_nodes.items() if all_nodes == set(nodes))
+        yield next(cnet for cnet, nodes in cnet_nodes.items() if all_nodes == set(nodes))
+        return None
     except StopIteration:
         configured_nodes = reduce(add, cnet_nodes.values(), [])
         if any(n in configured_nodes for n in all_nodes):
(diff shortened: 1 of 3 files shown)


harvester_e2e_tests/integration/test_0_storage_network.py
```diff
@@ -41,6 +41,8 @@ def cluster_network(request, api_client, unique_name):
                 f"VLAN NIC configured nodes: {configured_nodes}\n"
                 f"All nodes: {all_nodes}\n"
             )
+    finally:
+        return None
```
albinsun, line 45: If `return` by `finally` here, wouldn't following code (47~74) will be reach?

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -437,7 +437,8 @@ def test_migrate(self, api_client, unique_vm_name, wait_timeout):
             f"{data}"
         )
 
-        new_host = next(h['id'] for h in host_data['data'] if cur_host != h['id'])
+        new_host = next(h['id'] for h in host_data['data']
+                        if cur_host != h['id'] and not h['spec'].get('taint'))
```
albinsun, line 441: It should be rare, but should we add a try-except block to log when get StopIteration when empty iterator (no available new host)?

lanfon72: I think this should be fine, when `next` the iterator, **StopIteration** will raised then this line (of code) will be highlight, and the variable already explain that it is about `new_host` which is empty. (by **StopIteration**)

Issue #905, opened 2023-08-01
[TC] Delete a volume which is in use


albinsun:
Delete a volume which is in use.

Ref. https://github.com/harvester/tests/issues/635

lanfon72:
is this duplicated with #635 ?

albinsun:
> is this duplicated with #635 ?

Hi @khushboo-rancher ,
Are these 2 TCs overlap #635 ?

khushboo-rancher:
That's okay, @noahgildersleeve is busy with some priority work, it is okay take these up. Thanks @albinsun

albinsun:
Daily test #691 passed 
![image](https://github.com/harvester/tests/assets/2773781/a13bdd85-f126-43ca-b02c-7b8664ef22f1)

PR #915, opened 2023-08-14
[TC] Delete volume bounded to VM

albinsun:
#### Enhancement
1. Add `polling_for` fixture to refactor repeated check loops

#### Implementation
1. Delete volume bounded on existing VM
   * Volume should NOT be able to delete when still bounded to existing VM
   * [[TC] Delete a volume which is in use #905](https://github.com/harvester/tests/issues/905)
2. Delete volume bounded on deleted VM
   * Volume should be able to delete when the bounded VM is deleted
   * [[e2e] [BUG] Volumes bounded on a deleted VM and unable to be removed #652](https://github.com/harvester/tests/issues/652)

#### Note
1. TCs are added in `test_volumes.py` since the testee here is volume.
2. Tested `integration/test_volumes.py` in harvester-runtests#365 using `v1.2.0-rc5`
   image

### DIFF
--- a/harvester_e2e_tests/integration/test_volumes.py
+++ b/harvester_e2e_tests/integration/test_volumes.py
@@ -1,8 +1,5 @@
 import yaml
 import pytest
-import polling2
-from time import sleep
-from datetime import datetime, timedelta
 from harvester_e2e_tests import utils
 from urllib.parse import urljoin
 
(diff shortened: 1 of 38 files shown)


khushboo-rancher (changes requested):
@albinsun Thanks for the PR, the utils.py is legacy code and we don't use it anymore. For more details check https://github.com/harvester/tests#contribute-
Could you please remove usage of utils.py

harvester_e2e_tests/integration/test_volumes.py
```diff
@@ -14,6 +15,95 @@
...
+    code, data = api_client.volumes.get(vol_name)
+    if 200 == code:
+        api_client.volumes.delete(vol_name)
+        # Wait until done so as not to affect image cleanup
+        endtime = datetime.now() + timedelta(seconds=wait_timeout)
+        while endtime > datetime.now():
+            code, data = api_client.volumes.get(vol_name)
+            if 404 == code:
+                break
+            sleep(sleep_timeout)
```
khushboo-rancher, line 104: Should we keep delete operation in try block? We might have test cases where we'll delete vms and volumes deleted as part of the test, in that case, will `api_client.vms.delete(vm_name)` and `api_client.volumes.delete(vol_name)` fail?

khushboo-rancher (commented):
LGTM, just one doubt

harvester_e2e_tests/integration/test_volumes.py
albinsun: Yes, TC may also have deletions, so I only **delete** if **get** returns 200.
But the delete-self still possible fail.
Just during refactoring, will also add a try block or rc verify on deletetion.

albinsun: Added RC check in latest commit

khushboo-rancher (approved):

PR #919, opened 2023-08-18
[CY]Ensure requests are intercepted before the page loads

WuJun2016:
Sometimes it happens that the page is already loaded, but cypress is still in intercept wait, I think this is a potential problem. (I don't have a way to provide a stable reproduction of the steps, it may have something to do with network speed)

image

The solution is to tell cypess which api to intercept before entering the page, in order to avoid Cypress not being able to intercept them even though the API request has already been completed

CI report: https://minio.provo.rancherlabs.com:31524/cypress-test-report/cypress/results/20230827-180209-842d76f9/index.html


n313893254 (approved):
LGTM, Thanks!

PR #932, opened 2023-09-06
Fix "implicitly has `any` type" lint

connorkuehl:
This was causing CI to fail on an unrelated PR.

Now it has an explicit `any` type :)

### DIFF
--- a/cypress/testcases/virtualmachines/scheduling.spec.ts
+++ b/cypress/testcases/virtualmachines/scheduling.spec.ts
@@ -14,7 +14,7 @@ describe('VM scheduling on Specific node', () => {
 
   it('Schedule VM on the Node which is Enable Maintenance Mode', () => {
     const hostList = Cypress.env('host');
-    const hostNames: string[] = hostList.map(host => host.name);
+    const hostNames: string[] = hostList.map((host: any) => host.name);
     const maintenanceNode = hostNames[0]
     const filterMaintenanceNames = hostNames.filter(name => name !== maintenanceNode);
 


connorkuehl:
and now it breaks CodeFactor. :(

I have zero experience with TypeScript or its ecosystem 😅 

@WuJun2016, @khushboo-rancher, do you know which type I'm supposed to plug in here? I presume this is coming from the Cypress env here: https://github.com/harvester/tests/blob/62f4aa8b4592626ee78c067a233fb28b15925d51/cypress/cypress.config.ts#L23 but it seems to be untyped JSON and different from the interface in `hosts.po.ts`.

khushboo-rancher:
I'm not sure, I think we can ignore this codeFactor check at the moment.

WuJun2016:
> and now it breaks CodeFactor. :(
> 
> I have zero experience with TypeScript or its ecosystem 😅
> 
> @WuJun2016, @khushboo-rancher, do you know which type I'm supposed to plug in here? I presume this is coming from the Cypress env here:
> 
> https://github.com/harvester/tests/blob/62f4aa8b4592626ee78c067a233fb28b15925d51/cypress/cypress.config.ts#L23
> 
> but it seems to be untyped JSON and different from the interface in `hosts.po.ts`.

You can change the code in the image below so that CodeFactor doesn't report errors, but I agree with @khushboo-rancher  suggestion that we ignore codeFactor checks for now.

image

khushboo-rancher (approved):

khushboo-rancher:
Merging this now as this is blocking https://github.com/harvester/tests/pull/817

PR #935, opened 2023-09-12
Teardown unbound LH backupvolumes

albinsun:
## Changes
1. Add teardown for unbound LH backupvolumes
   * Delete `lhbackupvolumes` not belong to any backup (generally deleted by backup teardown)
   * Manipulate via API `v1/harvester/longhorn.io.backupvolumes`.
   * Thus cleanup uncessary files on `backup-target`.
2. ~Enhance `polling_for` fixture and introduce into `integration/test_backup_restore.py`~
   * ~Reduce duplicated while check loops~

### Cleanup lhbackupvolumes
1. NFS Before: 330 files (343M)
   ```
   pcloud00 1# find nfsshare -type f | wc -l
   330
   pcloud00 1# du -sh nfsshare
   343M    nfsshare
   ```

1. Run TC
   * filter `NFS and integration and test_connection`
   * `harvester-runtests#367`
     image

1. NFS After: 11 files (44K)
   ```
   pcloud001 # find nfsshare -type f | wc -l
   11
   pcloud001 # du -sh nfsshare
   44K     nfsshare
   ```

### Does not break current tests
1. `integration/test_backup_restore.py`
   * `harvester-runtests#368`
     image

2. `integration/test_volumes.py`
   * `harvester-runtests#369`
     image

### DIFF
--- a/harvester_e2e_tests/integration/test_backup_restore.py
+++ b/harvester_e2e_tests/integration/test_backup_restore.py
@@ -124,6 +124,33 @@ def config_backup_target(api_client, conflict_retries, backup_config, wait_timeo
 
     yield spec
 
+    # remove unbound LH backupVolumes
+    code, data = api_client.lhbackupvolumes.get()
+    assert 200 == code, "Failed to list lhbackupvolumes"
+
+    check_names = []
+    for volume_data in data["items"]:
+        volume_name = volume_data["metadata"]["name"]
+        backup_name = volume_data["status"]["lastBackupName"]
+        if not backup_name:
+            api_client.lhbackupvolumes.delete(volume_name)
+            check_names.append(volume_name)
+
+    endtime = datetime.now() + timedelta(seconds=wait_timeout)
+    while endtime > datetime.now():
+        for name in check_names[:]:
+            code, data = api_client.lhbackupvolumes.get(name)
+            if 404 == code:
+                check_names.remove(name)
+        if not check_names:
+            break
+        sleep(3)
+    else:
+        raise AssertionError(
+            f"Failed to delete unbound lhbackupvolumes: {check_names}\n"
+            f"Last API Status({code}): {data}"
+            )
+
     # restore to original backup-target and remove backups not belong to it
     code, data = api_client.settings.update('backup-target', origin_spec)
     code, data = api_client.backups.get()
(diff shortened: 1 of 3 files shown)


albinsun:
Temporarily convert to draft.
Open https://github.com/harvester/harvester/issues/4529 to check whether or not this is a bug.

albinsun:
Per discussed in https://github.com/harvester/harvester/issues/4529#issuecomment-1720394098, this is indeed current LH behavior.

IMO, since not only NFS but also MinIO (S3) has those legacy files, I think we can use this teardown for now and discuss whether need to open an enhancemant to LH.

lanfon72:
Consider that unnecessary backup files will not affect test cases, so I would prefer that not clear them up would more align users behavior.

and for `test_backup_restore.py`, It is not necessary to re-organize or refactor test cases unless any fails or potential bug inside, this would reduce our effort to review codes then write more automation test cases.

albinsun:
> Consider that unnecessary backup files will not affect test cases, so I would prefer that not clear them up would more align users behavior.
> 
> and for `test_backup_restore.py`, It is not necessary to re-organize or refactor test cases unless any fails or potential bug inside, this would reduce our effort to review codes then write more automation test cases.

Thank you.
Basically I agree for the user behavior alignment.

I also agree that we should not refactor too much at a time.
But proper refactor can increase the maintainability for further extension. I think it would be good to do it gradually.

albinsun:
Hi @khushboo-rancher,
Need your comment for this ticket, thx.

harvester_e2e_tests/integration/test_backup_restore.py
```diff
@@ -124,6 +115,32 @@ def config_backup_target(api_client, conflict_retries, backup_config, wait_timeo
 
     yield spec
 
+    # remove unbound LH backupVolumes
```
albinsun, line 118: Since main parts of this PR is doing refactor.
For legacy backupvolume cleanup, please just refer line 118 to 143.

khushboo-rancher (approved):

khushboo-rancher:
>1. Should we need to open an enhancement to LH?
>2. Should we proceed to clean legacy metadata on NFS?
(one time clean is also a choice. Just run this feature branch in target env. w/o merging)

1. I think LH behavior is designed for a reason, we can handle volumebackup and backup as per their behavior. 
2. I see no harm to have the clean up code here as this is integration test, we are testing the individual functionality. So having unnecessary data can make debugging difficult.

albinsun:
Rebased to latest main branch.

Tested in `harvester-runtests#382`
image

PR #947, opened 2023-09-24
e2e rework: vm test cases

lanfon72:


lanfon72:
Tested on `harvester-runtests#386`

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -596,6 +613,287 @@ def test_create_stopped_vm(api_client, stopped_vm, wait_timeout):
...
+                                           for (cn, cd), (nn, nd) in zip(nodes_res, nodes_res[1:]))
+
+        # Calculate the maximum resource
+        vm_spec, namespace = api_client.vms.Spec(1, 2), 'default'
+        vm_spec.add_image("disk-0", image['id'])
+        data = vm_spec.to_dict(unique_vm_name, namespace)
+
+        if res_type == 'cpu':
+            exp = -2
+            data['spec']['template']['spec']['domain'][res_type]['cores'] = int(expected_res) + 1
```
albinsun, line 639: when `cd["cpu"] == nd["cpu"]` then `expected_res == nd["cpu"] == cd["cpu"]`, thus `int(expected_res) + 1` will exceed `cd["cpu"]`. is that expected?

lanfon72: The `expected_res` would be a `float`, and it would be used in **resources.limits** and **resources.requests**.
As the snapshot below, the `cores` would only expect an **Integer** and showing the number in OS (like `lscpu`), so that's why we do `int(expected_res) + 1`, we can limit/request VM's CPU be something like `11.3`, but OS would need show `12`.

image

harvester_e2e_tests/fixtures/virtualmachines.py
```diff
@@ -112,95 +136,253 @@ def __init__(self, vm_api, wait_timeout, snooze=3):
...
+                return False, ctx
+            return True, ctx
+
+        def wait_restarted(self, vm_name, endtime=None, callback=default_cb, **kws):
+            ctx = ResponseContext('vm.get_status', *self.vms.get_status(vm_name, **kws))
+            if 404 == ctx.code and callback(ctx):
+                return False, ctx
+
+            options = dict(old_pods=set(ctx.data['status']['activePods'].items()))
+            ctx = ResponseContext('vm.restart', *self.vms.restart(vm_name, **kws), options)
```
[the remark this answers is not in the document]
lanfon72: `VMChecker` is designed to check whether the VM entered the state during the `wait_timeout`.
For example,  `checker.wait_stopped` will return **(vm_entered_state, context)** to indicate the result, It is not related to test logic but eliminate polling steps.  (that's why we won't do `assertion` inside)

As it is designed to check the state change, it should be fine to use other methods to get the expected state, it is also one of the reason to use `Context` object instead of return a tuple directly, the context will indicate the finally status and which API be called.

and that's why we are not checking status code in **204** but **404**, because the user (caller) will only care about the final state (VM stopped), and when we call `vm.stop`, it might return 204 or 500, but the VM will stopped after all.

harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -596,6 +613,287 @@ def test_create_stopped_vm(api_client, stopped_vm, wait_timeout):
...
+        new_cpus = (int(out) + 1) + 1  # (shift to 1) + 1
+
+        # Update CPUs
+        code, data = api_client.vms.get(unique_vm_name)
+        assert 200 == code, (code, data)
+        spec = api_client.vms.Spec.from_dict(data)
+        spec.cpu_cores = new_cpus
+        code, data = api_client.vms.update(unique_vm_name, spec)
+        assert 200 == code, (code, data)
+        vm_restarted, ctx = vm_checker.wait_restarted(unique_vm_name)
```
albinsun, line 776: about the test logic i mean, for example here.
'restart' is a test step (thus we assert it), but the 'vms.restart' is executed in checker instead of TC self. 
it,s better to decouple responsibility so actions only in TC and checker only do validation.

lanfon72: We had tested all VM operations in `TestVMOperations`, so in other test cases, operations would be one of steps, not check points.

The checker is not designed for validation, as I explained earlier, it is used to observe the state change; and for `restart` state, observe the active pod of VMI is different is the current way to confirm the VM is restarted.
so when you look into the `wait_restarted`, it is coupled with the first result of `vm.get_status`.

vm_checker is a fixture to use the VMChecker _instance_ as a helper to observe state change, and VMChecker _class_ isn’t have any responsibility for test case.  The class is designed based on SOLID, SRP for the class is VM and State, methods are State observations.

You can imaging that the class be implemented in harvester_api, and we are just use it to help our testing.

albinsun (approved):
LGTM

PR #950, opened 2023-09-27
* [e2e] change file formating: CRLF -> LF

lanfon72:
To prevent bugs while editing this file. This PR will NOT change any content excepts line break.


albinsun (approved):
LGTM

PR #964, opened 2023-10-22
* [e2e] fix fixture type error bug

lanfon72:
To fix the default value type error.


albinsun (approved):
LGTM, please also comment test report # for reference.

PR #965, opened 2023-10-23
[e2e] fix bug in TestVMResource

lanfon72:
## Changes
1. Fix fixture `unset_cpu_memory_overcommit`
fix the side-effect from #964, when `overcommit-config` using default value, the response would be something like:
    ```python
    data = {'apiVersion': 'harvesterhci.io/v1beta1',
     'default': '{"cpu":1600,"memory":150,"storage":200}',
     'kind': 'Setting',
     'metadata': {'creationTimestamp': '2023-10-16T14:10:44Z',
                  'generation': 1,
                  'managedFields': [{'apiVersion': 'harvesterhci.io/v1beta1',
                                     'fieldsType': 'FieldsV1',
                                     'fieldsV1': {'f:default': {}, 'f:status': {}},
                                     'manager': 'harvester',
                                     'operation': 'Update',
                                     'time': '2023-10-16T14:10:44Z'}],
                  'name': 'overcommit-config',
                  'resourceVersion': '7286',
                  'uid': '3aeb7877-2edb-44c3-ad3d-f1f8538f6a39'},
     'status': {}}
    ```
    `overcommit-config` will not accept empty value for cpu/memory/storage, so we would need to use `default` value instead of a empty dict.
    and we would also need overwrite `spec.storage` for sure.
    The bug presented in `harvester-runtests#399` and `harvester-runtests#401`

2. Fix bug in `test_create_schedule_on_maximum`
Add one more expected condition when to check that there have schedulable hosts and the VM be scheduled on expected host.
3. Fix bug in `test_update_schedule_on_maximum`
Same as above and also add a teardown to prevent affect to following test cases. (the fixture `stopped_vm`)

### DIFF
--- a/harvester_e2e_tests/integration/test_vm_functions.py
+++ b/harvester_e2e_tests/integration/test_vm_functions.py
@@ -661,8 +661,14 @@ class TestVMResource:
                     f"VM started but not hosted on expected host: {expected_host}"
                 )
             else:
-                # ???: node's resources released while creating VM, so the VM be hosted on others
-                assert expected_host in [name for name, _ in schedulables], (
+                # ???: node's resources released while creating VM
+                expected_conditions = [
+                    # VM still be hosted on expected host
+                    expected_host == vmi['status']['nodeName'],
+                    # VM be hosted on others
+                    expected_host in [name for name, _ in schedulables],
+                ]
+                assert any(expected_conditions), (
                     f"VM started on another host {vmi['status']['nodeName']},"
                     f" and the resource of expected host {expected_host!r} be updated."
                 )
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/integration/test_vm_functions.py
```diff
@@ -660,8 +661,14 @@ def test_create_schedule_on_maximum(
                     f"VM started but not hosted on expected host: {expected_host}"
                 )
             else:
-                # ???: node's resources released while creating VM, so the VM be hosted on others
-                assert expected_host in [name for name, _ in schedulables], (
+                # ???: node's resources released while creating VM
+                expected_conditions = [
```
albinsun, line 665: If `vmi['status']['nodeName']` is the node which VM indeed running currently.
Should the condition be
`vmi['status']['nodeName'] in [name for name, _ in schedulables]`?

lanfon72: `schedulables` is containing host that still able to host a VM with maximum resource.

There have 4 conditions:
1. VM be scheduled on the expected host
    a. There still have hosts able to host the VM (`schedulables != []`)
    b. There have **NOT** hosts able to host the VM (`schedulables == []`)
2. VM be **NOT** scheduled on the expected host
    a. There still have hosts able to host the VM (`schedulables != []` and `expected_host in schedulables`)
    b. There have **NOT** hosts able to host the VM (`schedulables == []`)  **<= should not happen**

In here, we adding `1a` as the first `expected_conditions`.

PR #991, opened 2023-11-23
* [apiclient] fix default version bug

lanfon72:
ref to https://github.com/pypa/packaging/issues/321

`LegacyVersion` is removed after `packaging: 22.0` (the lib is using by `pkg_resources`), then the `parse_version` will raise execpetion rather than return a `LegacyVersion`, so change the default version value to `0.0.0`


albinsun (approved):
LGTM

PR #988, opened 2023-11-22
* update pytest-html config for deprecation

lanfon72:
For `pytest-html > 4.0` https://pytest-html.readthedocs.io/en/latest/deprecations.html#render-collapsed

Issue #970, opened 2023-10-27
[TEST] Move the daily e2e test run on ECM lab


khushboo-rancher:
## What's the test to develop? Please describe

Move the daily e2e test run on ECM lab

## Describe the items of the test development (DoD, definition of done) you'd like

- [x] Provisioning harvester cluster using seeder.
- [x] Update the corresponding Jenkins job  to trigger the e2e test on the cluster.
- [ ] harvester/harvester-baremetal-ansible#17

Related items

- [ ] Drop config/files about python 3.6 as it deprecated already.
      e.g. [py36_pytest.ini](https://github.com/harvester/tests/pull/993/files#diff-63895fc437abb610852330100574c644977cb1a88c18f6c1af90b48db2abf0ba)

lanfon72:
close as Jenkins has deployed on ECM lab, dynamically Rancher provisioning would need add pipeline so we would track it in https://github.com/harvester/harvester-baremetal-ansible/issues/17

PR #993, opened 2023-11-24
* fix pytest config in different python version

lanfon72:
Ref to #988 

To fix the `pytest-html >= 4.0` will not support `python3.6`, and the configure option is not compatible.

to verify this, do `tox --showconfig` and the `commands` should be rendered as the file located.

### DIFF
--- a/py36_pytest.ini
+++ b/py36_pytest.ini
@@ -1,0 +1,2 @@
+[pytest]
+render_collapsed = True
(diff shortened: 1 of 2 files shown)


albinsun (approved):
LGTM.

Note that we should drop those configs/files for python `3.6` after we completely move to python `3.8+`. So add 1 new work item on https://github.com/harvester/tests/issues/970#issue-1966024794

Issue #974, opened 2023-10-31
Copy of  [TEST] Add test cases to cover csi-driver & cloud provider deployment and their checks - RKE


albinsun:
## What's the test to develop? Please describe
Add test cases to cover csi-driver & cloud provider deployment and their checks

## Prerequisite and dependency of test
* External Rancher

### Test Case 
- [ ] Check harvester-cloud-provider
- [ ] Check harvester-csi-driver 
- [ ] Deploy nginx and load balancer
- [ ] Deploy nginx (w/ csi)

### Possible Refactor
1. ~Try to get supported K8s version from Rancher instead of hardcoded.~
   > From test aspect, better to know exact version we test.

## Reference
1. https://github.com/harvester/tests/issues/900

Issue #960, opened 2023-10-13
Refactor rancher_integration suite


albinsun:
- [ ] Refactor cluster_network/vm_network checker.
      https://github.com/harvester/tests/pull/949#discussion_r1348254880
- [ ] Refactor `ChartSpec` by `create_data`.
       https://github.com/harvester/tests/pull/995#discussion_r1415817175

khushboo-rancher:
@albinsun I think, this is okay to close?

albinsun:
all work items done, close it.

PR #995, opened 2023-11-30
Add testcase on RKE1 cluster

albinsun:
## Changes
1. **Adding** test cases for `harvester-cloud-provider` chart on RKE1 cluster
   * Install chart/app
   * Nginx deployment
   * Load-Balancer to Nginx deployment
3. **Adding** test cases for `harvester-csi-driver` chart on RKE1 cluster
   * Install chart/app
   * Deployment with PVC
   
### `harvester-runtests#422` (RKE1)
image
> [!NOTE]  
> Failed LB case which has 200 OK status but `<pending>` external-ip, which is an known issue https://github.com/harvester/harvester/issues/4689

### `harvester-runtests#422` (RKE2)
image

## Note
K3s TCs will be add later ([#974](https://github.com/harvester/tests/issues/974)) together with refactor ([#960](https://github.com/harvester/tests/issues/960))

### DIFF
--- a/apiclient/rancher_api/managers.py
+++ b/apiclient/rancher_api/managers.py
@@ -4,7 +4,7 @@ import yaml
 from weakref import ref
 from collections.abc import Mapping
 
-from .models import UserSpec
+from .models import UserSpec, ChartSpec
 
 
 DEFAULT_NAMESPACE = "default"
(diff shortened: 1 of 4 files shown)


apiclient/rancher_api/managers.py
```diff
@@ -346,6 +346,24 @@ def create(self, name, cluster_id, *, raw=False):
...
+        url = self.PATH_fmt.format(cluster_id=cluster_id)
+        if namespace:
+            url = f"{url}/{namespace}"
+            if name:
+                url = f"{url}/{name}"
+        return self._get(url, raw=raw)
+
+    def create(self, cluster_id, namespace, name, raw=False):
+        url = self.CREATE_fmt.format(cluster_id=cluster_id) + "?action=install"
+        data = ChartSpec(cluster_id, namespace, name).to_dict()
```
lanfon72, line 363: The Spec is not designed as harvester_api, it would be a bit confusing.
I would suggest that if we are not let `Spec` as class attribute have more  customization, we can simply use `create_data` instead.  (then there would not need the `ChartSpec`)

lanfon72 (approved):
In general LGTM.

apiclient/rancher_api/managers.py
albinsun: Thanks for the comment.
Already put it into refactoring workitem. Will fix in https://github.com/harvester/tests/issues/960#issue-1941028775

PR #1024, opened 2023-12-21
Multiple vms rancher integration

albinsun:
## Changes
1. Add 3 workers case on RKE1 and RKE2 test class.
   * but marked as skip temporarily since it may fails in slow disk I/O env. (see Ref. 1)
2. Drop unused `rancher-version` config option
   * there is no reference in current codebase.

### Skip (currently)
* RKE2 (`harvester-runtests#448`)
  image
* RKE1 (`harvester-runtests#450`)
  image

> [!NOTE]  
> Failed LB case which has 200 OK status but `<pending>` external-ip, which is an known issue https://github.com/harvester/harvester/issues/4689

### Not Skip (Drop skip mark)
image
<details>
  <summary>screenshot</summary>

  * RKE2
    image
  * RKE1
    image
</details>

### Reference
1. Compare provo and seeder env. 
   * provo (cluster is not Active over 1 hr)
     image
   * seeder (cluster Active normally)
     image
     image


khushboo-rancher:
As discussed with @albinsun, we can make the number of nodes configurable which will be integer list. So, based on the list multiple node (1, 3, etc) tests can run. 
cc: @lanfon72

lanfon72:
@albinsun, @khushboo-rancher, I feel that make a list for it might not so useful for our test scenarios.
In general, 1 and 3 VMs had duplicated test cases like **CSI driver**, **charts**, **providers** ... etc; for 3+ VMs cluster, what we must to test on is: cluster scaling (especially scale down then scale up on 3 VMs).
So make a list to test on different is duplicating several test cases, we should identify test cases that whether it is general (could be verify on 1/3 VMs) or for specific VMs only.

For this PR, I would suggest that user another config option (with default, make 1 or 3) to adjust `machine_count` but not run duplicated, If those test cases is general on RKE cluster with any nodes, verify once would be good enough.

albinsun:
> @albinsun, @khushboo-rancher, I feel that make a list for it might not so useful for our test scenarios. In general, 1 and 3 VMs had duplicated test cases like **CSI driver**, **charts**, **providers** ... etc; for 3+ VMs cluster, what we must to test on is: cluster scaling (especially scale down then up on 3 VMs). So make a list to test on different counts will duplicate test cases, we should identify test cases that whether it is general (could be verify on 1/3 VMs) or for specific VMs only.
> 
> For this PR, I would suggest that user another config option (with default, make 1 or 3) to adjust `machine_count` but not run duplicated, If those test cases is general on RKE cluster with any nodes, verify once would be good enough.

Hi @lanfon72,
Do your mean using a `machine_count` config option in `config.yml` instead of parameterized test?

lanfon72 (approved):
In general LGTM, we could use another PR to improve the config option for machine counts .

PR #1059, opened 2024-01-16
[e2e] change test case orders

lanfon72:
## Changes
- **ADD** README.md to explain the rule
- **UPDATE** renaming files


albinsun (approved):
LGTM

harvester_e2e_tests/integration/README.md
```diff
@@ -0,0 +1,23 @@
+# Naming Of Test Files
+To distinguish test cases with order, each file should start with `test_<N>_`, and the `N` is the number point out its dependencies of functionality, for example, `volume` and `image` will not dpened on anything, so the number should be `1`; for `vm`, it at least depends on `volume` and `image`, so it's number should be at least `3`.
+
+Here is the table for existing functionality:
+
+| Number | Function | Description |
+| ------ | -------- | ----------- |
+| z | Terraform | Fixed for terraform related test cases |
+| 9 | Rancher | Fixed number for external Rancher integration |
+| 0 | Settings | Any settings could be tested individually if not interact with others |
```
albinsun, line 10: Better to list table in real order: `test_[0-9]_*` then `test_[a-z]_*`.

lanfon72: Let's change terraform to `9` and move rancher + terraform to `scenarios`, WDYT?

albinsun: It's reasonable to rename `integration` to `integrations`.

For test cases classification, It's ok as longs as we have clear definition in https://github.com/harvester/tests/blob/fafe470aa72101c74f5bb62e342bf750e1621258/README.md?plain=1#L85 to distinguish `scenario` and `integration`.

PR #1073, opened 2024-01-23
Create fleet-support-with-harvester.md

khushboo-rancher:
Add test cases for fleet with Harvester support.

Test steps are not written yet, should be updated while testing.


albinsun (approved):
LGTM

PR #1109, opened 2024-02-19
[e2e] add metadata expose to test report

lanfon72:
Expose cluster metadata into **environment section** like:
image


albinsun (approved):
LGTM

irishgordo (approved):
looks good :smile:  - lgtm - having the metadata will be useful if we need to jump on the cluster to take a look at something

PR #1118, opened 2024-02-22
add fixture *skip_version_if*

lanfon72:
To let us skip tests more flexible.

we might need to consider support more complicated case like:
- Version only support in `v1.2` but not `v1.2.1`
- Version support both `v1.2` and `v1.3` but not `v1.3.1`

### DIFF
--- a/harvester_e2e_tests/fixtures/api_client.py
+++ b/harvester_e2e_tests/fixtures/api_client.py
@@ -198,6 +198,44 @@ def skip_version_after(request, api_client):
                 )
 
 
+@pytest.fixture(autouse=True)
+def skip_version_if(request, api_client):
+    ''' To mark test case should be skip when hit the condition string.
+
+    Args:
+        *args: Version string prefixing with one of operators: `!=`, `==`, `>=`, `<=`, `>`, `<`
+    Keyword Args:
+        reason: The reason string for `pytest.skip`, default is:
+            "Cluster Version `{cluster_version}` is not included in versions: {versions}"
+        condition: Condition callable function to check compare result(bool), default is `all`
+    '''
+    default_reason = (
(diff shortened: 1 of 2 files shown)


albinsun (approved):
LGTM, just few minor comments
1. Does it easier for us to use built-in `skipif`?
   https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-skipif
1. `skip_version_if` covers current `skip_version_*`
   May consider unify them since sometimes people have differen understanding of word `before/after` about the boundary and still need to check docstring. :P
   * `skip_version_if("> v1.1.1", "< 2.2.2"`)
   * `skip_version_if("> v1.1.1")`
   * `skip_version_if("< v2.2.2")`

harvester_e2e_tests/fixtures/api_client.py
```diff
@@ -198,6 +198,44 @@ def skip_version_after(request, api_client):
...
+            )
+
+
+def version_check(vstring, version):
+    from operator import le, lt, ge, gt, ne, eq
+    ops = {"<=": le, "<": lt, ">=": ge, ">": gt, "!=": ne, "==": eq}
+
+    try:
+        op = target_ver = None
+        op, target_ver = re.search(r"([<>=!]+)\s?(.+)", vstring).groups()
```
irishgordo, line 233: Would you be able to leave a comment as to what's taking place in this `re.search`?
It's a tad difficult to parse :sweat_smile:

irishgordo (approved):

harvester_e2e_tests/fixtures/api_client.py
lanfon72: we might not remove the marker before `v1.2` EOL (unless 4293 will have a backport fix).

PR #1140, opened 2024-03-04
* [e2e] fix pytest version

lanfon72:
Upgrade test `build#58` failed, after some investigating, it is caused by new version of `pytest`, so this PR will fix `pytest` version to be less than `v8.0` as we don't sure whether the plugin is used in somewhere and we are not actually using new pytest features.

image

### DIFF
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,4 +1,4 @@
-pytest
+pytest == 7.*
 pytest-drop-dup-tests
 pytest-ordering
 pytest-html


lanfon72:
The actual affect version should be `pytest 8.1.0` which just released today.

albinsun (approved):
LGTM

irishgordo (approved):
LGTM :smile: :+1:  - thanks for the version pinning - 

We may find that we would benefit from pinning all versions of all python dependencies.
Additionally pinning versions of Jenkins plugins too.

lanfon72:
pytest yanked `8.1.0` https://github.com/pytest-dev/pytest/issues/12069

but it should still be fine to fix our pytest version in `v7.x`.

PR #1166, opened 2024-03-18
[e2e] improve image fixtures

lanfon72:
## Changes
- **ADD** new config option _ubuntu-image-url_ to configure the specific ubuntu image for testing
- **UPDATE** fixtures `focal_image` to use the existing **image_ubuntu**
- **UPDATE** some test logic in _test_multiple_migrations_

This would let us to test x86/arm based images on demand.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -20,6 +20,7 @@ sleep-timeout: 3
 node-scripts-location: 'scripts/vagrant'
 
 opensuse-image-url: https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.3/images/openSUSE-Leap-15.3.x86_64-NoCloud.qcow2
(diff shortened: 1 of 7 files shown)


config.yml
```diff
@@ -20,6 +20,7 @@ sleep-timeout: 3
 node-scripts-location: 'scripts/vagrant'
 
 opensuse-image-url: https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.3/images/openSUSE-Leap-15.3.x86_64-NoCloud.qcow2
+ubuntu-image-url: https://cloud-images.ubuntu.com/releases/releases/22.04/release/ubuntu-22.04-server-cloudimg-arm64.img
```
khushboo-rancher, line 23: Do we want to use arm64 image by default?

lanfon72: I think no, let me update it.

khushboo-rancher (approved):

Issue #1127, opened 2024-02-26
[e2e] [bug] VM network interacting timing bug


lanfon72:
## What's the test to develop? Please describe
Test report in Provo lab's `harvester-runtests build#471`

- `harvester_e2e_tests/integrations/test_5_vm_networks_interact.py::TestBackendNetwork::test_mgmt_network_connection`
- `harvester_e2e_tests/integrations/test_5_vm_networks_interact.py::TestBackendNetwork::test_vlan_to_mgmt_connection`

## Describe the items of the test development (DoD, definition of done) you'd like
Test cases hit timing problem when/after VM rebooted.  It's the test case implementation bug.

lanfon72:
also happened on ECM lab's `harvester-install-and-test build#33`
- `integrations/test_5_vm_networks.py::TestVMNetwork::test_add_vlan`
- `integrations/test_5_vm_networks_interact.py::TestBackendNetwork::test_delete_vlan_from_multiple`
- `integrations/test_5_vm_networks_interact.py::TestBackendNetwork::test_mgmt_network_connection`
- `integrations/test_5_vm_networks_interact.py::TestBackendNetwork::test_vlan_to_mgmt_connection`

PR #474, opened 2022-08-29
fix e2e testing errors in August

lanfon72:
## Changes
- Update `utils.py` to fix:
    - subprocess' stdout/stderr message swapped (using by terraform)
    - reduce 1% maxmium CPU from the `most_available_cpu` to avoid VM hit the floor.

## Not Fix
Test cases under `apis/`, because we will re-work them. [^1]

### build 357
- [x] ~~scenarios/test_create_vm.py::test_create_windows_vm~~ **known issue:** https://github.com/harvester/harvester/issues/2535
- [x] ~apis/test_networks.py::test_create_network_using_terraform~ **known issue:** https://github.com/harvester/harvester/issues/2535
- [ ] apis/test_hosts.py::test_host_mgmt_maintenance_mode
- [ ] apis/test_hosts.py::test_host_reboot_maintenance_mode
- [x] ~~scenarios/test_create_vm.py::test_create_vm_do_not_start~~ **known issue:** https://github.com/harvester/harvester/issues/2677
- [x] scenarios/test_create_vm.py::test_create_vm_on_available_cpu_node
- [x] scenarios/test_create_vm.py::test_update_vm_on_available_cpu_node

### build 358
- [x] ~~scenarios/test_create_vm.py::test_create_vm_do_not_start~~ **known issue:** https://github.com/harvester/harvester/issues/2677

### build 359
- [x] ~~scenarios/test_create_vm.py::test_create_windows_vm~~ **known issue:** https://github.com/harvester/harvester/issues/2535

### build 360
- [x] ~apis/test_networks.py::test_create_network_using_terraform~ **Hard to reproduce**
- [ ] apis/test_hosts.py::test_host_reboot_maintenance_mode
- [ ] apis/test_hosts.py::test_host_poweroff_state
- [x] ~~scenarios/test_create_vm.py::test_create_vm_do_not_start~~ **known issue:** https://github.com/harvester/harvester/issues/2677
- [x] ~scenarios/test_create_vm.py::test_create_vm_on_available_cpu_and_memory_nodes~ **affected by previous test case (reboot/power-off node)**

### build 361
- [x] ~apis/test_networks.py::test_create_network_using_terraform~ **known issue:** https://github.com/harvester/harvester/issues/2535
- [x] ~~scenarios/test_create_vm.py::test_create_vm_do_not_start~~ **known issue:** https://github.com/harvester/harvester/issues/2677

### build 362
[...]

PR #1176, opened 2024-03-21
[e2e] fix machine type, ssh bugs

lanfon72:
## Changes
- **UPDATE** add fixture `machine_types` to adjust `pc` to `pc-q35` for newer Harvester version
- **UPDATE** add new check point for `test_update_node`
- **UPDATE** change to use `host_shell` and `vm_shell_from_host` to fix the bug of ssh to node using private key in our lab, to close https://github.com/harvester/tests/issues/1127

`add_vlan` already filed new issue in https://github.com/harvester/harvester/issues/5401

### DIFF
--- a/harvester_e2e_tests/integrations/test_3_vm_functions.py
+++ b/harvester_e2e_tests/integrations/test_3_vm_functions.py
@@ -174,6 +174,15 @@ def unset_cpu_memory_overcommit(api_client):
     api_client.settings.update('overcommit-config', spec)
 
 
+@pytest.fixture
+def machine_types(request, api_client):
+    major, minor, micro = api_client.cluster_version.release
+    if major > 1 or minor > 2 or (minor == 2 and micro > 1):
+        # > 1.y.z || > 1.3.z || > 1.2.1
+        return "_".join(request.param).replace("pc", "pc-q35").split("_")
+    return request.param
+
+
 @pytest.mark.p0
 @pytest.mark.virtualmachines
 @pytest.mark.dependency(name="minimal_vm")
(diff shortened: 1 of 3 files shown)


lanfon72:
Tested on both Provo lab `harvester-runtests#473, #474` (`v1.3.0`, `v1.2.1`) and ECM lab `harvester-runtests#7`(`v1.3.0`).
To notice that `update_vm_machine_types` in `v1.3.0` will failed it is because https://github.com/harvester/harvester/issues/5437

harvester_e2e_tests/integrations/test_3_vm_functions.py
```diff
@@ -174,6 +174,15 @@ def unset_cpu_memory_overcommit(api_client):
     api_client.settings.update('overcommit-config', spec)
 
 
+@pytest.fixture
+def machine_types(request, api_client):
+    major, minor, micro = api_client.cluster_version.release
+    if major > 1 or minor > 2 or (minor == 2 and micro > 1):
```
albinsun, line 180: A ignorable corner case is `0.3.0` 
```
>>> major, minor, micro = 0, 3, 0
>>> major > 1 or minor > 2 or (minor == 2 and micro > 1)
True
```

However, it think it will be more concrete if using `version_check` or `parse_version` explicitly.

lanfon72: Yeah but versions before v1.0.0 already EOL, I think we don't need to consider that.

If we use `parse_version` to check versions, the code will longer than this solution, as this might possibly happen in future, I thought this solution would better than use `parse_version`.

albinsun (approved):
LGTM

PR #1214, opened 2024-04-10
Add witness node manual test case

albinsun:
## Changes
1. Add manual test case for witness node feature introduced in `v1.3.0`

## Tickets
1. Feature: harvester/harvester#3266
1. Test: harvester/tests#999


TachunLin (approved):
LGTM.
Tested on the local hugo server, the content of image and format all display well.

lanfon72 (approved):

PR #1232, opened 2024-04-18
Fix wrong return code check in test_delete_volume_on_existing_vm

albinsun:
## Changes
1.Fix wrong status code check after deleting volume

image

### DIFF
--- a/harvester_e2e_tests/integrations/test_1_volumes.py
+++ b/harvester_e2e_tests/integrations/test_1_volumes.py
@@ -191,19 +191,11 @@ class TestVolumeWithVM:
 
         self.delete_vm(api_client, ubuntu_vm, polling_for)
 
-        code, data = api_client.volumes.delete(vol_name)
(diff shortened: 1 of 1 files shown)


lanfon72 (changes requested):
test steps for this which related to #652:

1. Create a VM `vm1` which having volume `vol1`
2. Delete `vm1` without delete volume `vol1`
3. Delete `vol1`
4. `vm1` should be deleted successfully
5. `vol1` should be deleted successfully

### Criteria
- Volume should be triggered to delete along with VM (the general behavior in UI)
- Volume should be deleting concurrently with VM, which means we would not check whether VM is deleted or not before delete the volume.
- When do delete volume, we might get `422` that it is attaching to VM, it's expected, and we will retry.

TachunLin:
Tested the execution on local v1.2-head Harvester.

image

albinsun:
Close this wrong implementation and will send another later.

PR #1233, opened 2024-04-18
test: Add Tests To Cover 4520

irishgordo:
* preform checks with image where image length is very long

Resolves: tests/issue-4520
See also: https://github.com/harvester/harvester/issues/4520

### DIFF
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -20,7 +20,7 @@ from time import sleep
 from datetime import datetime, timedelta
 
 import pytest
-
+from pytest_cases import parametrize_with_cases, case
 
 pytest_plugins = [
     "harvester_e2e_tests.fixtures.api_client",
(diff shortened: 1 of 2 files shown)


harvester_e2e_tests/apis/test_images.py
```diff
@@ -71,6 +72,15 @@ def test_create(self, api_client, unique_name, fake_image_file):
             f"Failed to upload fake image with error:{resp.status_code}, {resp.content}"
         )
 
+    def test_create_long_name(self, api_client, unique_name, fake_image_file):
+        resp = api_client.images.create_by_file(
+            "".join(list("".join([unique_name]*6))[0:43]), fake_image_file
```
albinsun, line 77: 1. Perhaps extract and add a module or class fixture `unique_name_long` and reuse here and in `test_get_and_delete_long_name`?

2. We can define `unique_name_long` fixture like:
   ```
   >>> uname
   'uname-123-456'
   >>> (uname*(43 // len(uname) + 1))[0:43]
   'uname-123-456uname-123-456uname-123-456unam'
   ```
   Thus we don't need to care about `unique_name` length.

harvester_e2e_tests/apis/test_images.py
```diff
@@ -83,6 +93,18 @@ def test_get(self, api_client, unique_name):
...
         assert unique_name == data['metadata']['name']
 
+    def test_get_and_delete_long_name(self, api_client, unique_name):
+        # Case 1: get all images
+        code, data = api_client.images.get()
+
+        assert len(data['items']) > 0, (code, data)
+
+        # Case 2: get created image with long name
+        code, data = api_client.images.get(("".join(list("".join([unique_name]*6))[0:43])))
```
lanfon72, line 103: The function signature of `join` is `iterable`, which means we can use string directly without convert it to **list**.
and the
```python
"".join(list("".join([unique_name]*6))[0:43])
```

would equals to

```python3
(unique_name*6)[:43]
```

irishgordo: Thanks for that mention :smile: - I guess I had forgot that join was unnecessary :sweat_smile:

harvester_e2e_tests/apis/test_images.py
irishgordo: I've force-pushed and updated a few things :smile:

irishgordo:
Introduced Pytest-Cases, as I think it would be a great candidate for flexible but complicated combination based testing.
For ease: 
[Pytest-Cases, Translated](https://smarie-github-io.translate.goog/python-pytest-cases/?_x_tr_sl=en&_x_tr_tl=zh-CN&_x_tr_hl=en&_x_tr_pto=wapp#why-pytest-cases)
[Pytest-Cases, Un-Translated](https://smarie.github.io/python-pytest-cases/#why-pytest-cases)

Screenshot from 2024-04-25 13-06-47

harvester_e2e_tests/apis/test_images.py
```diff
@@ -18,6 +18,7 @@
 
 from time import sleep
 from datetime import datetime, timedelta
+from pytest_cases import parametrize_with_cases, case
```
lanfon72, line 21: 3rd party library should be placed in 2nd section. (PEP8)

lanfon72 (approved):
In general LGTM

harvester_e2e_tests/apis/test_images.py
irishgordo: Updated :smile: :+1:

albinsun:
> Introduced Pytest-Cases, as I think it would be a great candidate for flexible but complicated combination based testing. For ease: [Pytest-Cases, Translated](https://smarie-github-io.translate.goog/python-pytest-cases/?_x_tr_sl=en&_x_tr_tl=zh-CN&_x_tr_hl=en&_x_tr_pto=wapp#why-pytest-cases) [Pytest-Cases, Un-Translated](https://smarie.github.io/python-pytest-cases/#why-pytest-cases)
> 
> Screenshot from 2024-04-25 13-06-47
> 
> cc: @albinsun @lanfon72
> 
> Hoping to gather your thoughts 🤔 😄

LGTM, this plug-in seems actively contributed and documented well!

albinsun (approved):

PR #1247, opened 2024-04-26
Sort report with execution time and add Time field in report

albinsun:
From recent discussion, I think we can enhance report by sorting with execution order thus helps to get the context of skipped & failed TCs.

## Changes
1. Sort report with execution time
   Ref. https://pytest-html.readthedocs.io/en/latest/user_guide.html#results-table-sorting
2. Add Time field in report
   Ref. https://pytest-html.readthedocs.io/en/latest/user_guide.html#modifying-the-results-table

### Before: Sort by `result` (pytest-html default)
report-before

### After: Sort by `original` (best effort as the order of execution)
report-after
You can sort with new added **Time** field explicitly
image

### DIFF
--- a/harvester_e2e_tests/conftest.py
+++ b/harvester_e2e_tests/conftest.py
@@ -17,6 +17,7 @@
 
 import pytest
 import yaml
+from datetime import datetime
 from pytest_dependency import DependencyManager as DepMgr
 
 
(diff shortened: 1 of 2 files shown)


lanfon72 (approved):
It looks the `Time` is end time of execution, should we change the `Time` to `EndTime`?

PR #1248, opened 2024-04-29
[e2e] fix upgrade bugs

lanfon72:
## Changes
- **REMOVE** intercept method to shut down VMs during upgrade, as https://github.com/harvester/harvester/issues/5076 been applied
- **ADD** checker to check VM is created before check VMI is existence
- **ADD** TimeoutError to exceptions

### DIFF
--- a/harvester_e2e_tests/integrations/test_upgrade.py
+++ b/harvester_e2e_tests/integrations/test_upgrade.py
@@ -1,7 +1,6 @@
 import re
 import json
 import yaml
-import socket
 from time import sleep
 from operator import add
 from functools import reduce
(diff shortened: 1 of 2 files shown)


harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -240,22 +239,6 @@ def check(self, data):
             for func in self.intercepts():
                 func(data)
 
-        def intercept_v121_vm(self, data):
-            if "v1.2.1" != api_client.cluster_version.raw:
-                return
-            if self._v121_vm:
```
albinsun, line 246: Perhaps also drop class variable `_v121_vm` (Line 232) since it's only referenced by `intercept_v121_vm`?

albinsun (approved):
In general LGTM.

PR #1269, opened 2024-05-16
Correct test flow for TC delete_volume_on_deleted_vm

albinsun:
## Changes
1.Fix wrong test flow for TC `test_delete_volume_on_deleted_vm`.

## Verification
`harvester-runtests/8` on ECM lab
image

### DIFF
--- a/harvester_e2e_tests/integrations/test_1_volumes.py
+++ b/harvester_e2e_tests/integrations/test_1_volumes.py
@@ -178,32 +178,26 @@ class TestVolumeWithVM:
     def test_delete_volume_on_deleted_vm(self, api_client, ubuntu_image, ubuntu_vm, polling_for):
         """
         1. Create a VM with volume
-        2. Delete volume should reply 422
-        3. Delete VM but not volume
-        4. Delete volume should reply 200
+        2. Delete VM but not volume
+        3. Delete volume concurrently with VM
+        4. VM should be deleted
+        5. Volume should be deleted
         Ref. https://github.com/harvester/tests/issues/652
         """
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/integrations/test_1_volumes.py
```diff
@@ -178,32 +178,26 @@ def test_delete_volume_on_existing_vm(self, api_client, ubuntu_image, ubuntu_vm,
...
-        code, data = api_client.volumes.delete(vol_name)
-        assert 422 == code, f"Should fail to delete volume\n{code}, {data}"
-
-        self.delete_vm(api_client, ubuntu_vm, polling_for)
+        api_client.vms.delete(vm_name)
 
-        code, data = api_client.volumes.delete(vol_name)
-        assert code in (200, 204), f"Fail to delete volume\n{code}, {data}"
+        polling_for("Delete volume",
+                    lambda c, d: c and c != 422,
```
lanfon72, line 194: what's the different to check `200` and `422`?

albinsun: We can catch the case that fail to delete due to still attaching on VM.

albinsun: Per discussion, `422` seems the coupled part between VM and volume. But yeh, I think we can just assert `200` for more explicit logic.

lanfon72: I feel that `422` would need a comment to explain it is going to check that the volume is detached, `200` would simply self-explained that the volume is deleted successfully.

lanfon72 (approved):

PR #1278, opened 2024-05-20
fixed some out of date links and added in a couple of scenarios.

noahgildersleeve:
This fixed some old doc links and added in some updated element to out of date tests


albinsun (approved):
LGTM
Better to have screentshot in manual TC.

PR #1307, opened 2024-06-03
Update manual TCs steps according release testing

albinsun:
## Changes
1. TC `Negative Power down the node where the VM is getting replaced by the restore`
   * Add detail steps
3. TC `Verify that VMs stay up when disks are evicted`
   * Correct step order
   * Storage class's replica count should be `3` in this context


lanfon72 (approved):
LGTM

TachunLin (approved):
LGTM

PR #1318, opened 2024-06-10
config: allow user-config

m-ildefons:
Allow a user to maintain a config file in their home directory, which overrides the default settings.
This makes it possible to execute the test suite from a clean git repo state and reduces the probability that config changes are accidentally committed to the repo.

### DIFF
--- a/harvester_e2e_tests/conftest.py
+++ b/harvester_e2e_tests/conftest.py
@@ -50,41 +50,35 @@ def check_depends(self, depends, item):
 DepMgr.checkDepend, DepMgr._check_depend = check_depends, DepMgr.checkDepend
 
 
-def merge_config(user_config_path, default_config_path):
+def merge_config():
     """
     Merge two config files. The user config takes precedence over the default
     config.
     """
 
-    try:
-        with open(user_config_path, 'r') as ucnf:
-            user_data = yaml.safe_load(ucnf)
-    except FileNotFoundError:
-        user_data = {}
-
-    try:
-        with open(default_config_path, 'r') as dcnf:
-            default_data = yaml.safe_load(dcnf)
-    except FileNotFoundError:
-        default_data = {}
+    paths = [
+        os.path.join(os.getcwd(), 'config.yml'),
+        os.getenv('HARVESTER_TESTS_CONFIG', None),
+        os.path.join(os.getenv('HOME'), '.config', 'harvester', 'tests',
+                     'config.yml')
+    ]
 
     config = {}
-    for key in default_data:
-        config[key] = user_data.get(key, default_data[key])
+    for path in paths:
+        if path:
+            try:
+                with open(path, 'r') as cnf:
+                    data = yaml.safe_load(cnf)
+            except FileNotFoundError:
+                data = {}
 
-    for key in user_data:
-        if not key in default_data.keys():
-            config[key] = user_data[key]
+            config.update(data)
 
     return config
 
 
 def pytest_addoption(parser):
-    user_home = os.getenv('HOME')
-    user_config_path= os.path.join(user_home, '.config', 'harvester', 'tests',
-                                   'config.yml')
-    default_config_path = os.path.join(os.getcwd(), 'config.yml')
-    config_data = merge_config(user_config_path, default_config_path)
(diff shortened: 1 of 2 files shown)


harvester_e2e_tests/conftest.py
```diff
@@ -49,9 +50,42 @@ def check_depends(self, depends, item):
...
+            config[key] = user_data[key]
+
+    return config
+
+
 def pytest_addoption(parser):
-    with open('config.yml') as f:
-        config_data = yaml.safe_load(f)
+    user_home = os.getenv('HOME')
+    user_config_path= os.path.join(user_home, '.config', 'harvester', 'tests',
```
albinsun, line 84: IMO it's better to use an env. vairable like `HVST_TESTS_CONFIG_PATH` to give user the flexibility instead of fixed path.
e.g. `user_config_path = os.getenv('HVST_TESTS_CONFIG_PATH')`

harvester_e2e_tests/conftest.py
```diff
@@ -49,9 +50,42 @@ def check_depends(self, depends, item):
...
+        default_data = {}
+
+    config = {}
+    for key in default_data:
+        config[key] = user_data.get(key, default_data[key])
+
+    for key in user_data:
+        if not key in default_data.keys():
+            config[key] = user_data[key]
+
```
albinsun, line 78: May be more pythonic?
```
>>> dcnf = {'A': 1, 'B': 2}
>>> ucnf = {'a': 10, 'B': 20, 'c': 30}
>>> ucnf | dcnf                         # python 3.9+
{'a': 10, 'B': 2, 'c': 30, 'A': 1}
>>> 
>>> {**ucnf, **dcnf}                    # python 3.5+
{'a': 10, 'B': 2, 'c': 30, 'A': 1}
```
Ref. https://stackoverflow.com/questions/38987/how-do-i-merge-two-dictionaries-in-a-single-expression-in-python

harvester_e2e_tests/conftest.py
m-ildefons: Good point. I've added the variable `HARVESTER_TESTS_CONFIG` as possible location for the config file.
I still want it to look in a default location outside of the repository, since this way dropping a file somewhere is enough and the user doesn't need to set an environment variable as well.
I've also put a short line in the readme to explain all this.

harvester_e2e_tests/conftest.py
```diff
@@ -421,3 +449,18 @@ def pytest_html_results_table_header(cells):
 
 def pytest_html_results_table_row(report, cells):
     cells.insert(1, f'<td class="col-time">{datetime.utcnow()}</td>')
+
+
+def pytest_report_header(config):
```
lanfon72, line 454: I don't think it is a good idea to expose sensitive data, it is not necessary for the report.

lanfon72 (commented):
make it as a OS environment variable is not a good idea, it is not explicitly point out where the configuration comes from.

we can simply add new configure option `--config` and let the **default value** be `config.yml` would gracefully resolve the config file name problem.

for me, I will not accept any OS environment or implicit variables injection, because we are not making a tool, it is for test, everything should be explicitly.

albinsun:
> we can simply add new configure option `--config` and let the **default value** be `config.yml` would gracefully resolve the config file name problem.

Agree, it's more explicit than env. variable approach.

lanfon72:
> > we can simply add new configure option --config and let the default value be config.yml would gracefully resolve the config file name problem.
> 
> This is **not** simple at all.
> 
> The way it currently works is that we load the config data first and then populate the _default values_ for the command line options with the data loaded from the config file. _Then_ the command line is parsed and the final `pytest.config.Config` object is constructed.
> 
> Having a command line option that defines which file to load the config data from would mean the that we would need to _update_ the final `pytest.config.Config` object, since the command line parsing would need to happen first.

we would not need other command line options, to let the config filename configurable, we will remove other options, they would only be update in the config file.

khushboo-rancher (changes requested):
Idea looks good to me. Are we logging/printing the config path anywhere? Having the config path in report or in console log will help.

Also, agree to the point, let's not print the sensitive data in the test report.

PR #1323, opened 2024-06-13
Fix dependencies

torchiaf:
- Fixed adm-zip dependency
- Moving @types/cypress in dev-dependencies


a110605 (approved):
LGTM. 
adm-zip v0.5.13 release had [methods/inflater.js](https://github.com/cthackers/adm-zip/commit/689b30024bf40e6f9cf0068f46f17105a9a84168) changed causes the issue we encountered in cypress.

 Pin to the version lower than 0.5.13 works.

PR #1341, opened 2024-06-25
[e2e] implement vm template related function and test case

### LINKED ISSUE #1194: [e2e] [BUG] Harvester v120 cann't start new VM from saved VM template

A clear and concise description of what the test you want to develop.

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

- [ ] `item 1`

## Additional context

Add any other context or screenshots about the test request here.

related issue: harvester/harvester#4589

lanfon72:
## Changes
- **ADD** TemplateSpec
- **ADD** `create_template` in VM manager
- **UPDATE** TemplateManager to use TemplateSpec
- **UPDATE** VMSpec to load volume claim correctly
- **UPDATE** existing vm template test case structure
- **ADD** `test_create_vm_from_template_with_data` to close #1194

### DIFF
--- a/apiclient/harvester_api/managers/virtualmachines.py
+++ b/apiclient/harvester_api/managers/virtualmachines.py
@@ -34,14 +34,14 @@ class VirtualMachineManager(BaseManager):
 
     def create(self, name, vm_spec, namespace=DEFAULT_NAMESPACE, *, raw=False):
         if isinstance(vm_spec, self.Spec):
-            vm_spec = vm_spec.to_dict(name, namespace)
+            vm_spec = self.Spec.to_dict(vm_spec, name, namespace)
         path = self.PATH_fmt.format(uid="", ns=namespace)
         return self._create(path, json=vm_spec, raw=raw)
 
     def update(self, name, vm_spec, namespace=DEFAULT_NAMESPACE, *,
                raw=False, as_json=True, **kwargs):
         if isinstance(vm_spec, self.Spec):
-            vm_spec = vm_spec.to_dict(name, namespace)
+            vm_spec = self.Spec.to_dict(vm_spec, name, namespace)
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         return self._update(path, vm_spec, raw=raw, as_json=as_json, **kwargs)
 
@@ -113,4 +113,13 @@ class VirtualMachineManager(BaseManager):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         json = dict(diskName=disk_name)
         params = dict(action="removeVolume")
+        return self._create(path, params=params, json=json, raw=raw)
+
+    def create_template(
+        self, name, template_name, keep_data=False, description="", namespace=DEFAULT_NAMESPACE,
+        *, raw=False
+    ):
+        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
+        json = dict(description=description, name=template_name, withData=keep_data)
+        params = dict(action="createTemplate")
         return self._create(path, params=params, json=json, raw=raw)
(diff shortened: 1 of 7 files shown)


albinsun (commented):
Having a trial run on `harvester-runtests#15` but hit fail
image
Here `data['spec']['defaultVersionId']` is an empty string.
Do we miss something here?

apiclient/harvester_api/managers/virtualmachines.py
```diff
@@ -34,14 +34,14 @@ def get_status(self, name="", namespace=DEFAULT_NAMESPACE, *, raw=False, **kwarg
...
-            vm_spec = vm_spec.to_dict(name, namespace)
+            vm_spec = self.Spec.to_dict(vm_spec, name, namespace)
         path = self.PATH_fmt.format(uid="", ns=namespace)
         return self._create(path, json=vm_spec, raw=raw)
 
     def update(self, name, vm_spec, namespace=DEFAULT_NAMESPACE, *,
                raw=False, as_json=True, **kwargs):
         if isinstance(vm_spec, self.Spec):
-            vm_spec = vm_spec.to_dict(name, namespace)
+            vm_spec = self.Spec.to_dict(vm_spec, name, namespace)
```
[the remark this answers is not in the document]
lanfon72: It's the way to cast variable `vm_spec` to class `VMSpec`, as the `TemplateSpec` is derived class of `VMSpec` and `to_dict` is modified to return another data structure.

The implementation is following the concept of Liskov substitution principle, object of derived class should able to cast and be used in base class, vice versa.  That's why the document of `isinstance` is:
```
isinstance(obj, class_or_tuple, /)
    Return whether an object is an instance of a class or of a subclass thereof.
    
    A tuple, as in ``isinstance(x, (A, B, ...))``, may be given as the target to
    check against. This is equivalent to ``isinstance(x, A) or isinstance(x, B)
    or ...`` etc.
```

lanfon72:
update to fix the timing bug in checking template version

harvester_e2e_tests/integrations/test_4_vm_template.py
```diff
@@ -0,0 +1,195 @@
...
+            if 404 == code:
+                break
+            sleep(3)
+
+        for vol in tmpl_spec.volumes:
+            vol_name = vol['volume']['persistentVolumeClaim']['claimName']
+            api_client.volumes.delete(vol_name)
+
+        code, data = api_client.templates.delete(unique_name)
+        assert 200 == code, (code, data)
```
albinsun, line 195: Please confirm this DELETE do response `200`
https://github.com/harvester/tests/blob/97cbba4202e7b233bde6e277ea5d962af3de630a/apiclient/harvester_api/managers/templates.py#L8

Although UI uses different `v1/harvester/harvesterhci.io.virtualmachinetemplates/default/test` but it replies `204`
image

albinsun (approved):
LGTM

harvester_e2e_tests/integrations/test_4_vm_template.py
lanfon72: the endpoint starts with `v1/harvester` is Harvester customized APIs, it will return `204` in DELETE operation, the usual endpoint in `apis` will return `200` with context like:
```
{'kind': 'Status', 'apiVersion': 'v1', 'metadata': {}, 'status': 'Success', 'details': {'name': 'u1', 'group': 'harvesterhci.io', 'kind': 'virtualmachinetemplates', 'uid': '5eda6fff-12ff-44e1-acab-632493a0fc25'}}
```

customized might cause unexpected behavior like https://github.com/harvester/harvester/issues/4641, so we would usually use endpoints in `apis/` (and as you can see, there would have version *v1beta1* in apis, but `v1/harvester` haven't).

PR #1350, opened 2024-07-01
[e2e] add new test cases for addons

lanfon72:
## Changes
- **ADD** AddonManager and model
- **ADD** basic test case to `get`, `enable` and `disable` addons

### DIFF
--- a/apiclient/harvester_api/managers/addons.py
+++ b/apiclient/harvester_api/managers/addons.py
@@ -28,3 +28,6 @@ class AddonManager(BaseManager):
             data['spec']['enabled'] = enable
             code, data = self._update(path, data, as_json=True)
         return code, data
+
+    def disable(self, uid):
+        return self.enable(uid, False)
--- a/apiclient/harvester_api/managers/addons.py
+++ b/apiclient/harvester_api/managers/addons.py
@@ -1,0 +1,30 @@
+from collections.abc import Mapping
+
+from harvester_api.models.addons import BaseAddonSpec, MonitoringAddonSpec
+from .base import BaseManager, merge_dict
+
+
+class AddonManager(BaseManager):
+    PATH_fmt = "v1/harvester/harvesterhci.io.addons/{uid}"
+    Spec = BaseAddonSpec
+    MonitoringAddonSpec = MonitoringAddonSpec
+
+    def get(self, uid="", *, raw=False):
+        return self._get(self.PATH_fmt.format(uid=uid))
+
+    def update(self, uid, spec, *, raw=False, as_json=True, **kwargs):
+        path = self.PATH_fmt.format(uid=uid)
+        _, data = self.get(uid)
+        if isinstance(spec, self.Spec):
+            spec = spec.to_dict(data)
+        elif isinstance(spec, Mapping) and as_json:
+            spec = merge_dict(spec, data)
+        return self._update(path, spec, raw=raw, as_json=as_json, **kwargs)
+
+    def enable(self, uid, enable=True):
+        path = self.PATH_fmt.format(uid=uid)
+        code, data = self.get(uid)
+        if 200 == code:
+            data['spec']['enabled'] = enable
+            code, data = self._update(path, data, as_json=True)
+        return code, data
--- a/apiclient/harvester_api/models/addons.py
+++ b/apiclient/harvester_api/models/addons.py
@@ -1,0 +1,133 @@
+from json import loads, dumps
+
+from yaml import safe_load, safe_dump
+
+
+class _BaseConfig:
+    def __init__(self, data):
+        self._val = data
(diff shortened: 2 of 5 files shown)


apiclient/harvester_api/managers/addons.py
```diff
@@ -0,0 +1,30 @@
...
+    def update(self, uid, spec, *, raw=False, as_json=True, **kwargs):
+        path = self.PATH_fmt.format(uid=uid)
+        _, data = self.get(uid)
+        if isinstance(spec, self.Spec):
+            spec = spec.to_dict(data)
+        elif isinstance(spec, Mapping) and as_json:
+            spec = merge_dict(spec, data)
+        return self._update(path, spec, raw=raw, as_json=as_json, **kwargs)
+
+    def enable(self, uid, enable=True):
```
khushboo-rancher, line 24: Can we add `disable` method instead of handling with `enable=False`? It will be similar but more readable & convenient while using it. https://github.com/harvester/tests/pull/1350/files#diff-a398935b69540b5832e6addb8b4e43c3344826e390608865d887b1a98107ca4bR45

khushboo-rancher (changes requested):
LGTM in general, a minor comment

apiclient/harvester_api/managers/addons.py
lanfon72: To implement like this is considered to align the method in `HostManager`:
https://github.com/harvester/tests/blob/7fa14f5251356ce99faa3d993f718e1ea3589031/apiclient/harvester_api/managers/hosts.py#L29

of course we can still add a method `disable` as the shortcut to call `enable(..., False)`, then it will have only 1 argument, is it necessary? :thinking:

khushboo-rancher: I think, it will still be convenient for people to look for `disable` method when they are trying to use this.

apiclient/harvester_api/models/addons.py
```diff
@@ -0,0 +1,133 @@
...
+    _name = "vm-import-controller"
+    _id = "harvester-system/vm-import-controller"
+
+    def __init__(self, value=None, enabled=False):
+        super().__init__(value, enabled)
+        self.limits = value.get('resources', {}).get('limits', {})
+        self.requests = value.get('resources', {}).get('requests', {})
+
+    def to_dict(self, data):
+        return dict(spec=dict(enabled=self.enable, valuesConetne=dumps(self.value)))
```
albinsun, line 126: typo, should be `valuesContent`?
Should also check usage in callers.

apiclient/harvester_api/models/addons.py
```diff
@@ -0,0 +1,133 @@
...
+    @property
+    def enable(self):
+        return self._enable
+
+    @enable.setter
+    def enable(self, val):
+        self._enable = val
+
+    def to_dict(self, data):
+        return dict(spec=dict(enabled=self.enable, valuesConetne=safe_dump(self.value)))
```
albinsun, line 38: typo, should be `valuesContent`?
Should also check usage in callers.

apiclient/harvester_api/models/addons.py
```diff
@@ -0,0 +1,133 @@
...
+
+    @property
+    def enable(self):
+        return self._enable
+
+    @enable.setter
+    def enable(self, val):
+        self._enable = val
+
+    def to_dict(self, data):
```
albinsun, line 37: `data` argument is not been used?

apiclient/harvester_api/models/addons.py
```diff
@@ -0,0 +1,133 @@
...
+class VMImportControllerAddonSpec(BaseAddonSpec):
+    _name = "vm-import-controller"
+    _id = "harvester-system/vm-import-controller"
+
+    def __init__(self, value=None, enabled=False):
+        super().__init__(value, enabled)
+        self.limits = value.get('resources', {}).get('limits', {})
+        self.requests = value.get('resources', {}).get('requests', {})
+
+    def to_dict(self, data):
```
albinsun, line 125: `data` argument is not been used?

apiclient/harvester_api/models/addons.py
lanfon72: ref to https://github.com/harvester/tests/blob/7d41a56d02f276091e045696caf31511dd03dfbf/apiclient/harvester_api/models/settings.py#L129-L131, the argument is the placeholder for sub-classes in case they need the value from outer.

apiclient/harvester_api/managers/addons.py
albinsun: Should be able to resolve this as `disable()` already added.

albinsun (approved):
LGTM

khushboo-rancher (approved):

PR #1351, opened 2024-07-02
[e2e] [apiclient] fix load volume claim bug in VMSpec

lanfon72:
## Changes
- **UPDATE** VMSpec to check volume claim template is existed before access.


lanfon72:
We will hit the bug when loading an existing VM like:
image

khushboo-rancher (approved):

PR #1353, opened 2024-07-03
Bump RKE version to v1.28

albinsun:


lanfon72 (approved):
LGTM, but I think it is a example (default) value, we might not need update them to align testing environment (especially `password`)

albinsun:
> LGTM, but I think it is a example (default) value, we might not need update them to align testing environment (especially `password`)

Yep, we can just keep the example value.
But this provides user/qa a quick idea for the current defualt value in daily e2e env.
(And password is just for lazy :sweat_smile:)

PR #1356, opened 2024-07-03
fix terraform scripts

m-ildefons:
1) Silence output of wget to make logs more readable.
2) Force unzip to never prompt when extracting files.
   This can break CI pipelines because there the prompt will never be
   answered.


lanfon72 (approved):
LGTM

CI will never hit the prompt issue because we will download `terraform` to different folder as fixtures below:
https://github.com/harvester/tests/blob/4059bc64520faffdcec641e40f6d8716a790ae31/harvester_e2e_tests/fixtures/terraform.py#L191-L205

The prompt problem will happen only when executing the script independently.

PR #1366, opened 2024-07-08
removed tests that are invalid

noahgildersleeve:
This removes a couple of tests that are invalid since you can't decrease the size of a volume in longhorn. You can trim it but not actually decrease the volume size.


albinsun (approved):
LGTM as we can NOT requests decreasing as by current design storage can't be less than previous value.
image

PR #1367, opened 2024-07-09
feat: add terraform-provider-harvester to config.yaml

FrankYang0529:
We use `--terraform-provider-harvester` in `tf_provider_version` fixture, but this config is not in `config.yaml`.


lanfon72 (approved):
LGTM

PR #102, opened 2021-11-12
Include markers for Backup & Restore test cases

yarunachalam:
Include markers for test cases based on priority.
Include comment with test case number and description

PR #1384, opened 2024-07-15
[e2e][backend] Add test cases for restore backup with snapshot created

TachunLin:

Issue #1045 

According to issue https://github.com/harvester/harvester/issues/4954

We need to add the backend e2e test to the `vm_backup_restore` integration test   
To cover the case when a vm both have backup and snapshot created on it, when we restore this vm from backup.
It should be restore successfully.  

Added the following test script:
1. test_with_snapshot_restore_with_new_vm 
    - Restore vm also have snapshot created to a new vm
2. test_with_snapshot_restore_replace_retain_vols
    - Restore vm also have snapshot created to replace the existing vm and retain volume
  
Test result (Trigger locally and execute test on remote ecm lab machine)
1. Test can **PASS** all test cases in the `TestBackupRestoreWithSnapshot` class  
  image

2. Test can **PASS** most of the test cases in the `test_4_vm_backup_restore.py` file
  image

* In order to not affect all existing test cases in the `TestBackupRestore` and `TestBackupRestoreOnMigration`  and `TestMultipleBackupRestore` , consider the stability and future test scalability, thus I create a separate class for the new test scenario.


harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,210 @@ def test_restore_replace_with_vm_shutdown_command(
...
+    def test_connection(self, api_client, backup_config, config_backup_target):
+        code, data = api_client.settings.backup_target_test_connection()
+        assert 200 == code, f'Failed to test backup target connection: {data}'
+
+    @pytest.mark.dependency(depends=["TestBackupRestoreWithSnapshot::test_connection"], param=True)
+    def tests_backup_vm(self, api_client, wait_timeout, backup_config, base_vm_with_data):
+        unique_vm_name = base_vm_with_data['name']
+
+        # Create backup with the name as VM's name
+        code, data = api_client.vms.backup(unique_vm_name, unique_vm_name)
```
albinsun, line 649: We use `api_client.vms.backup(unique_vm_name, unique_vm_name)` many times in this suite.
It's ok if previous one teardown successfully but will broken if not.
May consider a unique backup name.

image

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,210 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        with vm_shell_from_host(
+            base_vm_with_data['host_ip'], base_vm_with_data['vm_ip'],
+            base_vm_with_data['ssh_user'], pkey=pri_key
+        ) as sh:
+            out, err = sh.exec_command(f"echo {pub_key!r} > {base_vm_with_data['data']['path']}")
+            assert not err, (out, err)
+            sh.exec_command('sync')
+
+        # Restore VM into new
+        restored_vm_name = f"{backup_config[0].lower()}-restore-{unique_vm_name}"
```
albinsun, line 699: Same, consider use an unique name to avoid previous teardown fail (previous `restored_vm` still there).
image

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,210 @@ def test_restore_replace_with_vm_shutdown_command(
...
+
+            out, err = sh.exec_command(f"cat {backup_data['path']}")
+
+        assert backup_data['content'] in out, (
+            f"cloud-init writefile failed\n"
+            f"Executed stdout: {out}\n"
+            f"Executed stderr: {err}"
+        )
+
+        # teardown: delete restored vm and volumes
```
albinsun, line 739: teardown snapshot

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,210 @@ def test_restore_replace_with_vm_shutdown_command(
...
+            raise AssertionError(
+                f"Failed to Delete VM({restored_vm_name}) with errors:\n"
+                f"Status({code}): {data}"
+            )
+        for vol in vm_spec.volumes:
+            vol_name = vol['volume']['persistentVolumeClaim']['claimName']
+            api_client.volumes.delete(vol_name)
+
+    @pytest.mark.dependency(depends=["TestBackupRestoreWithSnapshot::tests_backup_vm"], param=True)
+    def test_with_snapshot_restore_replace_retain_vols(
```
albinsun, line 759: Can we parameterize `delete_volumes` thus can test `Delete Previous Volumes` for both `Delete` and `Retain`?

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,210 @@ def test_restore_replace_with_vm_shutdown_command(
...
+                f"VM {unique_vm_name} Started {wait_timeout} seconds"
+                f", but cloud-init still in {out}"
+            )
+            out, err = sh.exec_command(f"cat {backup_data['path']}")
+
+        assert backup_data['content'] in out, (
+            f"cloud-init writefile failed\n"
+            f"Executed stdout: {out}\n"
+            f"Executed stderr: {err}"
+        )
```
albinsun, line 830: Please teardown

albinsun (changes requested):

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 
I set the restored_vm_name to use a more precise name to distinguish with other test class.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion.
A good design and structure of our backup and restore test is when we finalize the `vm_with_data` fixture. 
All related snapshot will be deleted accordingly. 
Thus we can leverage this manner to tear down the snapshots as well.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: It's a good idea to use parameterize config to cover both the `Delete` and `Retain` test cases.

While according to the test issue https://github.com/harvester/tests/issues/1045 

We just select the `Restore backup to replace existing (retain volume)` 

The reason is when the vm also have snapshot created, even if we shutdown the vm, the backend would check and prevent to restore to replace existing with delete volume.   
image

Thus here we test the replace existing with retain volume only.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the check and suggestion. 

Indeed after test case `test_with_snapshot_restore_replace_retain_vols` complete execution, the retain volume would exists and all backup, snapshot and vm would be cleaned up by fixture.  

image

I have tried different ways to clean up the retain volume like:
1. Add check clean up the retained volume at the end of the test case
2. Add a new fixture with the name cleanup_retain_volume to cleanup volume right after the base_vm_with_data fixture
3. Add a function inside the test case and use the request.addfinalizer() to call the function 

But the result is it would cause the vm stuck in `Terminating` and all snapshot are also not deleted
  image
  image

I also tried with manual operation from UI, when delete the retain volume then delete vm, it works fine. 
While when setting breakpoint and test with api call, the vm would also stuck in `Terminating`

Thus I add the code to check whether the retain volume exists only. 
 
To consider the tradeoff and not cause side effect to other test cases in the file, not sure whether we can keep the retain volume in this test case.

albinsun: Test case aiming for automate UI operations, they should have the same behavior.
However, teardown retained volume is not our main target and has low priority.
If still can not figure out why they behave differently, please comment a `# TODO` or `# FIXME` for come back it later.

TachunLin: Thanks for the suggestion. 
I added a `# TODO` to remind the retain volume to think of and come back it later after find the root cause and solution.

albinsun:
Have a quick try using ECM raven but fails on: 
1. `test_with_snapshot_restore_with_new_vm[NFS]`
2. `test_with_snapshot_restore_replace_retain_vols[NFS]`

Please help to check, thx.

`harvester-runtests/43`
image

bk201:
@TachunLin Please help check the error and make sure the test run successfully in Jenkins.

TachunLin:
Thanks for the reminder. I checked the test report of `harvester-runtests/43`. 
Most of the S3 related test was failed from `test_connection[S3]::setup` 

The reason is we not yet have the backup bucket created in our ecm lab minio artifact endpoint
I have created two more buckets `ravens` and `falcons` for future testing requirement on these machines. 

Then I set the same `config.yml` which used by the `harvester_run_test` pipeline from my local to trigger the same test to remote `ravens` cluster.  

The result is when I execute `TestBackupRestoreWithSnapshot` or the entire `test_4_vm_backup_restore.py` 
Both of them can pass most of the test cases.

* The `TestBackupRestoreWithSnapshot` class
image

* The `test_4_vm_backup_restore` file 
image

Next I trigger a new test `harvester-runtests/49`, there I get some expected failure like the following:

* `TestBackupRestore::test_restore_with_new_vm[S3] and [NFS]`, `TestBackupRestoreWithSnapshot::test_with_snapshot_restore_with_new_vm[S3] and [NFS]`

  ```
  E       AssertionError: Failed to Start VM(s3-restore-0735071682-09h31m16s956301-07-24) with errors:
  E         Status: 404
  E         API Status(404): {'type': 'error', 'links': {}, 'code': 'NotFound', 'message': 'virtualmachines.kubevirt.io "s3-restore-0735071682-09h31m16s956301-07-24" not found', 'status': 404}
  E       assert False
  ```
  image


* `test_restore_replace_with_delete_vols[S3] and [NFS]`, `TestBackupRestoreWithSnapshot::test_with_snapshot_restore_replace_retain_vols[S3] and [NFS]`

  ```
  E       AssertionError: cloud-init writefile failed
  E         Executed stdout: ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQC2J8e7zDDo/Mfgg4cmvt4OJYXOuY+LMfNnl6lQzdVhXJTNnnf2ulA+GMnqDsw2o5QCZ/bYkfXIvhnIHYh9PChucUujFMKhz2F3+q8fXQZqt+p6koAj7toMdmpd66rS8+x9Krmk7rS/0iZn13jqyjSIIsZ0/5fEM13jpVpWIUFC2w==
  E         
  E         Executed stderr: 
  E       assert '0708196929-09h31m16s956301-07-24' in 'ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAAgQC2J8e7zDDo/Mfgg4cmvt4OJYXOuY+LMfNnl6lQzdVhXJTNnnf2ulA+GMnqDsw2o5QCZ/bYkfXIvhnIHYh9PChucUujFMKhz2F3+q8fXQZqt+p6koAj7toMdmpd66rS8+x9Krmk7rS/0iZn13jqyjSIIsZ0/5fEM13jpVpWIUFC2w==\n'
  ```
  
  
  image


I would continue to investigate what cause these tests failed on the Jenkins run test jobs while works fine when trigger from local.

albinsun:
> ...
> 
> I would continue to investigate what cause these tests failed on the Jenkins run test jobs while works fine when trigger from local.

Hi @TachunLin ,
I checked our daily tests and found 2 basic restore test cases are failed too.
1. _TestBackupRestore::test_restore_with_new_vm[S3,NFS]_
2. _TestBackupRestore::test_restore_replace_with_delete_vols[S3,NFS]_

If they are PASS if trigger from local like you mentioned, then we should suspect it's an environment issue and investigate ECM lab.

CC @lanfon72 

`harvester-install-and-test-e2e-daily#328` (`v1.3-head`)
image

`harvester-install-and-test-e2e-daily#329` (`v1.2-head`)
image

lanfon72:
> > ...
> > I would continue to investigate what cause these tests failed on the Jenkins run test jobs while works fine when trigger from local.
> 
> Hi @TachunLin , I checked our daily tests and found 2 basic restore test cases are failed too.
> 
> 1. _TestBackupRestore::test_restore_with_new_vm[S3,NFS]_
> 2. _TestBackupRestore::test_restore_replace_with_delete_vols[S3,NFS]_
> 
> If they are PASS if trigger from local like you mentioned, then we should suspect it's an environment issue and investigate ECM lab.
> 
> CC @lanfon72
> 
> `harvester-install-and-test-e2e-daily#328` (`v1.3-head`) image
> 
> `harvester-install-and-test-e2e-daily#329` (`v1.2-head`) image

Those test cases are known issue in https://github.com/harvester/harvester/issues/4640
in KVM environment, they might not always reproduced.

albinsun:
Hi @TachunLin,
The flaky test case `test_restore_with_new_vm` is due to VM instance does not created immediately after trigging `api_client.backups.restore`, so `vm_checker.wait_ip_addresses` assert VM is not there and fail the test.

For detail, please refer to the quick fix PR
* harvester/tests/pull/1419

CC @khushboo-rancher @lanfon72

TachunLin:
Thank you @albinsun for finding the root cause and created PR https://github.com/harvester/tests/pull/1419 to fix the flaky case of `test_restore_with_new_vm`.  I am really appreciated for your help. 

I also added the `vm_checker.wait_getable` function into the `test_with_snapshot_restore_with_new_vm` and `test_with_snapshot_restore_replace_retain_vols`. 

And trigger the new test on the main Jenkins vm towards the `raven` cluster to run `TestBackupRestoreWithSnapshot` class only. 

It can well fix the `test_with_snapshot_restore_with_new_vm` test cases. 
image

And I also trigger the entire `test_4_vm_backup_restore`, it can also **PASS** most of the back and restore cases except those flaky cases failed before for other reason.

image

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,220 @@ def test_restore_replace_with_vm_shutdown_command(
...
+@pytest.mark.parametrize(
+    "backup_config", [
+        pytest.param("S3", marks=pytest.mark.S3),
+        pytest.param("NFS", marks=pytest.mark.NFS)
+    ],
+    indirect=True)
+class TestBackupRestoreWithSnapshot:
+
+    @pytest.mark.dependency()
+    def test_connection(self, api_client, backup_config, config_backup_target):
```
lanfon72, line 640: we already tested the connection in previous class, we don't need duplicate the test case.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -626,6 +626,220 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        code, data = api_client.settings.backup_target_test_connection()
+        assert 200 == code, f'Failed to test backup target connection: {data}'
+
+    @pytest.mark.dependency(depends=["TestBackupRestoreWithSnapshot::test_connection"], param=True)
+    def tests_backup_vm(self, api_client, wait_timeout, backup_config, base_vm_with_data):
```
lanfon72, line 645: same as above, it looks you are going to reuse the test case?

lanfon72 (changes requested):
please move those test cases into `TestBackupRestore` (in the last) to reduce test time, as both are depends on the same `test_backup_vm`, I feel that it is not necessary to create another class for it.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
lanfon72: Snapshot function is based on the volume, so delete volume will not work with snapshot should be the expected feature, or maybe we can discuss the case in sync up meeting to double confirm it.

TachunLin:
Thanks for the check and suggestion. 
In the begging to add restore with backup related test cases, I also have the same plan to add all test cases under the existing `TestBackupRestore` class.

But when I actually add the new `test_with_snapshot_restore_with_new_vm` and `test_with_snapshot_restore_replace_retain_vols` to the end of `TestBackupRestore` class.

After trigger the test, we can see while execution on the two new test cases. 
The existing virtual machine can't do any restore vm action to the new or replace the existing. 

* Execution on the `test_with_snapshot_restore_with_new_vm` 
  - We can find the vm did not respond to the restore action to the new vm
  
  https://github.com/user-attachments/assets/4d5e7760-66b0-432e-ba0e-96b76250a582


* Execution on the `test_with_snapshot_restore_with_new_vm` 
  - the vm also have no response for the action to replace the existing

  https://github.com/user-attachments/assets/839a32e9-5550-4dec-8059-1bed965e39e3


* On the test report, we can find the two new test cases failed 
  image

* Both of them got the `connection failed` error
  image


I think the reason to cause this failure may related to in the entire `TestBackupRestore` class, we use the same and the only one virtual machines. 
And since the virtual machine have already done with the restore to new vm and replace with existing vm in the previous tests `test_restore_with_new_vm` and `test_restore_replace_with_delete_vols`. 

Thus I plan to create a separate class `TestBackupRestoreWithSnapshot` to make the restore with snapshot related tests can run on a separate and clean vm without affecting all the existing test cases. 

Indeed it will increase the execution time to rerun the test connection and take backup test.

In a trade off, The pros to have a separate class can have the following benefits:
1. Make the new restore backup with snapshot cases can running well 
2. Maintain the stability of the backup and restore test without affecting with existing tests
3. For future scalability, in case of we need to add more restore backup with restore related test in the future  

According to this, may I continue to use the separate class `TestBackupRestoreWithSnapshot` for the new restore backup with snapshot test cases. Many thanks.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 
The reason to have the test connection running again is to maintain the full function of the backup and restore test cases inside the new test class `TestBackupRestoreWithSnapshot` 

According to the test result in the follow up comment
https://github.com/harvester/tests/pull/1384#issuecomment-2275375643

And consider the trade off of pros and cons, may I keep these tests in the new test class. Many thanks.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 
The reason to have the test connection running again is to provide a new backup for all the test cases inside the new test class `TestBackupRestoreWithSnapshot`.

According to the test result in the follow up comment
https://github.com/harvester/tests/pull/1384#issuecomment-2275375643

And consider the trade off of pros and cons, may I keep these tests in the new test class. Many thanks.

lanfon72: if you want to decouple new test cases without using existed one, you can just add another fixture to spawn up VM and do the backup (as it is the prerequisite of new test cases).

The problem here is, if you added those two duplicated (totally same as previous class) test cases, which means the total of test cases in report is not reflect to what we actually implemented and covered tests.

to add them after existed test cases would be more easy than create another fixture, that's why I suggested so.

albinsun:
@TachunLin , Any update on this PR?
Do we have any ongoing work item?

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
albinsun: @TachunLin, do we have any ongoing work items here?
the current point is we should not have 2 same tests
* `TestBackupRestore:tests_backup_vm`
* `TestBackupRestoreWithSnapshot:tests_backup_vm`

I think one way is to make `TestBackupRestoreWithSnapshot:tests_backup_vm` a class fixture.

albinsun: @TachunLin ,
Per discussed, let you know that harvester/tests/issues/1462 has been fixed so there should be no connection fail.
You can continue surveying the refactor of this task item, thx.

TachunLin: Thanks @albinsun for fixing the connection failed issue. 
According to the discussion and based on the [review comment](https://github.com/harvester/tests/pull/1384#discussion_r1709900479). 

I think it would be better moving the new snapshot restore test cases right after existed test cases in the same `TestBackupRestore`. It would solve the following concerns: 

1.  No reuse and duplication of `test_connection` and `tests_backup_vm`
2.  Save time without running the take backup action twice 

I already made the fix and tested working on the local to remote ECM lab machine
  image

Will try again on the main Jenkins jobs to double confirm

TachunLin: Tested `PASS` on the main Jenkins Harvester run test jobs [#102](http://172.19.99.74/job/harvester-runtests/102/)

All the new snapshot restore tests can execute well inside the `TestBackupRestore` test class. 
http://172.19.99.74/job/harvester-runtests/102/TestReport/

image

The S3 related tests are blocked to proceed due to the environment broken caused by vpn disconnection yesterday.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 
After moving the snapshot restore test cases into the original `TestBackupRestore`. 
When test successfully complete, all the test generated volume will be automatically cleanup.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
         spec = api_client.backups.RestoreSpec.for_existing(delete_volumes=True)
         code, data = api_client.backups.restore(unique_vm_name, spec)
         assert 201 == code, f'Failed to restore backup with current VM replaced, {data}'
-        vm_getable, (code, data) = vm_checker.wait_getable(unique_vm_name)
+
```
albinsun, line 608: Why remove `wait_getable`?
Won't possible to hit https://github.com/harvester/tests/pull/1419?

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        # take vm snapshot
+        code, data = api_client.vm_snapshots.create(unique_vm_name, vm_snapshot_name)
+        assert 201 == code
+
+        deadline = datetime.now() + timedelta(seconds=wait_timeout)
+        while deadline > datetime.now():
+            code, data = api_client.vm_snapshots.get(vm_snapshot_name)
+            if data.get("status", {}).get("readyToUse"):
+                break
+            print(f"waiting for {vm_snapshot_name} to be ready")
```
albinsun, line 656: Drop temporary code.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+            sleep(3)
+        else:
+            raise AssertionError(f"timed out waiting for {vm_snapshot_name} to be ready")
```
albinsun, line 659: Please have `code` and `data` in error message like other `assers`.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+
+        assert 200 == code
+        assert data.get("status", {}).get("readyToUse") is True
```
albinsun, line 662: Should these 2 lines be in the while loop?

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        vm_running, (code, data) = vm_checker.wait_status_running(unique_vm_name)
+        assert vm_running, (
+            f"Failed to restore VM({unique_vm_name}) with errors:\n"
+            f"Status({code}): {data}"
+        )
+
+        # Check VM Started then get IPs (vm and host)
+        vm_got_ips, (code, data) = vm_checker.wait_ip_addresses(unique_vm_name, ['default'])
+        assert vm_got_ips, (
+            f"Failed to Start VM({unique_vm_name}) with errors:\n"
```
albinsun, line 673: Do we have any restore or restart here?
If want to check VM still Running and accessible, then perhaps alter the error message.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        # take vm snapshot
+        code, data = api_client.vm_snapshots.create(unique_vm_name, vm_snapshot_name)
+        assert 201 == code
+
+        deadline = datetime.now() + timedelta(seconds=wait_timeout)
+        while deadline > datetime.now():
+            code, data = api_client.vm_snapshots.get(vm_snapshot_name)
+            if data.get("status", {}).get("readyToUse"):
+                break
+            print(f"waiting for {vm_snapshot_name} to be ready")
```
albinsun, line 775: same as previous.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+            sleep(3)
+        else:
+            raise AssertionError(f"timed out waiting for {vm_snapshot_name} to be ready")
+
+        assert 200 == code
+        assert data.get("status", {}).get("readyToUse") is True
```
albinsun, line 781: same as previous.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        assert vm_running, (
+            f"Failed to restore VM({unique_vm_name}) with errors:\n"
+            f"Status({code}): {data}"
+        )
+
+        # Check VM Started then get IPs (vm and host)
+        vm_got_ips, (code, data) = vm_checker.wait_ip_addresses(unique_vm_name, ['default'])
+        assert vm_got_ips, (
+            f"Failed to Start VM({unique_vm_name}) with errors:\n"
+            f"Status: {data.get('status')}\n"
```
albinsun, line 793: same as previous.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -605,9 +605,227 @@ def test_restore_replace_with_vm_shutdown_command(
...
+        base_vm_with_data['host_ip'], base_vm_with_data['vm_ip'] = host_ip, vm_ip
+
+        # mess up the existing data
+        with vm_shell_from_host(
+            base_vm_with_data['host_ip'], base_vm_with_data['vm_ip'],
+            base_vm_with_data['ssh_user'], pkey=pri_key
+        ) as sh:
+            out, err = sh.exec_command(f"echo {pub_key!r} > {base_vm_with_data['data']['path']}")
+            assert not err, (out, err)
+            sh.exec_command('sync')
```
albinsun, line 691: Does messing operation also needed in restore new?
As it's restore new, the messed data will not get back on base_vm.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the check. 
I think that would probably because during the rebase process I miss something for not adding this check back. 
Just added the `wait_getable` action back to the original places.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the check. 
The purpose to add the line to check vm_snapshot is ready was to ensure the snapshot are actually created before proceeding to the next step.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 
Add the `code` and `data` in error message like others in the code.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the check. 
The VM did not restart before, just make sure the vm is running after taking the snapshot. 
Just updated the assert error message accordingly.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 

Indeed these 2 lines can be put into the previous while loop since a redundant check exists on  
`data.get("status", {}).get("readyToUse")`. Update the code accordingly.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion.
Add the code and data in error message like others in the code.

harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
TachunLin: Thanks for the suggestion. 

Indeed these 2 lines can be put into the previous while loop since a redundant check exists on `data.get("status", {}).get("readyToUse")`. 
Update the code accordingly.

PR #1408, opened 2024-07-29
[e2e] add support bundle test cases

lanfon72:
## Changes
- **ADD** new test case `test_secret_file_exists` ref to #603
- **ADD** new test case `test_hardware_info_exists` ref to #569

### DIFF
--- a/harvester_e2e_tests/apis/test_support_bundle.py
+++ b/harvester_e2e_tests/apis/test_support_bundle.py
@@ -16,6 +16,7 @@
 # you may find current contact information at www.suse.com
 
 import re
+import tarfile
 from io import BytesIO
 from time import sleep
 from zipfile import ZipFile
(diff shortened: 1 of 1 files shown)


khushboo-rancher (commented):
LGTM.

Can we take this time and handle few more things in supportbundle tests as below
1. The test_longfile_exists doesn't check all the log files, can we include all the log files checks?
2. I think we can skip the test_delete for supportbundle. It would be good if we have supportbundle after tests run.

harvester_e2e_tests/apis/test_support_bundle.py
```diff
@@ -104,6 +106,79 @@ def test_logfile_exists(self, support_bundle_state):
...
+                fails.append(it)
+
+        assert not fails, (
+            f"Got {len(fails)} incorrect item(s) from secret file,"
+            " expected type should be 'rke.cattle.io/machine-plan'"
+            " and `data` field should not be empty but got:\n"
+            f"{fails}"
+        )
+
+        support_bundle_state.fio.seek(0)
```
albinsun, line 137: Is this necessary?
If so, perhaps better move to `L#121`?

lanfon72:
> LGTM.
> 
> Can we take this time and handle few more things in supportbundle tests as below
> 
> 1. The test_longfile_exists doesn't check all the log files, can we include all the log files checks?
> 2. I think we can skip the test_delete for supportbundle. It would be good if we have supportbundle after tests run.

1. could you make the list of **all** log files? the `test_longfile_exists` was implemented based on https://github.com/harvester/tests/issues/350, I don't have no idea how many log files is necessary and we need to verify them.
2. the `delete` method is not available for users, the test case is targeting to test the API's functionality, and we save the SB as a temporarily file, it will be removed after test finished.  and according to https://github.com/harvester/tests/issues/1037, support bundles will be clean up automatically, if we already have test report, do we still need the support bundle? (excepts upgrade testing)

harvester_e2e_tests/apis/test_support_bundle.py
lanfon72: It is necessary when following test cases going to read the `fio` again.  and yeah move to L121 would be good.

khushboo-rancher:
>I don't have no idea how many log files is necessary and we need to verify them.

If you can't add test here as you don't have idea, could you help create a test ticket to cover it.

>if we already have test report, do we still need the support bundle? 

Yes

lanfon72:
> > I don't have no idea how many log files is necessary and we need to verify them.
> 
> If you can't add test here as you don't have idea, could you help create a test ticket to cover it.
> 
> > if we already have test report, do we still need the support bundle?
> 
> Yes

so yeah, please create another ticket for your request, not ASK them in PRs.

lanfon72:
I don't want to fight in public space, but I FREQUENTLY FEEL uncomfortable when a PR is resolving some issues, but it be asked to stack up to fix other problems.

albinsun (dismissed):

jbrockmeyer (approved):

PR #1409, opened 2024-07-30
test: implement kubeconfig-default-token-ttl-minutes api setting tests

irishgordo:
Issue: https://github.com/harvester/tests/issues/1392
Resolves: test/implement-1392
See also: https://github.com/harvester/harvester/issues/6011

- base cases for `kubeconfig-default-token-ttl-minutes` setting in api (positive & negative)
- manager enhancement 
- model enhancement 

- N/A

- does baseline test of updating the setting from default value of 0 
- only applicable with Harvester versions greater than or equal to v1.3.1 (currently)

### DIFF
--- a/harvester_e2e_tests/apis/test_settings.py
+++ b/harvester_e2e_tests/apis/test_settings.py
@@ -162,3 +162,27 @@ class TestUpdateInvalidBackupTarget:
             f"S3 backup-target should check key/secret/bucket/region"
             f"API Status({code}): {data}"
         )
+
+
+@pytest.mark.p0
+@pytest.mark.settings
+class TestUpdateKubeconfigDefaultToken:
+    @pytest.mark.skip_version_before("v1.3.2", reason="Issue#5891 fixed after v1.3.2")
+    def test_invalid_kubeconfig_ttl_min(self, api_client):
+        KubeconfigTTLMinSpec = api_client.settings.KubeconfigDefaultTokenTTLSpec.TTL
+        spec = KubeconfigTTLMinSpec(99999999999999)
+        code, data = api_client.settings.update('kubeconfig-default-token-ttl-minutes', spec)
+        assert 422 == code, (
+            f"Kubeconfig Default Token TTL Minutes should not exceed 100yrs\n"
+            f"API Status({code}): {data}"
+        )
+
+    @pytest.mark.skip_version_before('v1.3.1')
+    def test_valid_kubeconfig_ttl_min(self, api_client):
+        KubeconfigTTLMinSpec = api_client.settings.KubeconfigDefaultTokenTTLSpec.TTL
+        spec = KubeconfigTTLMinSpec(172800)
+        code, data = api_client.settings.update('kubeconfig-default-token-ttl-minutes', spec)
+        assert 200 == code, (
+            f"Kubeconfig Default Token TTL Minutes be allowed to be set for 120 days\n"
+            f"API Status({code}): {data}"
+        )
--- a/apiclient/harvester_api/managers/settings.py
+++ b/apiclient/harvester_api/managers/settings.py
@@ -1,6 +1,7 @@
 from collections.abc import Mapping
 
-from harvester_api.models.settings import BaseSettingSpec, BackupTargetSpec, StorageNetworkSpec
+from harvester_api.models.settings import BaseSettingSpec, BackupTargetSpec, \
+    StorageNetworkSpec, KubeconfigDefaultTokenTTLSpec
 from .base import BaseManager, merge_dict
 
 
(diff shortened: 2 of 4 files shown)


irishgordo:
Prior to v1.4.0-Dev, `test_invalid_kubeconfig_ttl_min` will fail, as it was a bug in v1.3.1...

harvester_e2e_tests/apis/test_settings.py
```diff
@@ -162,3 +162,35 @@ def test_invalid_S3(self, api_client):
...
+@pytest.mark.p0
+@pytest.mark.settings
+@pytest.mark.settings
+@pytest.mark.skip_version_before('v1.3.1')
+class TestUpdateValidKubeconfigTTLMinuteSetting:
+    def test_valid_kubeconfig_ttl_min(self, api_client):
+        KubeconfigTTLMinSpec = api_client.settings.KubeconfigDefaultTokenTTLSpec.TTL
+        spec = KubeconfigTTLMinSpec(172800)
+        code, data = api_client.settings.update('kubeconfig-default-token-ttl-minutes', spec)
+        print("here")
```
albinsun, line 192: Please remove temporary code.

harvester_e2e_tests/apis/test_settings.py
```diff
@@ -162,3 +162,35 @@ def test_invalid_S3(self, api_client):
             f"S3 backup-target should check key/secret/bucket/region"
             f"API Status({code}): {data}"
         )
+
+
+@pytest.mark.p0
+@pytest.mark.negative
+@pytest.mark.settings
+@pytest.mark.settings
+@pytest.mark.skip_version_before('v1.3.1')
```
albinsun, line 171: This issue seems fixed in `v1.4.0`.
In current design, `skip_version_before(target_ver)` actually includes `target_ver`. (See _fixtures/api_client.py_)
So these will still be picked with `v1.3.1` cluster.
image


Suggest to use `skip_version_if` instead, it's more explicit, e.g.
`@pytest.mark.skip_version_if("> v1.2.0", "<= v1.4.0", reason="Issue#4293 fix after v1.4.0")`

harvester_e2e_tests/apis/test_settings.py
```diff
@@ -162,3 +162,35 @@ def test_invalid_S3(self, api_client):
             f"S3 backup-target should check key/secret/bucket/region"
             f"API Status({code}): {data}"
         )
+
+
+@pytest.mark.p0
+@pytest.mark.negative
+@pytest.mark.settings
+@pytest.mark.settings
```
albinsun, line 170: Duplicated mark.

harvester_e2e_tests/apis/test_settings.py
```diff
@@ -162,3 +162,35 @@ def test_invalid_S3(self, api_client):
...
+@pytest.mark.skip_version_before('v1.3.1')
+class TestUpdateInvalidKubeconfigTTLMinuteSetting:
```
albinsun, line 172: Any particular reason to have respective classes?
I think we can refer `TestUpdateInvalidStorageNetwork` to have just one 
`TestUpdateKubeconfigDefaultToken`.

albinsun (changes requested):

harvester_e2e_tests/apis/test_settings.py
irishgordo: Thanks for that mention :smile: :+1: - I've cleaned up that debugging code in the latest push

harvester_e2e_tests/apis/test_settings.py
irishgordo: That's a great mention :smile: :+1:  - I've implemented that decorator on the 1 negative test that will fail.
The other, positive test, I've just left the prior decorator, as that should always pass.

harvester_e2e_tests/apis/test_settings.py
irishgordo: That does make sense to condense them :smile: :+1:  - into one test class since there is not particularly a lot of negative or positive cases that require complex steps :smile: - I've gone ahead and implemented this in the most recent push.

harvester_e2e_tests/apis/test_settings.py
irishgordo: Great catch :smile: :+1: - I've cleaned up the extra decorator - thanks for that mention

harvester_e2e_tests/apis/test_settings.py
irishgordo: Actually, it's fixed in `v1.3-head` - so it will be fixed/working in v1.3.2 - so I just set the decorator to that - and updated that case :smile:

albinsun (approved):
LGTM

khushboo-rancher (approved):

Issue #1418, opened 2024-08-02
[TEST] Check flaky tests case TestBackupRestore::test_restore_with_new_vm 


albinsun:
## What's the test to develop? Please describe
Flaky daily tests case `TestBackupRestore::test_restore_with_new_vm` due to `virtualmachines.kubevirt.io XXX not found` seems a false alarm cuz seldom hit in local env.
![image](https://github.com/user-attachments/assets/28a275a4-30c1-45b4-a0be-ef64600224db)

## Prerequisite and dependency of test
1. Seldom hit in local env.
2. Hit often in daily test env.

## Describe the items of the test development (DoD, definition of done) you'd like
n/a

## Additional context
n/a

PR #1419, opened 2024-08-02
Wait VM becomes getable after triggering restore

albinsun:
### Which issue(s) this PR fixes:
* Issue #1418 

### What this PR does / why we need it:
1. ADD `vm_checker.wait_getable` between `api_client.backups.restore` and `vm_checker.wait_ip_addresses`
   * `vm_checker.wait_ip_addresses` calls `api_client.vms.start` in the callback chain and will assert FAIL if target VM does not exist.
   * `api_client.backups.restore` needs some time to communicate with `backup-target` to restore VM. So depends on the env., VM instance may not be created immediately.

### Special notes for your reviewer:
This is a quick fix.
IMO, coder tends to expect function `wait_XXX` does some polling GET, thus some error-prone. To enhance, we may:
1. Only polling GET and call POST, PUT, DELETE explicitly.
1. Refactor function name to something like `create_AAA_and_wait_XXX`.
1. ...

#### Verification
* Before (`harvester-runtests#61`)
  image

* After (`harvester-runtests#62`)
  image

### DIFF
--- a/harvester_e2e_tests/fixtures/virtualmachines.py
+++ b/harvester_e2e_tests/fixtures/virtualmachines.py
@@ -170,6 +170,17 @@ def vm_checker(api_client, wait_timeout, sleep_timeout, vm_shell):
             finally:
                 self.snooze, self.wait_timeout = s, t
 
+        def wait_getable(self, vm_name, endtime=None, callback=default_cb, **kws):
+            endtime = endtime or self._endtime()
+            while endtime > datetime.now():
+                ctx = ResponseContext('vm.get', *self.vms.get(vm_name, **kws))
+                if 200 == ctx.code and callback(ctx):
+                    break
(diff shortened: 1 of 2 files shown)


TachunLin (approved):
LGTM. 

Thank you @albinsun for checking and bringing the solution to fix the `test_restore with_new_vm` related test cases. 

I added the `vm_checker.wait_getable` into all related test case in the `test_4_vm_backup_restore.py` and also test along with the new class `TestBackupRestoreWithSnapshot` for PR https://github.com/harvester/tests/pull/1384

And trigger a new test toward the `raven` cluster. 

The result is all `test_restore_with_new_vm` test cases are running well and also **PASS** most of the backup and restore test cases except those flaky cases already failed before for other reason.
image

PR #1437, opened 2024-08-08
[apiclient] fix KeyError when removing empty key

lanfon72:
## Changes
add **default** value when removing the key to prevent `KeyError` raised.


TachunLin (approved):
LGTM

albinsun (approved):

PR #1485, opened 2024-08-28
Implement cluster size check and loggin addon test case on test_upgrade.py

albinsun:
#### Which issue(s) this PR fixes & What it does:
1. harvester/harvester/issues/6425
   * Skip `test_upgrade.py::TestInvalidUpgrade::test_degraded_volume` when cluster size < 3.
   * If not skip, 1 and 2 nodes cluster will just start upgrade in this negtive case Class.
2. #901
   * Implement `test_preq_setup_logging` TODO
   * Post-upgrade check `rancher-logging` addon status in `test_verify_logging_pods` 

n/a

Verification
image

### DIFF
--- a/harvester_e2e_tests/integrations/test_upgrade.py
+++ b/harvester_e2e_tests/integrations/test_upgrade.py
@@ -501,7 +501,9 @@ class TestInvalidUpgrade:
         """
         # https://github.com/harvester/harvester/issues/6425
         if cluster_state.size < 3:
-            pytest.skip(f"Degraded only checked when nodes >= 3, skip for {cluster_state.size}.")
+            pytest.skip(
+                f"Degraded volumes only checked on 3+ nodes cluster, skip on {cluster_state.size}."
+            )
 
         vm_name, ssh_user, pri_key = stopped_vm
         vm_started, (code, vmi) = vm_checker.wait_started(vm_name)
--- a/harvester_e2e_tests/integrations/test_upgrade.py
+++ b/harvester_e2e_tests/integrations/test_upgrade.py
@@ -30,18 +30,12 @@ def cluster_state(request, unique_name, api_client):
 
     state = ClusterState()
 
-    # target version
     if request.config.getoption('--upgrade-target-version'):
         state.version_verify = True
         state.version = request.config.getoption('--upgrade-target-version')
     else:
         state.version_verify = False
         state.version = f"version-{unique_name}"
-
-    # cluster size
-    code, data = api_client.hosts.get()
-    assert 200 == code, (code, data)
-    state.size = len(data['data'])
 
     return state
 
@@ -500,9 +494,11 @@ class TestInvalidUpgrade:
         4. Upgrade should fail.
         """
         # https://github.com/harvester/harvester/issues/6425
-        if cluster_state.size < 3:
+        code, data = api_client.hosts.get()
+        assert 200 == code, (code, data)
+        if (cluster_size := len(data['data'])) < 3:
             pytest.skip(
-                f"Degraded volumes only checked on 3+ nodes cluster, skip on {cluster_state.size}."
+                f"Degraded volumes only checked on 3+ nodes cluster, skip on {cluster_size}."
             )
 
         vm_name, ssh_user, pri_key = stopped_vm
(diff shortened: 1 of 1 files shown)


lanfon72:
We should check whether it is a feature or not.
If it's a bug, the failure of test case will be expected, which means we would not need to skip it. (or skip it when the bug is not fixed in previous version.)

albinsun:
> We should check whether it is a feature or not. If it's a bug, the failure of test case will be expected, which means we would not need to skip it. (or skip it when the bug is not fixed in previous version.)

Currently it's by design.
Ref. https://github.com/harvester/harvester/issues/6425#issuecomment-2313858588

lanfon72:
> > We should check whether it is a feature or not. If it's a bug, the failure of test case will be expected, which means we would not need to skip it. (or skip it when the bug is not fixed in previous version.)
> 
> Yes, currently it's by design (feature). Ref. [harvester/harvester#6425 (comment)](https://github.com/harvester/harvester/issues/6425#issuecomment-2313858588)

Let's add it as reference with comment in the code.
`cluster_state` is used to store the information before and after upgrade, so we didn't use it in this test case, we shouldn't too.

I think I would need to add some docstring to explain the `ClusterState` :thinking:

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -494,6 +499,10 @@ def test_degraded_volume(
         3. Immediately upgrade Harvester.
         4. Upgrade should fail.
         """
+        # https://github.com/harvester/harvester/issues/6425
+        if cluster_state.size < 3:
+            pytest.skip(f"Degraded only checked when nodes >= 3, skip for {cluster_state.size}.")
```
khushboo-rancher, line 504: Typo: Degraded volumes only checked when nodes >= 3 :)

khushboo-rancher (approved):

harvester_e2e_tests/integrations/test_upgrade.py
albinsun: Fixed in 702026f.
image

albinsun:
Guys, as we on the way of `test_upgrade.py` enhancement, so I also implement a TODO and #901 together. (Not a good practice I know, will avoid next time...)
Please help to review the new committed logging addon part.

Sorry for inconvenience, no more task item will be add to this PR. :P

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -546,10 +557,31 @@ def test_degraded_volume(
 @pytest.mark.upgrade
 @pytest.mark.any_nodes
 class TestAnyNodesUpgrade:
+    @pytest.mark.skip_version_if("< v1.2.0", reason="logging is implemented as addon since v1.2.0")
```
lanfon72, line 560: We can't do `skip` in the setup case, otherwise we will lose the test case in old version.
(as `test_verify_logging_pods` is depends on this, so all `setup` test case should never be skipped.)

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -735,6 +767,8 @@ def test_verify_logging_pods(self, api_client):
         """ Verify logging pods and logs
         Criteria: https://github.com/harvester/tests/issues/535
         """
+        code, data = api_client.addons.get("cattle-logging-system/rancher-logging")
```
lanfon72, line 770: You would need to check whether the cluster version is the versions with logging addon first.

lanfon72 (changes requested):
I need to explain how the `TestUpgrade` works, then you can know why the implementation need to change.

As the manual test, steps would like:
1. Set up environment before upgrade
2. Perform upgrade
3. Verify the environment is expected

 What we do in upgrade automation is,
1. test case starts with `test_pre_setup` is going to be executed before upgrade
2. test case `test_perform_upgrade` will start the upgrade
3. test case starts with `test_verify` is going to verify the test case (after upgraded) hit criteria

cluster state changes during upgrade will be recorded in `cluster_state`. (that's why we will not use it in `TestInvalidUpgrade`)

Let's look into the case of `setup_logging`, we move logging to addon after `v1.2.0`, so we need to consider following cases:
1. A upgrade to `v1.2.0` (implemented)
2. `>v1.2.0` upgrade to B

When you implement case `2`, you would still need to consider the case `1` in both `pre_setup` and `verify`.

albinsun:
Hi @lanfon72 ,
Have tried to refactor according comments in 7516c7e, please help to review, thanks. 
* Single node
  image
* Three nodes
  image

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -482,9 +482,8 @@ def test_version_compatibility(
         api_client.upgrades.delete(data['metadata']['name'])
         api_client.versions.delete(version)
 
-    def test_degraded_volume(
-        self, api_client, wait_timeout, vm_shell_from_host, vm_checker, upgrade_target, stopped_vm
-    ):
+    def test_degraded_volume(self, api_client, cluster_state, wait_timeout, vm_shell_from_host,
```
lanfon72, line 485: it looks we forgot to remove `cluster_state` in parameters.

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -547,9 +554,44 @@ def test_degraded_volume(
...
-        # TODO: enable addon if > v1.2.0
-        return
+    def test_preq_logging_pods(self, api_client, wait_timeout, sleep_timeout):
+        # logging is an addon instead of built-in since v1.2.0
+        if api_client.cluster_version.release >= (1, 2, 0):
+            addon = "cattle-logging-system/rancher-logging"
+
+            code, data = api_client.addons.get(addon)
+            assert 200 == code, (code, data)
+            assert not data.get('spec', {}).get('enabled', True), (code, data)
```
lanfon72, line 564: IMO, we can just skip the set up when the `addon` had been enabled, it would let us to reuse existing cluster to do the upgrade testing.

another reason is, we are not testing the setting, we setting up the environment to hit the prerequisite.

harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -750,6 +796,10 @@ def test_verify_logging_pods(self, api_client):
                 "\n".join(f"Pod({n})'s phase({p}) is not expected." for n, p in fails)
             )
 
+        # teardown
+        if api_client.cluster_version.release >= (1, 2, 0):
```
lanfon72, line 800: to notice that if we didn't enable addon before upgrade (consider that the addon already be enabled before testing), we should not disable it.

harvester_e2e_tests/integrations/test_upgrade.py
albinsun: Removed in 6a19ea8

harvester_e2e_tests/integrations/test_upgrade.py
albinsun: Enhanced in 6a19ea8 with `logging_addon` fixture.

harvester_e2e_tests/integrations/test_upgrade.py
albinsun: Enhanced in 6a19ea8 by logging_addon fixture.

albinsun:
Verification for 6a19ea8
* Logging addon enabled already
  logging_enabled_already
* Logging addon enabled yet
  logging_enabled_yet

lanfon72 (approved):
LGTM

khushboo-rancher (approved):

PR #1495, opened 2024-09-02
Fix incorrect if-condition in test_create_with_invalid_url

brandboat:
before https://github.com/harvester/harvester/commit/13bb968e86cefb43227a18544b4ed1a40a1090b8, if a vm image failed to recognize the image url, we didn't have any condition in status until retry 3 times, while after the commit, we will have RetryLimitExceeded condition in `status` upon retry failure in the first round retry. Now the test didn't wait until the right condition and do assertion thus cause test failed.

This pr mainly move the assertion part into the while loop, make sure the assertion is correct until timeout.

N/A

N/A

### DIFF
--- a/harvester_e2e_tests/apis/test_images.py
+++ b/harvester_e2e_tests/apis/test_images.py
@@ -236,7 +236,7 @@ def test_create_with_invalid_url(api_client, gen_unique_name, wait_timeout):
     while endtime > datetime.now():
         code, data = api_client.images.get(unique_name)
         image_conds = data.get('status', {}).get('conditions', [])
-        if len(image_conds) > 0:
+        if "Initialized" == image_conds[-1].get("type"):
             break
         sleep(3)
 


lanfon72:
Hi @brandboat, It looks the change is not related to #1173 which related to `checksum`.

and as you mentioned, the new condition `RetryLimitExceeded` will be added to fix the bug in #1173, I want to know whether the `initialized` would still be the last one of conditions or not?

brandboat:
> Hi @brandboat, It looks the change is not related to https://github.com/harvester/tests/issues/1173 which related to checksum.

You are right, should I create another issue ?

> and as you mentioned, the new condition RetryLimitExceeded will be added to fix the bug in https://github.com/harvester/tests/issues/1173, I want to know whether the initialized would still be the last one of conditions or not?

It depends, but for the invalid_url test case, the Initialized condition is still the last one in the array. For me it's a bit unusual to guarantee the condition's position, so I adjusted the test to locate the Initialized condition rather than assuming it’s always the last element.

lanfon72:
> > Hi @brandboat, It looks the change is not related to #1173 which related to checksum.
> 
> You are right, should I create another issue ?
> 

No, but please don't refer this to #1173.

> > and as you mentioned, the new condition RetryLimitExceeded will be added to fix the bug in #1173, I want to know whether the initialized would still be the last one of conditions or not?
> 
> It depends, but for the invalid_url test case, the Initialized condition is still the last one in the array. For me it's a bit unusual to guarantee the condition's position, so I adjusted the test to locate the Initialized condition rather than assuming it’s always the last element.

if so, we can simply change the:
```
if len(image_conds) > 0:
```
into
```
if "Initialized" == image_conds[-1].get("type"):
```

to fix the bug for new change.

As the test case scenario is focusing on `invalid_url`, we would not check the while loop is timed out or not.
(actually we will never hit timed out in the case)

lanfon72 (approved):
LGTM

bk201 (approved):

Issue #1465, opened 2024-08-22
[TEST] Fix test_z_terraform.py::test_create_ssh_key::setup


albinsun:
## What's the test to develop? Please describe

* `test_z_terraform.py::test_create_ssh_key::setup` FAIL on `terraform init`
  ![image](https://github.com/user-attachments/assets/cad8b136-a00b-4323-b954-d6907587e1d5)
* Seems caused by clean terraform cache not work.
  The `rv` object is async and not return yet.
  ![image](https://github.com/user-attachments/assets/6c02594a-d2d5-4580-b829-2cdd2cb8fda8)

lanfon72:
The bug introduced from #1421, the original idea that we should use latest version for `0.0.0-dev` to get correct sub class (`TerraformResource_063`), but still need to write `0.0.0-dev` in **provider.tf**.

The fix #1421 did let us to get correct sub class, but the **provider.tf** still wrote latest version `0.6.4` as its version; then when we try to install `0.6.4` as the provider, terraform will try to search whether the global path `~/.terraform/plugins/...` contains it, and as we were put `0.0.0-dev` in the folder, so the installation will fail. (as the snapshot you provided.)

To fix this, we have several things need to do:
1. Make sure we can get latest sub class and use `0.0.0-dev` as the version if user provided it.
2. Remove the global installed `0.0.0-dev` (I had removed it)
3. When user use `0.0.0-dev`, DO NOT install it to global path, we can use `terraform init -plugin-dir` as the alternative path

PR #1507, opened 2024-09-05
[e2e] fix terraform fixture bug

lanfon72:
## Changes
- **UPDATE** terraform fixture to fix bug when using `0.0.0-dev`

### DIFF
--- a/harvester_e2e_tests/fixtures/terraform.py
+++ b/harvester_e2e_tests/fixtures/terraform.py
@@ -170,12 +170,12 @@ def tf_script_dir(request):
 @pytest.fixture(scope="session")
 def tf_provider_version(request, harvester_metadata):
     version = request.config.getoption('--terraform-provider-harvester')
-    import requests
-    resp = requests.get("https://registry.terraform.io/v1/providers/harvester/harvester")
-    latest = max(resp.json()['versions'], key=parse_version)
-    version = None if version == '0.0.0-dev' else version
-    harvester_metadata['Terraform Harvester Provider Version'] = f"{version} || {latest}"
-    return version or latest
+    if not version:
+        import requests
+        resp = requests.get("https://registry.terraform.io/v1/providers/harvester/harvester")
+        version = max(resp.json()['versions'], key=parse_version)
+    harvester_metadata['Terraform Harvester Provider Version'] = f"{version}"
+    return version
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


lanfon72:
Tested on `harvester-runtests#101`
image

Those fail cases caused by the ISO server which is unavailable now.

harvester_e2e_tests/fixtures/terraform.py
```diff
@@ -223,7 +223,9 @@ def tf_rancher(rancher_api_client, tf_script_dir, tf_provider_rancher_ver, tf_ex
 @pytest.fixture(scope="session")
 def tf_resource(tf_provider_version):
     converter = Path("./terraform_test_artifacts/json2hcl")
-    return TerraformResource.for_version(tf_provider_version)(converter)
+    # update `0.0.0-dev` to get newest class
+    version = "8.8.99" if tf_provider_version == '0.0.0-dev' else tf_provider_version
```
irishgordo, line 227: qu: I'm sort of confused with the setting of `8.8.99`? is that to allign with something in:
https://github.com/harvester/tests/blob/586d6a718573a718f53ee3e7674fb0c647bf373f/apiclient/harvester_api/api.py#L38-L55

I was expecting this to read maybe something like:
```python
version = None if tf_provider_version == '0.0.0-dev' else tf_provider_version
```

:sweat_smile:  - but maybe I'm missing something...

Don't want this to be blocking, because other than this qu, this looks good :smille:

irishgordo (changes requested):
Just had a small qu :smile: - non-blocking

harvester_e2e_tests/fixtures/terraform.py
lanfon72: yeah it is align to the version in `api.py`, you can check the explanation in https://github.com/harvester/tests/issues/1465#issuecomment-2310612259, the version in `tf_resource` will be used to get the correct implementation of `TerraformResource`, it should be a valid version (type: str), we will use it to determine which `TerraformResource` class we can use.

Classes of `TerraformResource` is implemented to align with the Provider version, for example, *Terraform Harvester Provider* implemented `cloudinit_secret` in `v0.6.3`[^1], so we implement `TerraformResource_063` to suppor the new implemented feature and mark it will `support_to = "0.6.3"`.

[^1]: https://github.com/harvester/terraform-provider-harvester/releases/tag/v0.6.3

albinsun (approved):
LGTM

irishgordo (approved):

irishgordo:
LGTM

PR #1529, opened 2024-09-17
Test Information

m-ildefons:
Annotate the test information with Rancher/RKE version information. This helps the test report HTML files to be less ambiguous about the configuration under test.

### DIFF
--- a/apiclient/rancher_api/api.py
+++ b/apiclient/rancher_api/api.py
@@ -58,6 +58,8 @@ class RancherAPI:
             ver = data['value']
             # XXX: fix master-xxx-head to 8.8.8, need the API fix the problem
             self._version = parse_version('8.8.8' if 'master' in ver else ver)
+
+            self._version.raw = ver
         return self._version
 
     def __repr__(self):
--- a/harvester_e2e_tests/fixtures/rancher_api_client.py
+++ b/harvester_e2e_tests/fixtures/rancher_api_client.py
@@ -7,7 +7,7 @@ from rancher_api import RancherAPI
 
 
 @pytest.fixture(scope="session")
-def rancher_api_client(request):
+def rancher_api_client(request, harvester_metadata):
     endpoint = request.config.getoption("--rancher-endpoint")
     password = request.config.getoption("--rancher-admin-password")
     ssl_verify = request.config.getoption("--ssl_verify", False)
@@ -16,6 +16,9 @@ def rancher_api_client(request):
     api.authenticate("admin", password, verify=ssl_verify)
 
     api.session.verify = ssl_verify
+
+    harvester_metadata['Rancher Endpoint'] = endpoint
+    harvester_metadata['Rancher Version'] = api.cluster_version.raw
 
     return api
 
@@ -42,18 +45,20 @@ def _pickup_k8s_version(versions, target_version):
 
 
 @pytest.fixture(scope='session')
-def rke1_version(request, rancher_api_client):
+def rke1_version(request, rancher_api_client, harvester_metadata):
     target_ver = request.config.getoption("--k8s-version")
 
     code, data = rancher_api_client.settings.get("k8s-versions-current")
     assert 200 == code, (code, data)
     supported_vers = data["value"].split(",")
 
-    return _pickup_k8s_version(supported_vers, target_ver)
+    ver = _pickup_k8s_version(supported_vers, target_ver)
+    harvester_metadata['RKE1 Version'] = ver
+    return ver
 
 
 @pytest.fixture(scope="session")
-def rke2_version(request, api_client, rancher_api_client):
+def rke2_version(request, rancher_api_client, harvester_metadata):
     target_ver = request.config.getoption("--k8s-version")
 
     # Ref. https://github.com/rancher/dashboard/blob/master/shell/edit/provisioning.cattle.io.cluster/rke2.vue  # noqa
@@ -61,15 +66,19 @@ def rke2_version(request, api_client, rancher_api_client):
     assert resp.ok
     supported_vers = [r['id'] for r in resp.json()['data']]
 
-    return _pickup_k8s_version(supported_vers, target_ver)
+    ver = _pickup_k8s_version(supported_vers, target_ver)
+    harvester_metadata['RKE2 Version'] = ver
+    return ver
 
 
 @pytest.fixture(scope="session")
-def k3s_version(request, api_client, rancher_api_client):
+def k3s_version(request, rancher_api_client, harvester_metadata):
     target_ver = request.config.getoption("--k8s-version")
 
     resp = rancher_api_client._get("v1-k3s-release/releases")
     assert resp.ok
     supported_vers = [r['id'] for r in resp.json()['data']]
 
-    return _pickup_k8s_version(supported_vers, target_ver)
+    ver = _pickup_k8s_version(supported_vers, target_ver)
+    harvester_metadata['K3S Version'] = ver
+    return ver


apiclient/rancher_api/api.py
```diff
@@ -56,8 +56,16 @@ def cluster_version(self):
...
             code, data = self.settings.get('server-version')
             ver = data['value']
-            # XXX: fix master-xxx-head to 8.8.8, need the API fix the problem
-            self._version = parse_version('8.8.8' if 'master' in ver else ver)
+
+            try:
+                # XXX: fix master-xxx-head to 8.8.8, need the API fix the problem
+                cur, _, _ = ver.split('-')
+                cur = "v8.8.8" if "master" in cur else cur
+                self._version = parse_version(f"{cur}.99")
```
albinsun, line 64: Is this for branch head?
e.g. `v1.4-ba586eaf-head` -> `v1.4.99`
```
>>> ver = "v1.4.0"
>>> parse_version('8.8.8' if 'master' in ver else ver)
<Version('1.4.0')>    # OK
>>> 
>>> ver = "v1.4.0-rc1"
>>> parse_version('8.8.8' if 'master' in ver else ver)
<Version('1.4.0rc1')>    # OK
>>> 
>>> ver = "v1.4.0-dev-20240101"
>>> parse_version('8.8.8' if 'master' in ver else ver)
<Version('1.4.0.dev20240101')>    # OK
>>> 
>>> ver = "master"
>>> parse_version('8.8.8' if 'master' in ver else ver)
<Version('8.8.8')>    # OK
>>> 
>>> ver = "v1.4-ba586eaf-head"
>>> parse_version('8.8.8' if 'master' in ver else ver)
Traceback (most recent call last):
  ...
    raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources._vendor.packaging.version.InvalidVersion: Invalid version: 'v1.4-ba586eaf-head'
>>> 
>>> cur, _, _ = ver.split('-')
>>> cur = "v8.8.8" if "master" in cur else cur
>>> parse_version(f"{cur}.99")
<Version('1.4.99')>
```

apiclient/rancher_api/api.py
```diff
@@ -56,8 +56,16 @@ def cluster_version(self):
...
+            except ValueError:
+                self._version = parse_version(ver)
```
albinsun, line 66: It's better to keep current implementation in try and new one in exception.
```
try:
    self._version = parse_version('8.8.8' if 'master' in ver else ver)
except:
    #YOUR_NEW_IMPLEMENTATION
```

albinsun (commented):
Rancher part LGTM. Have some comments on Harvester part.

apiclient/rancher_api/api.py
m-ildefons: I copied this code from the harvester API part assuming that Rancher follows a similar release naming. I updated it to match their release naming as best I could.

apiclient/rancher_api/api.py
m-ildefons: No, why?

m-ildefons:
The goal of this PR is to enhance the metadata annotations of the test results and go from this:

Screenshot at 2024-09-17 10-04-48

to this:
Screenshot at 2024-09-17 10-05-02

apiclient/rancher_api/api.py
```diff
@@ -56,8 +56,14 @@ def cluster_version(self):
         if not self._version:
             code, data = self.settings.get('server-version')
             ver = data['value']
-            # XXX: fix master-xxx-head to 8.8.8, need the API fix the problem
-            self._version = parse_version('8.8.8' if 'master' in ver else ver)
+
+            # Rancher versions look like v2.9.2, or v2.9.2-something
```
lanfon72, line 60: Although Rancher is not using `master-head`, we would still need to consider cases like:
1. `v2.9.1-rc6`
2. `v2.9-2f9a7509687f559d772be75d61df7f7409803883-head`

As the second case would be the latest of `v2.9`, we need to fill the patch version for it. (to notice that there will not have something like `v2.9.1-sha-head`)

so, I would suggest that we could update it to something like
https://github.com/harvester/tests/blob/4d4d22b397e4fa8ff73ce193c6fce6b78624d356/apiclient/harvester_api/api.py#L43-L55

which we already implemented in `HarvesterAPI`.
To notice that Python usually use try-except to control the flow called EAFP.[^1]

[^1]: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

lanfon72 (changes requested):

apiclient/rancher_api/api.py
m-ildefons: That was what I did initially, but @albinsun wasn't happy with it.

albinsun: I was talking about the order but not drop error handling.
My thought was
1. Existing implementation works for most cases
   (Except head form like `v2.9-2f9a7509687f559d-head`)
2. I thought you are dealing the head form case, so it's natually to suggest
    ```
    try:
       # Existing implementation
    except:
       # Dealing exceptions
    ```

Anyway, per discussed, It's reasonable for me to align with Harvester API side implementation.

m-ildefons: I think I have a good implementation now. It tries hard to parse the version, it keeps the suffix information for later reference and if all fails, it behaves like the Harvester API implementation.
I noticed that with my version of the `pkg_resources` library `parse_version` never raises an exception, but will produce a `LegacyVersion` object in case it doesn't know how to deal with it:
```
Python 3.6.15 (default, Sep 23 2021, 15:41:43) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pkg_resources import parse_version
>>> parse_version("foobar-noversion")
<LegacyVersion('foobar-noversion')>
```
Therefore I'm quite sure that the exception case will not raise another error.

albinsun: As I know our daily env. adopts python 3.10 which has different behavior on `parse_version`
```
Python 3.10.14 (main, Mar 24 2024, 08:01:04) [GCC] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pkg_resources import parse_version
<stdin>:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
>>> parse_version("foobar-noversion")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/albin/Workspace/albinsun/tests/.tox/py310/lib/python3.10/site-packages/pkg_resources/_vendor/packaging/version.py", line 200, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
pkg_resources._vendor.packaging.version.InvalidVersion: Invalid version: 'foobar-noversion'
```

lanfon72: `LegacyVersion` had been dropped in https://github.com/pypa/packaging/pull/407

apiclient/rancher_api/api.py
```diff
@@ -55,9 +55,27 @@ def __init__(self, endpoint, token=None, session=None):
...
+            version = data['value']
+
+            # Rancher versions look like v2.9.2, or v2.9.2-suffix but also
+            # possibly v2.9-abcdef0123-head
+            try:
+                # Split off the version name
+                # parse the version name into a Version object if possible
+                # save the suffixes as one string, e.g.:
+                # v2.9.2               => <Version('2.9.2')>, suffix = None
+                # v2.9.2-rc1           => <Version('2.9.2')>, suffix = "rc1"
```
lanfon72, line 67: To notice that the suffix is meaningful for the Version instance to compare with others, for example:
image

so we should only handle the case like `v2.9-2f9a7509687f559d772be75d61df7f7409803883-head`, as I know, rancher will not use `-dev-<date>` format, so it is fine to not consider the case.

albinsun (approved):
LGTM

m-ildefons:
I've reduced this PR to the bare essential, which is to have the raw version string in the test metadata.

lanfon72 (approved):

PR #1539, opened 2024-09-19
Remove legacy image in test_delete_image_recreate

brandboat:
n/a

During testing for https://github.com/harvester/harvester/issues/4293, I noticed that some images were not being cleaned up after the test finished. Automating the image cleanup would greatly reduce effort, especially when running the same method multiple times. Looping the test 20–30 times can fill up the node's disk quickly as images consume significant space.

I didn't run all e2e test but only `test_delete_image_recreate`, if this approach breaks anything or is not appropriate, feel free to close this pr, thanks.

n/a

### DIFF
--- a/harvester_e2e_tests/integrations/test_1_images.py
+++ b/harvester_e2e_tests/integrations/test_1_images.py
@@ -318,6 +318,9 @@ class TestBackendImages:
         get_image(api_client, unique_name)
         delete_image(api_client, unique_name, wait_timeout)
 
+        get_image(api_client, image_name)
+        delete_image(api_client, image_name, wait_timeout)
+
     @pytest.mark.p0
     def test_create_invalid_file(
         self, api_client, gen_unique_name, fake_invalid_image_file, wait_timeout


harvester_e2e_tests/integrations/test_1_images.py
```diff
@@ -318,6 +318,9 @@ def test_delete_image_recreate(
         get_image(api_client, unique_name)
         delete_image(api_client, unique_name, wait_timeout)
 
+        get_image(api_client, image_name)
```
lanfon72, line 321: The file of `image_name` should not existed as it already deleted in previous line?
Could you double confirm that it is caused by this test case or others?
(or do you have the test report for reproduce?)

brandboat: Thanks for the comment, what you mentioned in [L319](https://github.com/harvester/tests/blob/4d4d22b397e4fa8ff73ce193c6fce6b78624d356/harvester_e2e_tests/integrations/test_1_images.py#L319) is to delete image name of `unique_name`, while L322 (this pr) wants to delete image name of `image_name` which was created in [L300](https://github.com/harvester/tests/blob/4d4d22b397e4fa8ff73ce193c6fce6b78624d356/harvester_e2e_tests/integrations/test_1_images.py#L300).

Below is the video of testing `delete_image_recreate` harvester UI image page (before patching this pr) by using command as follows:
```sh
tox -e python3.11 --result-json=test_result.json -- -l harvester_e2e_tests --json-report --html=test_report/test_result.html -m 'not delete_host and not upgrade' -k 'delete_image_recreate or create_image_url'
```

https://github.com/user-attachments/assets/79c1b3fc-63da-42c7-9d84-3d9721bf01d7

As you can see, there were two images remained after test finished, `opensuse-11h10m13s466515-09-20` and `ubuntu-11h10m13s466515-09-20`.

khushboo-rancher: @lanfon72 If agreed with @brandboat's comment, please resolve the conversation and proceed.

lanfon72: No, it is actually a test case design problem.

The image `image_name` which created in test case **test_create_image_url**, and it will be used in **test_delete_image_recreate** and **test_edit_image_in_use**, and finally be deleted in **test_edit_image_in_use** (the last line)

so in this case, if we just execute those two test cases, the image will exist is expected. (by the implementation)

and if we include this change, the test case **test_edit_image_in_use** will fail in the first assertion.

lanfon72:
@brandboat, thanks for your time to investigate this issue and trying to fix it.
but as it is the side-effect when the whose  test class `TestBackendImages` was not fully executed, so we can't include the change. (otherwise another test case will break when the test class fully executed, as I mentioned in https://github.com/harvester/tests/pull/1539#discussion_r1795885053)

I will convert this to an issue, then we will try to remove the side effect to make sure no assets will remain.

brandboat:
> @brandboat, thanks for your time to investigate this issue and trying to fix it.
but as it is the side-effect when the whose test class TestBackendImages was not fully executed, so we can't include the change. (otherwise another test case will break when the test class fully executed, as I mentioned in https://github.com/harvester/tests/pull/1539#discussion_r1795885053)
> 
> I will convert this to an issue, then we will try to remove the side effect to make sure no assets will remain.

Thanks for the detailed explanation ! So this pr isn't the right approach, I'll close it. Thanks.

PR #1580, opened 2024-10-09
fix: 1568 Fixing S3 Connection Building

### LINKED ISSUE #1568: [TEST] TestBackupRestore::test_connection[S3] test fails making consecutive tests fail for backup & restore

```
 @pytest.mark.dependency()
    def test_connection(self, api_client, backup_config, config_backup_target):
        code, data = api_client.settings.backup_target_test_connection()
>       assert 200 == code, f'Failed to test backup target connection: {data}'
E       AssertionError: Failed to test backup target connection: {"errors":["can't connect to backup target \u0026{Type:s3 Endpoint:http://172.19.99.205:9000 AccessKeyID:minioadmin SecretAccessKey:minioadmin BucketName:giants BucketRegion:ecmlab Cert: VirtualHostedStyle:false}, error: secrets \"harvester-backup-target-secret\" not found"]}
E       assert 200 == 503

api_client = HarvesterAPI('https://172.19.xyz.x ...

```

irishgordo:
* fixes s3 connection building, allowing us to wait for a short period of time to allow the secret for longhorn-system to be built on Harvester side

Resovles: fix/1568
See also: https://github.com/harvester/tests/issues/1568

Fixes #1568 

Adds a bit of time before yielding back the backup config -> as Harvester takes a bit more time usually for it to have the secret built in longhorn-system in v1.4.0-rc1

In the video, it is demo'd with an external bare-metal hp dl160 2 node v1.4.0-rc1 cluster that we can indeed replicate the issue outlined in 1568.  We see we are moving to fast trying to build the fixture and the secret hasn't yet built on Harvester side in time.  But if we give it a bit of a sleep, to allow Harvester to catch up to build the secret before yielding back the spec, everything is good.  The video illustrates the points back to back.

https://github.com/user-attachments/assets/0acb4385-0fdb-49d6-a090-3117dfcbc2e5

### DIFF
--- a/harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
+++ b/harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
@@ -121,6 +121,8 @@ def config_backup_target(api_client, conflict_retries, backup_config, wait_timeo
         f'Failed to update backup target to {backup_type} with {config}\n'
         f"API Status({code}): {data}"
     )
+    # sleeping to allow longhorn secret to be built on Harvester side
+    sleep(5)
 
     yield spec
 


harvester_e2e_tests/integrations/test_4_vm_backup_restore.py
```diff
@@ -121,6 +121,8 @@ def config_backup_target(api_client, conflict_retries, backup_config, wait_timeo
         f'Failed to update backup target to {backup_type} with {config}\n'
         f"API Status({code}): {data}"
     )
+    # sleeping to allow longhorn secret to be built on Harvester side
+    sleep(5)
```
khushboo-rancher, line 125: Do we have any way of checking the secret existence? Can we check that in a loop with some sleep?

noahgildersleeve: Could we loop through checking the status of the backup target by looping through something like `api_client.settings.backup_target_test_connection`? Will that return true if the secret isn't there?

irishgordo:
I've converted this back to "draft" as the fix will require checking a status on Longhorn that @lanfon72  was mentioning .
Sleeping is a band-aid -> and really this is most noticable on a loadout of:
1. A seperate located MinIO (S3 Compatible) Instance, running on a VM, "not" on MinIO Operator based Tenant (so no K8s helm chart based install of MinIO) -> located over a 1GB Symetrical Backbone Network and is backed by SSD not NVMe
2. A seperate Harvester Cluster, running on bare-metal, "that does not run" the MinIO instance -> backed by a 1GB Symetrical Backbone Network and is backed by SSD not NVMe
3. Running these tests on a seperate host/workstation, that is also backed by a 1GB Symetrical Backbone Network

It's not that this test is "failing" per say due to the test itself it's the hardware + network based setup that is impacting this test.

When running locally or over faster storage + network -> test failures happens less frequently.

Issue #1570, opened 2024-10-04
[TEST] Update open suse image and ubuntu in CI to 15.6 & 24.04 respectively


khushboo-rancher:
## What's the test to develop? Please describe

Update the [config file](https://github.com/harvester/tests/blob/main/config.yml) to get the latest versions of open suse to 15.6 and  ubuntu to 24.04

noahgildersleeve:
Fixed via #1598

PR #1598, opened 2024-10-16
updated backing images for openSUSE and Ubuntu.

noahgildersleeve:
Also added in a fixture for vm_network_interact to make those pass

Issue #1570 

This updates the OS images for the tests
I also added in a missing fixture I found while validating the images

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -22,11 +22,11 @@ sleep-timeout: 3
 # script location to manipulate node power cycle
 node-scripts-location: 'scripts/vagrant'
 
-opensuse-image-url: https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.5/images/openSUSE-Leap-15.5.x86_64-NoCloud.qcow2
+# Images and their sha512sums
+opensuse-image-url: https://rsync.opensuse.org/repositories/Cloud%3A/Images%3A/Leap_15.6/images/openSUSE-Leap-15.6.x86_64-NoCloud.qcow2
 # sha512sum for opensuse image-url
 opensuse-checksum: ''
-
-ubuntu-image-url: https://cloud-images.ubuntu.com/releases/releases/22.04/release/ubuntu-22.04-server-cloudimg-amd64.img
+ubuntu-image-url: https://cloud-images.ubuntu.com/releases/releases/24.04/release/ubuntu-24.04-server-cloudimg-amd64.img
 # sha512sum for ubuntu image-url
 ubuntu-checksum: ''
 
--- a/harvester_e2e_tests/integrations/test_5_vm_networks_interact.py
+++ b/harvester_e2e_tests/integrations/test_5_vm_networks_interact.py
@@ -13,7 +13,8 @@ import paramiko
 pytest_plugins = [
     "harvester_e2e_tests.fixtures.api_client",
     "harvester_e2e_tests.fixtures.networks",
-    "harvester_e2e_tests.fixtures.virtualmachines"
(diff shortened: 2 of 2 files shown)


khushboo-rancher:
Have we run the tests with the changes?

harvester_e2e_tests/integrations/test_5_vm_networks_interact.py
```diff
@@ -13,7 +13,8 @@
 pytest_plugins = [
     "harvester_e2e_tests.fixtures.api_client",
     "harvester_e2e_tests.fixtures.networks",
-    "harvester_e2e_tests.fixtures.virtualmachines"
+    "harvester_e2e_tests.fixtures.virtualmachines",
+    "harvester_e2e_tests.fixtures.images"
```
albinsun, line 17: I don't get this.
Add a new global fixture but no new test case or local fixtures?

config.yml
```diff
@@ -22,11 +22,11 @@ sleep-timeout: 3
 # script location to manipulate node power cycle
 node-scripts-location: 'scripts/vagrant'
 
-opensuse-image-url: https://download.opensuse.org/repositories/Cloud:/Images:/Leap_15.5/images/openSUSE-Leap-15.5.x86_64-NoCloud.qcow2
+opensuse-image-url: https://rsync.opensuse.org/repositories/Cloud%3A/Images%3A/Leap_15.6/images/openSUSE-Leap-15.6.x86_64-NoCloud.qcow2
```
albinsun, line 25: Any difference between _rsync.opensuse.org_ and _download.opensuse.org_?

noahgildersleeve: I looked up the most recent image locations and this was what showed up. It shouldn't really matter since they are both running as web servers serving the files. The only downside I see is that if the server changes we'll have to change the URL again. Also this URL doesn't use colons in the URL which should help a bit with compatibility.

harvester_e2e_tests/integrations/test_5_vm_networks_interact.py
noahgildersleeve: I was validating that the OS image update didn't break any other tests and I saw that this test wasn't working. It was missing the the images fixture and was failing since it uses `image_opensuse` which is a part of  `fixtures.images`. I added in the fix to this PR since I was already creating it. 

I can add it in another PR, but I thought it was a small enough change to make here. The test passed with the proper

albinsun (approved):
LGTM

lanfon72 (approved):
LGTM

albinsun:
> Have we run the tests with the changes?

Guys, what is the criteria to change default image?

`opensuse-leap-v15.5` is reaching EOL but `ubuntu-v22.04` is not out-of-date yet.
Ref. https://endoflife.date/ubuntu.

lanfon72:
It's the default value, which means it will not break anything in Jenkins environment; the fixtures added is required if you run test on the test file directly (although we will not recommend that way)

noahgildersleeve:
> > Have we run the tests with the changes?
> 
> Guys, what is the criteria to change default image?
> 
> `opensuse-leap-v15.5` is reaching EOL but `ubuntu-v22.04` is not out-of-date yet. Ref. https://endoflife.date/ubuntu.

@albinsun I was updating Ubuntu due to the [ticket](https://github.com/harvester/tests/issues/1570) requirements. I assume we're going to 24.04 since it's the new LTS release and 22.04 isn't. I know that I just updated several of my environments to 24.04 since 20.04 is going EOL soon

Issue #1627, opened 2024-10-25
[BUG] Rancher Integration tests failing with Rancher v2.9.3


m-ildefons:
Rancher integration tests fail with Rancher v2.9.3, since the mechanism to create cloud credential for an imported Harvester cluster no longer works with kubeconfig files that aren't bound to a token in Rancher.
The test suite needs to be adapted to use the correct mechanism to generate the kubeconfig file via Rancher (instead of directly with Harvester) to make sure it's bound to a token

Logs:
```

api_client = HarvesterAPI('https://172.19.108.29', 'Bearer token-4n9bs:q5s2dc6v9jc5xvtlfg66886c9tpgdms9hzgbmw4s5vhj956qcw44n9')
rancher_api_client = RancherAPI('https://rancher.172.19.108.29.sslip.io', 'Bearer token-q52mh:gdc7rjgcqwrqdcj8gm4pdd2sr6hfm5dxn8xlh6dp7z6bzph2rmb2sh')
harvester_mgmt_cluster = {'id': 'c-m-lcjsqmsd', 'name': 'hvst-11h38m40s902870-10-25'}
unique_name = '11h38m40s902870-10-25'

    @pytest.fixture(scope='module')
    def harvester_cloud_credential(api_client, rancher_api_client,
                                   harvester_mgmt_cluster, unique_name):
        harvester_kubeconfig = api_client.generate_kubeconfig()
        code, data = rancher_api_client.cloud_credentials.create(
            unique_name,
            harvester_kubeconfig,
            harvester_mgmt_cluster['id']
        )
>       assert 201 == code, (
            f"Failed to create cloud credential with error: {code}, {data}"
        )
E       AssertionError: Failed to create cloud credential with error: 500, {'baseType': 'error', 'code': 'ServerError', 'message': 'failed to process harvester cloud credential expiration: failed to get harvester cloud credential expiration from kubeconfig: failed to get harvester credential config token: tokens.management.cattle.io "kubeconfig-user-rxsbfxpxfr" not found', 'status': 500, 'type': 'error'}
E       assert 201 == 500

api_client = HarvesterAPI('https://172.19.108.29', 'Bearer token-4n9bs:q5s2dc6v9jc5xvtlfg66886c9tpgdms9hzgbmw4s5vhj956qcw44n9')
code       = 500
data       = {'baseType': 'error',
 'code': 'ServerError',
 'message': 'failed to process harvester cloud credential expiration: failed '
[...]

PR #1628, opened 2024-10-25
Wip/fix get harvester kubeconfig

m-ildefons:

Issue # https://github.com/harvester/tests/issues/1627

Creating a cloud credential from a kubeconfig file that has not been created through Rancher is wrong, because then Rancher can't determine the expiration date of that kubeconfig and will return an error.
To fix this, the test suite needs to create the kubeconfig file through Rancher

### DIFF
--- a/apiclient/rancher_api/api.py
+++ b/apiclient/rancher_api/api.py
@@ -67,19 +67,27 @@ class RancherAPI:
 
     def _get(self, path, **kwargs):
         url = urljoin(self.endpoint, path)
-        return self.session.get(url, **kwargs)
+        resp = self.session.get(url, **kwargs)
+        print(f"{resp.request.method} {resp.request.url}")
+        return resp
 
     def _post(self, path, **kwargs):
         url = urljoin(self.endpoint, path)
-        return self.session.post(url, **kwargs)
+        resp = self.session.post(url, **kwargs)
+        print(f"{resp.request.method} {resp.request.url}")
+        return resp
 
     def _put(self, path, **kwargs):
         url = urljoin(self.endpoint, path)
-        return self.session.put(url, **kwargs)
+        resp = self.session.put(url, **kwargs)
+        print(f"{resp.request.method} {resp.request.url}")
+        return resp
 
     def _delete(self, path, **kwargs):
         url = urljoin(self.endpoint, path)
-        return self.session.delete(url, **kwargs)
+        resp = self.session.delete(url, **kwargs)
+        print(f"{resp.request.method} {resp.request.url}")
+        return resp
 
     def authenticate(self, user, passwd, **kwargs):
         path = "v3-public/localProviders/local?action=login"
--- a/apiclient/rancher_api/managers.py
+++ b/apiclient/rancher_api/managers.py
@@ -31,7 +31,7 @@ class BaseManager:
             raise ReferenceError("API object no longer exists")
         return self._api()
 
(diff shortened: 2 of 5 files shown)


FrankYang0529 (approved):

apiclient/rancher_api/api.py
```diff
@@ -67,19 +67,27 @@ def __repr__(self):
 
     def _get(self, path, **kwargs):
         url = urljoin(self.endpoint, path)
-        return self.session.get(url, **kwargs)
+        resp = self.session.get(url, **kwargs)
+        print(f"{resp.request.method} {resp.request.url}")
```
lanfon72, line 71: please remove the test output.

apiclient/rancher_api/managers.py
```diff
@@ -678,6 +678,7 @@ def delete(self, name, *, raw=False):
 
 class ClusterManager(BaseManager):
     PATH_fmt = "v3/cluster/{uid}"
+    PATH1_fmt = "v3/clusters/{uid}"
```
lanfon72, line 681: it looks we didn't use this, is there any reason to add it?

harvester_e2e_tests/fixtures/settings.py
```diff
@@ -15,7 +15,9 @@ def __init__(self):
 
         def _storage_net_configured(self):
             code, data = self.settings.get('storage-network')
-            if (cs := data.get('status', {}).get('conditions')):
```
lanfon72, line 18: The assignment expressions (`:=`) is introduced in python 3.8, which is the version we targeting to support, is there any reason to replace it?

harvester_e2e_tests/integrations/test_9_rancher_integration.py
```diff
@@ -130,7 +130,14 @@ def harvester_mgmt_cluster(api_client, rancher_api_client, unique_name, polling_
 @pytest.fixture(scope='module')
 def harvester_cloud_credential(api_client, rancher_api_client,
                                harvester_mgmt_cluster, unique_name):
-    harvester_kubeconfig = api_client.generate_kubeconfig()
+    code, data = rancher_api_client.clusters.generate_kubeconfig(
```
lanfon72, line 133: I think the two configurations are same, could you check with @albinsun about the change?

lanfon72 (commented):
I think all changes are not harmful, but could you ask QAs to review PR rather than engineers next time?

and also cc @khushboo-rancher for your PR would be nice.

harvester_e2e_tests/fixtures/settings.py
m-ildefons: It's not supported in Python 3.6. We maintain a Python 3.6 environment in our tox config. Either we drop that as well, or we stick to Python 3.6 compatibility.

harvester_e2e_tests/integrations/test_9_rancher_integration.py
m-ildefons: They are not the same. One uses Rancher's API endpoint and authenticates with a Rancher token and the other one uses Harvester's API endpoint directly.
For these tests we need to use Rancher's API endpoint.
Please read the issue linked up top.

m-ildefons:
> I think all changes are not harmful, but could you ask QAs to review PR rather than engineers next time?

Your review was requested nearly a week ago.

Issue #2028, opened 2025-04-29
[TEST] Refactor to make sure fixtures are gracefully teardowned to avoid affect following test cases


albinsun:
## What's the test to develop? Please describe

Sometimes unteardowned fixtures (vm network, cluster network, ...) will affect following test case running. To let each test suite is self-consistent we need to actively check teardown is successful before get into next case.

## Prerequisite and dependency of test
n/a

## Describe the items of the test development (DoD, definition of done) you'd like
Idealy, teardown should only be handle in fixture cuz
1. Test case self does not guarantee PASS
2. Teardowns in test case will not be execute as long as test fail.
3. An exception is the object is generated inside test case self.
 
#### _harvester_e2e_tests/apis/*
- [ ] xxx
#### _harvester_e2e_tests/integrations/*
- [ ] test_0_storage_network.py 
- [ ] xxx

## Additional context
n/a

PR #2045, opened 2025-05-15
Refactor to not ignoring teardown on test_0

albinsun:
### Source Issue
* https://github.com/harvester/tests/issues/2028

### Implementation
Because storage network enable and disable are independent.
By separate single TC into enable and disable TCs we can ensure disable will still execute even enable is FAIL.

### Verification
image


lanfon72 (approved):
LGTM

irishgordo (approved):
lgtm :smile: :+1:

PR #2059, opened 2025-06-04
Avoid deleting upgrade CR to affect following upgrades before v1.5.0

albinsun:
* https://github.com/harvester/harvester/issues/7654

* Before v1.5.0 fix, when deleting upgrade CR, there is a high chance for upgradeLog does not cleanup managedchart all the time and affects following upgrade can not start.
* Adding skip marker in test cases includes upgrade CR deletion if current version < `v1.5.0`


starbops (approved):
LGTM, thank you.

PR #2062, opened 2025-06-10
feat: add harvester-version for self built Harvester testing

Yu-Jack:

Issue https://github.com/harvester/harvester/issues/6558

When testing self built Harvester, `skip_version_if` will fail to check. Hence, I add a new config to fake it for convenience.

For example, because self built Harvester has dirty commit which is just a commit hash, not a version `v1.x.x`, we can give a custom version.


Yu-Jack:
I opened it, thanks! https://github.com/harvester/tests/issues/2064 cc @lanfon72

PR #2066, opened 2025-06-13
feat: add copilot instruction

Yu-Jack:
Issue https://github.com/harvester/harvester/issues/8148


Yu-Jack:
This is example prompt.

---

## What I want to do

I want to test the creation of encrypted and decrypted images in Harvester.

Please help me:

- Add a new test class in a new file: `integrations/test_1_security_images.py`
- Use `integrations/test_1_images.py` as a reference

---

## Shared Fixtures

Please reuse or define the following shared fixtures:

1. **Secret with encryption parameters**

   ```json
   {
     "CRYPTO_KEY_CIPHER": "aes-xts-plain64",
     "CRYPTO_KEY_HASH": "sha256",
     "CRYPTO_KEY_PROVIDER": "secret",
     "CRYPTO_KEY_SIZE": "256",
     "CRYPTO_KEY_VALUE": "test",
     "CRYPTO_PBKDF": "argon2i"
   }
   ```

2. **StorageClass with encryption settings and secret reference**

   ```json
   {
     "csi.storage.k8s.io/node-publish-secret-name": "<secret_name>",
     "csi.storage.k8s.io/node-publish-secret-namespace": "<namespace>",
     "csi.storage.k8s.io/node-stage-secret-name": "<secret_name>",
     "csi.storage.k8s.io/node-stage-secret-namespace": "<namespace>",
     "csi.storage.k8s.io/provisioner-secret-name": "<secret_name>",
     "csi.storage.k8s.io/provisioner-secret-namespace": "<namespace>",
     "encrypted": "true",
     "migratable": "true",
     "numberOfReplicas": "3",
     "staleReplicaTimeout": "30"
   }
   ```

3. **Shared image**

   This is the base image used to create encrypted image.

---

## Main Test Cases

Please separate the following into individual test functions within the same class:

1. **Create an Encrypted Image**

   - Use the StorageClass created in step 2 of shared fixture.
   - Use the following spec snippet:
     ```python
     data["spec"]["securityParameters"] = {
       "cryptoOperation": "encrypt",
       "sourceImageName": source_image_name,
       "sourceImageNamespace": namespace,
     }
     data["spec"].pop("checksum", None)
     data["spec"].pop("url", None)
     spec.metadata.annotations["harvesterhci.io/storageClassName"] = storage_class_name
     spec.displayName = "<Encrypted Image Name>"
     ```

2. **Create a Decrypted Image**

   - Use the encrypted image created above as the source.
   - Use the following spec snippet:
     ```python
     data["spec"]["securityParameters"] = {
       "cryptoOperation": "decrypt",
       "sourceImageName": source_encrypted_image_name,
       "sourceImageNamespace": namespace,
     }
     data["spec"].pop("checksum", None)
     data["spec"].pop("url", None)
     spec.displayName = "<Decrypted Image Name>"
     ```

---

## Additional Notes

- Secret API: `apiclient/harvester_api/managers/secrets.py`
- StorageClass API: `apiclient/harvester_api/managers/storageclasses.py`
- These are Harvester APIs and should remain in the `apiclient/harvester_api/managers` directory.
- It only has `create_by_file` method in `apiclient/harvester_api/managers/images.py`, create another method to fit these test cases
- Only use `image_ubuntu`.

.github/instructions/copilot-instructions.md
```diff
@@ -0,0 +1,124 @@
...
+        shared_data = shared_fixture
+
+        # 2. Call the API
+        code, data = api_client.sss.xxx()
+
+        # 3. Assert the result
+        assert code == xxxx, f"Expected status code xxxx, got {code} with data: {data}"
+```
+
+## Harvester API Manager Template
```
lanfon72, line 92: Harvester API is not tight with our pytest test cases, we would separate it to another repo in near future, so this would not be used in here.

.github/instructions/copilot-instructions.md
```diff
@@ -0,0 +1,124 @@
...
+- Load the new resource in `apiclient/harvester_api/api.py` and register it in `apiclient/harvester_api/managers/__init__.py`.
+- Refer to `apiclient/harvester_api/managers/images.py` as a sample implementation.
+- Follow these standard status codes to validate Harvester API responses:
+  - `201` → Resource successfully created
+  - `200` → Operation succeeded (general case)
+
+## Don't Do
+
+- Do **not** use method names with `_` prefixes in tests, such as `api_client.[resource]._[method]`.
+- Do **not** define a function inside another function, for example:
```
lanfon72, line 26: for example as https://github.com/harvester/tests/blob/71d70d9ddcae47f3c1eefb57be3450341629d33b/harvester_e2e_tests/fixtures/virtualmachines.py#L119-L368

so it would not be **Don't**, but need more consideration.

lanfon72 (changes requested):
Harvester API would be more complicated, so I feel copilot would not be useful in it.

.github/instructions/copilot-instructions.md
Yu-Jack: If so, we can keep it here until we move it out.

.github/instructions/copilot-instructions.md
Yu-Jack: Okay. I didn't notice we have this pattern, thanks!

PR #2067, opened 2025-06-17
[apiclient] add customized build version detection

### LINKED ISSUE #2064: Fix skip_version_if condition when testing with self built Harvester
## Description

Since it's not a good practice to open another path to bypass it, so we should figure it out how to make this function work with self built Harvester.

Former Related PR: https://github.com/harvester/tests/pull/2062

## How to Reproduce

1. Build a new Harvester image.
2. Deploy it by replacing image in deployment yaml. 
3. Test it with `pytest harvester_e2e_tests/apis/test_volumes.py --username admin --password {password} --endpoint https://{ip}`

The pytest doesn't skip the test case `test_create_without_name` due to `skip_version_if` check.

The version log is:

```
Cluster Version: 3874194f-dirty
Checking version: >= v1.4.0 against 3874194f-dirty
```

You can just replace `rancher:harvester:v1.5.0` image with mine `jk82421/harvester:v1.5.0.1` to test it.

cc @lanfon72

lanfon72:
Issue #2064

let KubeVirt's version be the customized build's version, it might not be 100% correct, but close to.

this would fix #2064 directly.

### DIFF
--- a/apiclient/harvester_api/api.py
+++ b/apiclient/harvester_api/api.py
@@ -52,7 +52,12 @@ class HarvesterAPI:
                 cur = "v8.8" if "master" in cur else cur
                 self._version = parse_version(f"{cur}.99")
             except ValueError:
-                self._version = parse_version(ver)
+                # 1. va.b.c => valid version
+                # 2. 3874194f-dirty (customized build)
+                #   Use KubeVirt version as the customized version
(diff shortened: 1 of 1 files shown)


Yu-Jack (approved):
Thanks. It works.

PR #2238, opened 2025-10-28
[e2e] new test cases for CPU Pinning feature

lanfon72:

Issue #1420

- new method `hosts.cpu_manager`
- new property `VMSpec.cpu_pinning`
- new method `vm.get_migratables`
- new test cases `TestCPUManager` to test hosts' CPU Manager
- new test cases `TestPinCPUonVM` to test VM with CPU pinned

### DIFF
--- a/apiclient/harvester_api/managers/virtualmachines.py
+++ b/apiclient/harvester_api/managers/virtualmachines.py
@@ -143,4 +143,4 @@ class VMManager140(VirtualMachineManager):
     def get_migratables(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
         path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
         params = dict(action="findMigratableNodes")
-        return self.create(path, params=params, raw=raw)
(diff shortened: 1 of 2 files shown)


apiclient/harvester_api/managers/virtualmachines.py
```diff
@@ -139,3 +139,8 @@ class VMManager131(VirtualMachineManager):
 class VMManager140(VirtualMachineManager):
     support_to = "v1.4.0"
     Spec = VMSpec140
+
+    def get_migratables(self, name, namespace=DEFAULT_NAMESPACE, *, raw=False):
+        path = self.PATH_fmt.format(uid=f"/{name}", ns=namespace)
+        params = dict(action="findMigratableNodes")
+        return self.create(path, params=params, raw=raw)
```
asc5543: Does it should be `self.get_status`?

lanfon72: we should use `self._create` to call the post method, it's typo.

PR #2239, opened 2025-10-28
Introduce Robot framework

khushboo-rancher:
## PR fixes the issue:
https://github.com/harvester/tests/issues/2244

## Description:

This PR introduces a modular, maintainable, and scalable Robot Framework E2E test architecture for Harvester, structured as follows:

4-Layer Design :

Layer 1: .robot files define test cases.

Layer 2: .resource files manage test keywords.

Layer 3: Python keyword wrappers delegate all logic to component classes — no direct API calls here.

Layer 4: Resource components (VM, Volume, Image, Host, etc.) delegate to CRD and REST/API implementations, ensuring all API calls are isolated and test logic remains clean.

README file has all the details of the framework.

## Special notes for reviewers:
Please refer or run test 'Test VM Basic Lifecycle' for review purposes. The tests in negative directory are sample and will be modified later.

### DIFF
--- a/harvester_robot_tests/README.md
+++ b/harvester_robot_tests/README.md
@@ -1,0 +1,531 @@
+# Harvester Robot Framework E2E Tests
+
+This repository contains the Harvester end-to-end tests implemented using Robot Framework, following a layered architecture pattern.
+
+## Table of Contents
+
+- [Overview](#overview)
+- [Architecture](#architecture)
+- [Requirements](#requirements)
+- [Installation](#installation)
+- [Kubernetes Configuration (CRD Setup)](#kubernetes-configuration-crd-setup)
+- [Configuration](#configuration)
+- [Running Tests](#running-tests)
+- [Writing New Tests](#writing-new-tests)
+- [Directory Structure](#directory-structure)
+- [Troubleshooting](#troubleshooting)
+
+## Overview
+
+The Harvester Robot Framework test suite provides comprehensive end-to-end testing for Harvester HCI, including:
+
+- Virtual machine lifecycle management
+- Image operations (upload, download, management)
+- Volume operations and storage management
+- Network configuration (VLAN, cluster networks)
+- Backup and restore operations
+- Node management and HA scenarios
+- Rancher integration testing
+- Upgrade scenarios
+
+### Implementation Strategies
+
+This test framework supports **two implementation strategies**:
+
+1. **CRD (Kubernetes Custom Resources)** - Default ✅
+   - Uses Kubernetes APIs directly
+   - Works with KubeVirt and Harvester CRDs
+   - Requires direct cluster access
+   
+2. **REST (Harvester REST API)** - Alternative
+   - Uses Harvester REST API endpoints
+   - Suitable for remote testing
+   - Set `HARVESTER_OPERATION_STRATEGY=rest` to use
+
+## Architecture
+
+The test framework follows a strict 4-layer architecture:
+
+```
+┌────────────────────────────────────────────────────────────────┐
+│ Layer 1: tests/*.robot - Test Case Definition                  │
+│  - Define test scenarios and workflows                         │
+│  - Import Layer 2 resource files                               │
+└────────────────────────────────────────────────────────────────┘
+                          ↓
+┌────────────────────────────────────────────────────────────────┐
+│ Layer 2: keywords/*.resource - Keyword Definition              │
+│  - Define reusable keywords                                    │
+│  - Import Layer 3 Python keyword libraries                     │
+└────────────────────────────────────────────────────────────────┘
+                          ↓
+┌────────────────────────────────────────────────────────────────┐
+│ Layer 3: libs/keywords/*_keywords.py - Keyword Wrappers        │
+│  - Create component instances                                  │
+│  - Delegate to Layer 4 components                              │
+│  - NO direct API calls                                         │
+└────────────────────────────────────────────────────────────────┘
+                          ↓
+┌────────────────────────────────────────────────────────────────┐
+│ Layer 4: libs/{vm,image,host,volume}/*.py - Components         │
+│  - Component wrapper: delegates to CRD or REST implementation  │
+│  - CRD/REST implementation: makes actual API calls             │
+│  - Base: abstract base classes                                 │
+└────────────────────────────────────────────────────────────────┘
+```
+
+**Important**: Each layer can ONLY call functions from the next layer or the same layer. Skip-layer dependencies are strictly forbidden.
+
+## Requirements
+
+### Prerequisites
+
+1. A running Harvester cluster
+2. Python 3.8 or higher
+3. `kubectl` configured with access to the cluster (for CRD strategy)
+4. Access to Harvester API endpoint (for REST strategy)
+5. Kubernetes credentials configured in kubeconfig
+
+### Optional Requirements
+
+For specific test scenarios:
+
+- **Network Tests**: VLAN configuration with proper routing and DHCP
+- **Backup Tests**: S3-compatible storage or NFS server
+- **Node Management Tests**: Node power management scripts
+- **Rancher Integration Tests**: External Rancher cluster
+
+## Installation
+
+1. Clone the repository:
+```bash
+git clone https://github.com/harvester/harvester-robot-tests.git
+cd harvester-robot-tests
+```
+
+2. Create a virtual environment:
+```bash
+python3 -m venv venv
+source venv/bin/activate  # On Windows: venv\Scripts\activate
+```
+
+3. Install dependencies:
+```bash
+pip install -r requirements.txt
+```
+
+## Kubernetes Configuration (CRD Setup)
+
+### Default Strategy: CRD (Kubernetes Custom Resources)
+
+The test framework uses **CRD strategy by default**. This requires Kubernetes cluster access.
+
+### Step 1: Get Kubeconfig from Harvester Cluster
+
+#### From Harvester UI
+1. Login to Harvester Dashboard
+2. Go to `Support`
+3. Click `Download Kubeconfig`
+4. Save the file
+
+#### From Harvester Command Line
+```bash
+# SSH into a Harvester node
+ssh rancher@<harvester-ip>
+
+# Get the kubeconfig
+sudo cat /etc/rancher/rke2/rke2.yaml | sed 's/127.0.0.1/<HARVESTER_IP>/g'
+```
+
+### Step 2: Set Up Kubeconfig
+
+#### Option A: Using Default Kubeconfig Location
+
+```bash
+# The default location is $HOME/.kube/config
+# Copy your cluster's kubeconfig here:
+
+mkdir -p $HOME/.kube
+cp /path/to/your/harvester-kubeconfig.yaml $HOME/.kube/config
+chmod 600 $HOME/.kube/config
+
+# Verify access:
+kubectl get nodes
+```
+
+#### Option B: Using Custom Kubeconfig Location
+
+```bash
+# Export the path:
+export KUBECONFIG=/path/to/your/kubeconfig.yaml
+
+# Or set it permanently in shell profile:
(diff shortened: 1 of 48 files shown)


harvester_robot_tests/scripts/node-management/power_off.sh
lanfon72: The script looks like binding with the vagrant environment, but we already have scripts in https://github.com/harvester/tests/tree/main/scripts/vagrant , can't we reuse it?

harvester_robot_tests/tests/negative/vm_node_failure.robot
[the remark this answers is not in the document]
khushboo-rancher: This is just a sample, I'll refine this later. Thanks

khushboo-rancher:
This supports CRD now. All the test will execute CR operations instead of REST APIs by default.

harvester_robot_tests/libs/harvester_api.py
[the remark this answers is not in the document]
khushboo-rancher: Thanks for the suggestion, but implementing a robotframework-requests-style wrapper doesn't align with our architecture and goals:

- CRD-First Strategy: Our framework is designed to use Kubernetes CRDs as the primary testing interface, creating a full REST wrapper might add significant complexity to the framework.
- Layered Architecture: Our 4-layer design intentionally abstracts implementation details (CRD vs REST) away from test cases, providing domain-specific keywords rather than exposing low-level HTTP operations.

The current design allows us to:

- Write high-level, readable tests that don't change regardless of implementation strategy.
- Switch between CRD and REST transparently via strategy pattern.

If there are specific REST scenarios that require better support, we can enhance the existing HarvesterAPIClient or improve the REST strategy implementation rather than creating a new wrapper layer.

harvester_robot_tests/tests/regression/test_vm.robot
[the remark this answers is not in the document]
khushboo-rancher: I'm not sure where do you see sub tests. Could you check the test case again, it has been changed from last time you've reviewed and elaborate more where you want to keep running the test if the prevoius test step fails.

harvester_robot_tests/README.md
```diff
@@ -0,0 +1,529 @@
...
+
+### Environment Variables Reference
+
+The `.env.example` file contains all available configuration options:
+
+#### Required Variables (for CRD strategy):
+- `KUBECONFIG` - Path to kubeconfig file (defaults to `~/.kube/config`)
+
+#### Optional Variables:
+- `HARVESTER_ENDPOINT` - Harvester API endpoint (default: `https://10.115.252.130`)
```
albinsun: Is this a shared harvester endpoint for test developing?
If not, how about use default VIP `192.168.0.131` in _ipxe-example_?

harvester_robot_tests/libs/host/rest.py
```diff
@@ -0,0 +1,165 @@
+"""
+Host Rest Implementation - makes actual API calls using get_harvester_api_client()
+"""
+import time
+from datetime import datetime, timedelta
+from utility.utility import get_harvester_api_client
+from utility.utility import logging
+
+
+class Rest:
```
albinsun: `class Rest(Base):`

harvester_robot_tests/libs/image/rest.py
```diff
@@ -0,0 +1,142 @@
...
+Image Rest Implementation - makes actual API calls using get_harvester_api_client()
+"""
+import time
+from datetime import datetime, timedelta
+from utility.utility import get_harvester_api_client
+from utility.utility import get_retry_count_and_interval
+from utility.utility import logging
+
+
+class Rest:
```
albinsun: `class Rest(Base):`

harvester_robot_tests/libs/image/crd.py
```diff
@@ -0,0 +1,263 @@
...
+                condition_status="True",
+                timeout=timeout
+            )
+            logging(f"Image {namespace}/{image_name} is ready")
+            return True
+        except Exception as e:
+            logging(f"Image did not become ready: {e}")
+            raise
+
+    def wait_for_deleted(self, image_name, timeout=600):
```
albinsun: Can use global declared timeout?

harvester_robot_tests/libs/utility/utility.py
```diff
@@ -0,0 +1,117 @@
...
+        logger.info(message)
+
+
+def init_k8s_api_client():
+    """Initialize Kubernetes API client"""
+    # Suppress SSL warnings first
+    import urllib3
+    urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+    from kubernetes import client, config
```
albinsun: Just notice there is a 3rd-party library [KubeLibrary](https://github.com/devopsspiral/KubeLibrary) recommanded by RF 
See https://robotframework.org/#resource

Do you think it can help us to avoid rebuild some wheel?

innobead:
> > This supports CRD now. All the test will execute CR operations instead of REST APIs by default.
> 
> @khushboo-rancher as the concern that I and @noahgildersleeve mentioned in QA sync up, please have a discussion with @Vicente-Cheng and @innobead, and make sure that it is OK to re-implement CRD operations rather than re-use existing HarvesterAPI library when write robot's test cases.

All exposed interfaces, including the Public/REST API, UI, and Terraform provider, are solely relying on CR operations at the backend, which include validation and mutation. In fact, the new UI extension mainly depends on CR operations due to its design, but some action-driven operations still rely on the REST API.

Creating a CR-based library would be beneficial to cover tests from various interfaces. However, this will take time because not all mutation and validation have been moved to Harvester webhooks. Don't get blocked on execution, as this isn't an exclusive choice. __*The recommendation is to prioritize CR first, but fall back to the API if needed. You can improve this over time, but ultimately, CR should follow once the above issue is resolved.*__ 

P.S. Some mutation and validation logic are currently outside Harvester webhooks. We should eventually move them back to the Harvester webhooks to avoid code duplication or implementations scattered across various parts. A unified management approach is required. This will simplify maintenance and QA efforts in the future.

cc @harvester/maintainer  @Vicente-Cheng @jbrockmeyer @a110605

irishgordo:
For-what-it's-worth, it looks like Rancher's interoperability tests are using Harvester CR Operations / CRDs.

Take how Rancher is testing the importing of Harvester:
- https://github.com/rancher/tests/blob/32a475483166bb0b0f0629d7255a537952dffd88/interoperability/harvester/harvester.go#L6C15-L6C78 , which is targeting:  https://github.com/harvester/harvester/tree/master/pkg/apis/harvesterhci.io/v1beta1

That is juggling to modify the Harvester setting:
- https://github.com/harvester/harvester/blob/master/pkg/apis/harvesterhci.io/v1beta1/settings.go

So that the cluster registration url is set with exponential backoff included as helper-timeout-esque-methodologies and that Harvester can be imported into Rancher:
- https://github.com/rancher/tests/blob/32a475483166bb0b0f0629d7255a537952dffd88/interoperability/harvester/harvester.go#L97-L132

also using "Steve" (their big built up client... I think "Norman" is another one leveraged as well) for a helper function along the way:
- https://github.com/rancher/shepherd/blob/050d62ead063a3bc44cb0712852415bc7ede7ffd/clients/rancher/v1/client.go#L426-L446

I think with [Rancher's Sheppard](https://github.com/rancher/shepherd/) and their [interoperability Harvester tests that they currently have and run](https://github.com/rancher/tests/blob/main/interoperability/harvester/harvester.go) they tend to juggle CR Operations / CRDs versus API calls directly.

harvester_robot_tests/tests/negative/vm_node_failure.robot
```diff
@@ -0,0 +1,46 @@
+*** Settings ***
```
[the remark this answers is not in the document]
asc5543: If we would like to point out specific kind of test (like `negative` or `p0`), how about use tag directly, and leave the folder name is a test suite?
On the tests report, it can shows by Tag, and it can also run tests with specific tag.
image

harvester_robot_tests/keywords/host.resource
```diff
@@ -0,0 +1,125 @@
...
+
+Get healthy node names
+    [Documentation]    Get list of all healthy (Ready) nodes
+    ${node_names}=    get_healthy_nodes
+    RETURN    ${node_names}
+
+Cleanup test resources and restore nodes
+    [Documentation]    Clean up resources and ensure all nodes are restored
+    Cleanup test resources
+    Restore all nodes
```
asc5543: Does this line should add `cleanup_hosts` here?

harvester_robot_tests/keywords/host.resource
```diff
@@ -0,0 +1,125 @@
...
+Library          ../libs/keywords/host_keywords.py
+
+*** Keywords ***
+Get node ${node_id}
+    ${node_name}=    get_node_by_index    ${node_id}
+    RETURN    ${node_name}
+
+Cluster has at least ${count} nodes
+    [Documentation]    Verify cluster has minimum number of nodes
+    ${node_count}=    get_node_count
```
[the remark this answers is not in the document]
khushboo-rancher: I've implemented it now. As I was not using host in my test, I didn't add all the methods.

harvester_robot_tests/keywords/host.resource
khushboo-rancher: I'm not sure, what does `cleanup_hosts` mean here. We want to restore the nodes to their initial state, not delete them.

harvester_robot_tests/tests/regression/test_vm.robot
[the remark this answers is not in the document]
lanfon72: I am focusing on the framework perspective, in this test case, we didn't demonstrate how to setup and teardown the prerequisite of VM, as I described in previous comment, the topic is test VM life cycle, create and wait image be download should not be included in, they are the prerequisite of creating a VM, and should be tested in another test case.
In pytest, we using fixtures to setup/teardown prerequisites, but i didn't see any demonstration in this PR.

Vicente-Cheng:
Hi folks, please wait for me before merging.
I would like to check this PR as well.

harvester_robot_tests/tests/negative/vm_node_failure.robot
TachunLin: Tried to trigger the negative tests via robot command
  ```
  robot tests/negative/vm_node_failure.robo
  ```
* It failed with message like keyword not found

  ```
  est VM HA During Node Power Off :: Verify VM HA when node loses p... | FAIL |
  No keyword with name 'And Attach VM To Node' found.
  
  Also teardown failed:
  No keyword with name 'Restore all nodes' found.
  ------------------------------------------------------------------------------
  Vm Node Failure :: VM Node Failure Negative Test Cases                | FAIL |
  2 tests, 0 passed, 2 failed
  
  ```

  image

I tried to find these keyword like `And Attach VM To Node` but not exists. 
Not sure they need to be implemented later

harvester_robot_tests/scripts/node-management/power_off.sh
```diff
@@ -0,0 +1,17 @@
...
+# Power off node script
+# Usage: power_off.sh &lt;hostname&gt; &lt;host_ip&gt;
+HOSTNAME=$1
+HOST_IP=$2
+echo "Powering off node: $HOSTNAME ($HOST_IP)"
+# AWS example
+if [ "$HOST_PROVIDER" == "aws" ]; then
+INSTANCE_ID=$(cat /tmp/instance_mapping | jq -r ".\"$HOSTNAME\"")
+aws ec2 stop-instances --instance-ids $INSTANCE_ID
+fi
```
irishgordo: nit (non-blocking):
Harvester tends to be incompatible with every feasibly deploying on AWS EC2 instances.
The Harvester Cloud project attempted to experiment with AWS but there are too many issues with AWS Harvester Nodes (https://www.github.com/rancher/harvester-cloud/tree/main).
Would it be possible to just remove the entire block?
It also reduces the script's dependencies on external libraries x2 (aws command line tool & jq).
In the current similar script, AWS is not an option, https://github.com/harvester/tests/blob/main/scripts/vagrant/power_off.sh.example

PR #2271, opened 2025-11-15
Add test case for harvester-vm-dhcp-controller addon

khushboo-rancher:

Issue # https://github.com/harvester/tests/issues/2290

This PR adds a test case for below scenario:
1. Enable harvester-vm-dhcp-controller  addon
2. Create VM to verify the vm gets IP from dhcp

This test will fail on Harvester 1.6 and 1.7 until [the fix](https://github.com/harvester/vm-dhcp-controller/pull/74) is in.


noahgildersleeve (approved):
LGTM

noahgildersleeve:
This is using kubectl to create the yaml for the addon. I tested this locally with my Harvester KUBECONFIG loaded into the variable in my python env. It worked fine for the first two tests and skipped the others for the ip pool stuff since it wasn't added in my config.

PR #2372, opened 2026-01-02
Show extra info for Skipped (include XFailed) cases

albinsun:
Issue #2373

1. Easier check test report to know corresponding issue/reason of skipped cases, e.g.
   * Known issues (with URL)
   * Skipped byVersion
   * Skipped byPriorCase (dependency)
3. Refactor to unify `pytest.mark.skip_version_*` marks to the more explicit `pytest.mark.skip_version_if`.
4. For known issues, first use `skip_version_if()` if the milestone is decided. For those still in `Planning`, may use `xfail` (expected fail) temporarily.

 
For known issue to leverage this enhancement, skip `reason` should contain URL.

Verification (`harvester-install-and-test#9`)
image

### DIFF
--- a/harvester_e2e_tests/fixtures/api_client.py
+++ b/harvester_e2e_tests/fixtures/api_client.py
@@ -191,54 +191,37 @@ def expected_settings():
 
 
 @pytest.fixture(autouse=True)
-def skip_version_before(request, api_client):
-    mark = request.node.get_closest_marker("skip_version_before")
-    if mark:
-        cluster_ver = api_client.cluster_version
-        for target_ver in mark.args:
-            if '-head' not in cluster_ver.public and parse_version(target_ver) > cluster_ver:
-                return pytest.skip(
-                    f"Cluster Version `{api_client.cluster_version}` is not included"
-                    f" in the supported version (most >= `{target_ver}`)"
-                )
-
-
-@pytest.fixture(autouse=True)
-def skip_version_after(request, api_client):
-    mark = request.node.get_closest_marker("skip_version_after")
-    if mark:
-        cluster_ver = api_client.cluster_version
-        for target_ver in mark.args:
-            if not hasattr(cluster_ver, 'major') or parse_version(target_ver) <= cluster_ver:
-                return pytest.skip(
-                    f"Cluster Version `{api_client.cluster_version}` is not included"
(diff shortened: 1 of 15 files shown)


jbrockmeyer (approved):

harvester_e2e_tests/fixtures/api_client.py
```diff
@@ -190,55 +190,38 @@ def expected_settings():
...
         reason = mark.kwargs.get('reason', default_reason)
-        if mark.kwargs.get('condition', all)(r for *_, r in checks):
+        cond = mark.kwargs.get('condition', all)
+        if cond(r for *_, r in checks):
             versions = [f"{op} {v}" for op, v, _ in checks]
             return pytest.skip(
-                reason.format(cluster_version=cluster_ver, versions=versions)
+                reason.format(
+                    cluster_version=cluster_ver,
+                    condition=cond_str.get(cond, str(cond)),
```
lanfon72: We didn’t restrict the condition parameter to only any or all; instead, we allowed any callable function. As a result, it may fail if a user provides other callable functions.

To prevent this, we can simply use `callable.__name__` to retrieve the condition name, which also eliminates the need to maintain a mapping table.

harvester_e2e_tests/integrations/test_1_images.py
```diff
@@ -877,6 +879,8 @@ def test_image_complete_lifecycle(self, api_client, image_info, unique_name, wai
         delete_image(api_client, original_image, wait_timeout)
         delete_image(api_client, exported_image_id, wait_timeout)
 
+    @pytest.mark.skip_version_if(
+            ">= v1.7.0", "< v1.9.0", reason="https://github.com/harvester/harvester/issues/9515")
```
lanfon72: #9515 has been rescheduled to v1.9.0.
I would suggest to add `regression` in reason and not limit the **before** version, then when then issue be fixed, we can use the keyword to search and update the skip restriction.

And it would be helpful that add the test case path when filing the regression issue. (cc @asc5543)

albinsun: I plan to implement a xfail_version_if() in later refactoring PR, thx.
Thus the case status will become XPASS when it's fixed and we can update to skip restriction.

Issue #2436, opened 2026-01-27
[TEST] test_9_addons.py::TestVMDHCPControllerAddon::test_verify_vm_dhcp_controller fails due to DHCP IP assignment failed but VM has IP on default VM Status interfaces


asc5543:
## What's the test to develop? Please describe
test_9_addons.py::TestVMDHCPControllerAddon::test_verify_vm_dhcp_controller fails due to DHCP IP assignment failed but VM has IP on default VM Status interfaces

```
E                       AssertionError: DHCP IP assignment failed but VM has IP on default VM Status interfaces: [{'infoSource': 'domain', 'ipAddress': '10.52.2.137', 'ipAddresses': ['10.52.2.137'], 'linkState': 'up', 'mac': 'b2:b9:87:21:be:38', 'name': 'default', 'podInterfaceName': 'eth0', 'queueCount': 1}, {'infoSource': 'domain, multus-status', 'linkState': 'up', 'mac': '56:47:41:d4:9d:31', 'name': 'dhcp-net', 'podInterfaceName': 'podae5df0201be', 'queueCount': 1}]
```

<img width="2554" height="1318" alt="Image" src="https://github.com/user-attachments/assets/c5182c06-375b-4a6e-9d40-3c4819ba597c" />

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] ` test_9_addons.py::TestVMDHCPControllerAddon::test_verify_vm_dhcp_controller` PASS

PR #2443, opened 2026-01-29
Fix managed-dhcp addon tests

khushboo-rancher:

Issue # https://github.com/harvester/tests/issues/2436

1. Untagged network is used.
2. Removed the management network from vm.
3. Added the range of IP in config.

### DIFF
--- a/config.yml
+++ b/config.yml
@@ -11,8 +11,8 @@ vlan-id: 1
 # Physical NIC for VLAN. Default is "mgmt"
 vlan-nic: 'mgmt'
 ip-pool-subnet: '192.168.0.0/24'
-ip-pool-start: ''
-ip-pool-end: ''
+ip-pool-start: '192.168.0.200'
+ip-pool-end: '192.168.0.210'
 
 # Wait time for polling operations
 wait-timeout: 600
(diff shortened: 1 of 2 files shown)


config.yml
```diff
@@ -11,8 +11,8 @@ vlan-id: 1
 # Physical NIC for VLAN. Default is "mgmt"
 vlan-nic: 'mgmt'
 ip-pool-subnet: '192.168.0.0/24'
-ip-pool-start: ''
-ip-pool-end: ''
+ip-pool-start: '192.168.0.81'
```
irishgordo: will approve but everyone must be mindful that this conflicts with our default subnet for ipxe examples.

as in 192.168.0.131 for instance is vip, and does fall into this subnet.

when you have a subnet with two possible dhcp handlers it can be bad news 

we should think maybe in the future to shift this to be more ipxe-examples compatible

irishgordo (dismissed):
lgtm, ipxe-examples compatibility issues but not a big deal, user can always change settings, they just need to be mindful of things -> this will work however on our regular baremetal loadouts without issue

config.yml
asc5543: Agree with Mike, in ipxe-example, the dhcp_server has the range from 192.168.0.50 ~ 192.168.0.130.
Maybe we can use like 192.168.0.200 ~ 192.168.0.210 as the range here

asc5543 (dismissed):

PR #2455, opened 2026-02-04
Use MAC as dhcp-identifier for restored VMs

albinsun:
Apply network data on restored VM in upgrade suite thus VMs can get IP via DHCP without problem.
Ref. https://docs.harvesterhci.io/v1.7/vm/backup-restore#create-a-vm-backup

`test_preq_setup_vms` may fail due to restored VM can not get IP
image
image

#### Verification
image
image


asc5543 (dismissed):
LGTM

albinsun:
Check first NIC name in Ubuntu and OpenSUSE
image

Verify configuration is correct
image
image

Issue #2438, opened 2026-01-28
[TEST] test_4_vm_host_cpu_pinning.py::TestPinCPUonVM::test_disable_cpu_manager_when_vm_on_it::teardown got error due to there is volume use the image


asc5543:
## What's the test to develop? Please describe
test_4_vm_host_cpu_pinning.py::TestPinCPUonVM::test_disable_cpu_manager_when_vm_on_it::teardown got error due to there is volume use the image

```
E       AssertionError: (422, {'apiVersion': 'v1', 'code': 422, 'details': {'causes': [{'message': 'Cannot delete image default/ubuntu-16h00m0...used by volume default/pin-cpu-16h00m01s298469-01-28-disk-0', 'reason': 'FieldValueInvalid'}]}, 'kind': 'Status', ...})
E       assert 200 == 422
```

<img width="2558" height="1321" alt="Image" src="https://github.com/user-attachments/assets/c3d8aae7-f7c9-40dc-b48c-8159b2d2e2d9" />

## Prerequisite and dependency of test

Do teardown before attempting to remove the image

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `api_client.images.delete` success without assert

## Additional context

Add any other context or screenshots about the test request here.

PR #2471, opened 2026-02-11
Add retry to avoid race condition

asc5543:
Without retry to check volume is deleted, script will attempt to delete image directly, if volume still exists, it shows error

Issue #2438

To avoid race condition cause test_4_vm_host_cpu_pinning.py::TestPinCPUonVM::test_disable_cpu_manager_when_vm_on_it::teardown got error


noahgildersleeve (dismissed):
LGTM.

irishgordo (approved):
lgtm

albinsun (approved):

PR #2527, opened 2026-03-21
Incorporate image name storgae class change

khushboo-rancher:
#### Which issue(s) this PR fixes:

Issue #
https://github.com/harvester/harvester/issues/10227

#### What this PR does / why we need it:
Incorporate the image storgae name change - https://github.com/harvester/harvester/pull/9392

PR #2530, opened 2026-03-23
Backward compatibilty for image_uid

khushboo-rancher:
With https://github.com/harvester/tests/pull/2527 change, add the compatibility with <1.8.0.


noahgildersleeve (approved):
LGTM

PR #2531, opened 2026-03-24
Add volume snapshot test with vm part

asc5543:

Issue #472

Add volume snapshot test with vm part.

Test Pass in `v1.8.0-rc5` in the lastest changes
image

### DIFF
--- a/.github/workflows/gh-pages-pr-workflow.yml
+++ b/.github/workflows/gh-pages-pr-workflow.yml
@@ -37,7 +37,8 @@ jobs:
         working-directory: ./docs/
       - name: install requirements
         run: |
-          pip install -r test-requirements.txt
+          pip install --no-deps -e${{github.workspace}}/apiclient
+          pip install -r test-requirements.txt --require-hashes
       - name: Run pdoc3
         run: |
           pdoc --html -o . .
--- a/harvester_e2e_tests/integrations/test_1_volumes.py
+++ b/harvester_e2e_tests/integrations/test_1_volumes.py
@@ -668,16 +668,15 @@ def test_create_volume_snapshot_and_restore(
     7. Delete source volume, the snapshot should be deleted
     Ref. https://github.com/harvester/harvester/issues/2294
     """
-    image_id = None
     if source_type == "VM Image":
-        image_id = ubuntu_image['id']
-        unique_name = f"{unique_name}-image"
+        spec = api_client.volumes.Spec.for_image(ubuntu_image['info'])
+    else:
+        spec = api_client.volumes.Spec("10Gi")
 
     # Create a volume from source_type(Empty or VM Image)
-    spec = api_client.volumes.Spec("10Gi", storage_cls=None)
     kws = dict()
-    code, data = api_client.volumes.create(unique_name, spec, image_id=image_id, **kws)
-    assert 201 == code, (code, unique_name, data, image_id)
+    code, data = api_client.volumes.create(unique_name, spec, **kws)
+    assert 201 == code, (code, unique_name, data, spec.to_dict(unique_name, 'default'))
 
     # Wait the volume become ready
     volumes_ready, (code, data) = volume_checker.wait_volumes_ready([unique_name])
--- a/harvester_e2e_tests/integrations/test_4_vm_snapshot.py
+++ b/harvester_e2e_tests/integrations/test_4_vm_snapshot.py
@@ -597,3 +597,101 @@ class TestVMSnapshot:
             sleep(1)
         else:
             raise AssertionError(f"timed out waiting for {volumesnapshotname} to be deleted")
+
+
+@pytest.mark.p0
+@pytest.mark.virtualmachines
+@pytest.mark.volumes
+class TestVolumeSnapshotWithVM:
+    def test_snapshot_of_volume_on_vm(self, api_client,
+                                      source_vm, vm_checker,
+                                      volume_checker, wait_timeout,
+                                      host_shell, vm_shell,
+                                      ssh_keypair, polling_for):
+        """
+        1. Create a VM with volume
+        2. Write some data on the VM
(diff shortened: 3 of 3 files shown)


harvester_e2e_tests/integrations/test_1_volumes.py
```diff
@@ -623,6 +671,95 @@ def test_delete_volume_on_existing_vm(self, api_client, ubuntu_image, ubuntu_vm,
         # source
         assert ubuntu_image["id"] == annotations['harvesterhci.io/imageId'], (code, data)
 
+    def test_snapshot_of_volume_on_vm(self, api_client, ubuntu_vm, ubuntu_image,
```
lanfon72: I think this test case should move to `test_4_vm_snapshot` as it hardly depends on VM.

lanfon72 (commented):
and it looks we should refactor the class `class TestVolumeWithVM`.
3 test cases: `pause_vm`, `stop_vm` and `delete_vm` are duplicated in `test_3_vm_functions`, it looks like we are not testing additional **volumes** on those test cases.

and remaining test cases: `test_delete_volume_on_existing_vm` and `test_delete_volume_on_deleted_vm` should be moved to `test_3_vm_functions` as well.
Maybe we can move them into class `TestVMWithVolumes` in `test_3_vm_functions` or `test_3_vm`?

@albinsun the class was added by you, could you work on this?

harvester_e2e_tests/integrations/test_1_volumes.py
asc5543: Moved test_snapshot_of_volume_on_vm from `test_1_volumes` to `test_4_vm_snapshot`

.github/workflows/gh-pages-pr-workflow.yml
```diff
@@ -39,6 +39,8 @@ jobs:
         run: |
           pip install -r test-requirements.txt
       - name: Run pdoc3
+        env:
+          PYTHONPATH: ${{github.workspace}}/apiclient
         run: |
           pdoc --html -o . .
           mv harvester_e2e_tests/ ../docs/public/backend
```
irishgordo, line 45: We might not need:
```
env:
  PYTHONPATH: ${{}}
```

What happened is due to:
- https://github.com/harvester/tests/commit/8d14115898db511df96e4625aadc2978bd7ae023

We need to install the apiclient separately.
This is due to the hashing.
We can not have a requirements file that is trying to install hashes and then also try to also install our own package that isn't hashed.

@ihcsim had suggested that we do not "hash" our own source package of apiclient.

So what we did in tox.ini was
https://github.com/harvester/tests/blob/main/tox.ini

To just simply have the commands setup like:
```
# we're going against convention here but it offers us more control
# we install everything with the hashes, then content inside apiclient dir (ensuring dependencies come from hashes ie: requests, as to not attempt to fetch them again)
commands =
  ; {toxinidir}/terraform_test_artifacts/terraform_install.sh
  pip install -r{toxinidir}/test-requirements.txt --require-hashes
  pip install --no-deps -e{toxinidir}/apiclient
  find . -type f -name "*.pyc" -delete
  pytest -c {env:PYTEST_CONFIG:{toxinidir}/tox.ini} {posargs}
```

So my suggestion:
```suggestion
Lines changed: 2 additions & 0 deletions


Original file line number	Original file line	Diff line number	Diff line change
        run: |
          pip install --no-deps -e${{github.workspace}}/apiclient
          pip install -r test-requirements.txt --require-hashes
      - name: Run pdoc3
        run: |
          pdoc --html -o . .
          mv harvester_e2e_tests/ ../docs/public/backend
```

irishgordo (changes requested):

.github/workflows/gh-pages-pr-workflow.yml
asc5543: Thanks for the suggestion!
I updated the yaml for this change.

jbrockmeyer (approved):

albinsun (approved):

harvester_e2e_tests/integrations/test_4_vm_snapshot.py
```diff
@@ -597,3 +597,101 @@ def test_volume_snapshots_are_cleaned_up_after_source_volume_deleted(self, api_c
...
+        vol_name = (data["spec"]["template"]["spec"]["volumes"][0]
+                        ['persistentVolumeClaim']['claimName'])
+
+        vm_started, (code, vmi) = vm_checker.wait_started(vm_name)
+        assert vm_started, (code, vmi)
+
+        # Write 123 into test.txt
+        def action(sh):
+            _, _ = sh.exec_command("echo 123 > test.txt")  # nosec B601
+            _, _ = sh.exec_command("sync")  # nosec B601
```
irishgordo, line 633: nonblocking: this is great 😄  - perhaps in a future enhancement we can collect & audit an `md5sum` of it, to validate that when it's restored the data's file integrity is the same.

harvester_e2e_tests/integrations/test_4_vm_snapshot.py
```diff
@@ -597,3 +597,101 @@ def test_volume_snapshots_are_cleaned_up_after_source_volume_deleted(self, api_c
...
+        code, data = polling_for(
+            "VM do created",
+            lambda c, d: 200 == c and d.get('status', {}).get('printableStatus') == "Running",
+            api_client.vms.get, vm_with_restore_vol_name
+        )
+
+        # Check the data in the new VM
+        def actassert(sh):
+            out, _ = sh.exec_command("cat test.txt")  # nosec B601
+            assert "123" in out
```
irishgordo, line 686: same as: https://github.com/harvester/tests/pull/2531/changes#r3081079816

irishgordo (approved):
overall lgtm - thanks for the test addition

lanfon72 (approved):

PR #2546, opened 2026-03-31
[apiclient] update logic of cluster_version

lanfon72:

Issue #

Engineer team have new logic for cluster version, we need to handle those.
(discussion thread in slack.)

1. Move parse logic to `_normalize_version`
2. Change to use regex for most case
3. store SHA info as local version as `<VersionInstance>.local`
4. add unittest for `_normalize_version`
5. update legacy unttest cases of `test_api.py`

1. For *invalid* version, add new version item in `test_customized_build_version` and run unittest to make sure it will pass.
2. For *valid* version, add new version item to relevant test case and run unittest to make sure it will pass.

Run the command like `python -m unittest apiclient/tests/test_api.py`

image

### DIFF
--- a/apiclient/harvester_api/api.py
+++ b/apiclient/harvester_api/api.py
@@ -1,11 +1,23 @@
 from urllib.parse import urljoin
 
+import re
 import requests
 from pkg_resources import parse_version
 from requests.packages.urllib3.util.retry import Retry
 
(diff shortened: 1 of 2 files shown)


wheatdog (approved):
Thanks!

Issue #2576, opened 2026-04-15
[TEST] test_1_volumes.py::TestVolumeWithVM fails in VM creation


asc5543:
## What's the test to develop? Please describe
`harvester_e2e_tests/integrations/test_1_volumes.py::TestVolumeWithVM::test_delete_volume_on_existing_vm::setup`
`harvester_e2e_tests/integrations/test_1_volumes.py::TestVolumeWithVM::test_delete_volume_on_deleted_vm::setup`
<img width="1785" height="530" alt="Image" src="https://github.com/user-attachments/assets/30650739-7111-46f7-81c5-53d1d79e9efe" />

Will follow https://github.com/harvester/tests/pull/2538 to fix this

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `VM creation should succeed`

## Additional context

Add any other context or screenshots about the test request here.

asc5543:
Fix in #2577 
<img width="1898" height="612" alt="Image" src="https://github.com/user-attachments/assets/691a3b5d-605d-47d8-a53f-f18be38246c3" />

PR #2577, opened 2026-04-15
Update image_uid to fix VM's image creation

asc5543:

Issue #2576 

To fix VM creation fail

image


albinsun (approved):
LGTM

Issue #2574, opened 2026-04-14
[TEST] test_3_vm_functions.py::TestVMOperations::test_abort_migrate assert 204 == 500 when run all TestVMOperations


asc5543:
## What's the test to develop? Please describe

test_3_vm_functions.py::TestVMOperations::test_abort_migrate assert 204 == 500 when run all TestVMOperations 

<img width="1901" height="965" alt="Image" src="https://github.com/user-attachments/assets/b037b10b-b8bb-4bc5-b7b9-585bbe11013e" />

But it pass if run the dependcy and test_abort_migrate only
<img width="1901" height="965" alt="Image" src="https://github.com/user-attachments/assets/b6447da5-84f5-45b0-8599-708405e52682" />

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `test_abort_migrate` pass

## Additional context

Add any other context or screenshots about the test request here.

asc5543:
This issue happend due to `test_abort_migrate` run directly right after `test_migrate`, hit the race condition.
Add sleep after the `test_migrate` to avoid the race condition.

asc5543:
The issue still happens sometime, need more investigate to this.
<img width="1177" height="600" alt="Image" src="https://github.com/user-attachments/assets/467f6caa-c608-4d93-87e5-0f628073a055" />
<img width="1631" height="984" alt="Image" src="https://github.com/user-attachments/assets/9cd41022-eb35-489b-b421-5956e14beeee" />

asc5543:
Add to check the `state` of the `relationships`.
Even the migrate success, the state of source pod may still in `unavailable`, which will cause the `test_abort_migrate` fail due to multiple active pods.

```
        "relationships": [
            {
                "fromId": "default/vm-10h40m50s850340-04-21",
                "fromType": "kubevirt.io.virtualmachine",
                "rel": "owner",
                "state": "active",
                "message": "Resource is Ready",
            },
            {
                "toId": "default/virt-launcher-vm-10h40m50s850340-04-21-jnrtn",
                "toType": "pod",
                "rel": "owner",
                "state": "unavailable",
                "message": "containers with unready status: [compute]",
                "transitioning": True,
            },
            {
[...]

PR #2582, opened 2026-04-16
Fix test_migrate race condition

asc5543:

Issue #2574 

To avoid the race condition cause `test_abort_migrate` fail
Current solution is to check the latest `ctx.data['status']['conditions']` is `AgentConnected` or not.

image

### DIFF
--- a/harvester_e2e_tests/fixtures/virtualmachines.py
+++ b/harvester_e2e_tests/fixtures/virtualmachines.py
@@ -332,6 +332,7 @@ def vm_checker(api_client, wait_timeout, sleep_timeout, vm_shell):
             return True, (ctx.stdout, ctx.stderr)
 
         def wait_migrated(self, vm_name, new_host, endtime=None, callback=default_cb, **kws):
+            # ref: https://github.com/harvester/tests/pull/2582
             ctx = ResponseContext('vm.migrate', *self.vms.migrate(vm_name, new_host, **kws))
             if 404 == ctx.code and callback(ctx):
                 return False, ctx
@@ -342,6 +343,7 @@ def vm_checker(api_client, wait_timeout, sleep_timeout, vm_shell):
                 if (
                     not ctx.data['metadata']['annotations'].get("harvesterhci.io/migrationState")
                     and new_host == ctx.data['status']['nodeName']
+                    and "AgentConnected" == ctx.data['status']['conditions'][-1].get('type')
                     and callback(ctx)
                 ):
                     break
--- a/harvester_e2e_tests/integrations/test_3_vm_functions.py
+++ b/harvester_e2e_tests/integrations/test_3_vm_functions.py
@@ -468,7 +468,7 @@ class TestVMOperations:
             f"Last API Status({code}): {data}"
         )
 
-    def test_migrate(self, api_client, unique_vm_name, wait_timeout):
+    def test_migrate(self, api_client, unique_vm_name, vm_checker):
         """
         To cover test:
         - https://harvester.github.io/tests/manual/live-migration/migrate-turned-on-vm-to-another-host/ # noqa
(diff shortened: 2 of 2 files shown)


harvester_e2e_tests/integrations/test_3_vm_functions.py
```diff
@@ -493,21 +493,13 @@ def test_migrate(self, api_client, unique_vm_name, wait_timeout):
...
+        assert vm_migrated, (
+            f"Failed to Migrate VM({unique_vm_name}) from {cur_host} to {new_host}\n"
+            f"API Status({code}): {data}"
+        )
 
-        endtime = datetime.now() + timedelta(seconds=wait_timeout)
-        while endtime > datetime.now():
-            code, data = api_client.vms.get_status(unique_vm_name)
-            migrating = data['metadata']['annotations'].get("harvesterhci.io/migrationState")
-            if not migrating and new_host == data['status']['nodeName']:
```
lanfon72, line 503: I think it is not necessary to change to use `vm_checker`.  (they doing almost the same thing, this test case is implemented before the fixture)

And I reproduced this case, to make sure the migration already done, we can also check the `AgentConnected` be appended in `conditions`.

In short, we can just update:
```python3
# before
if not migrating and new_host == data['status']['nodeName']:
    ....
# after
if (
    not migrating
    and new_host == data['status']['nodeName']
    and "AgentConnected" == data['status']['conditions'][-1].get('type')
):
    ....
```

it would be very appreciated if you can also update the `vm_checker.wait_migrated`
```python
if (
    ...
    and "AgentConnected" == ctx.data['status']['conditions'][-1].get('type')
    and callback(ctx)
):
    ...
```

asc5543: Thanks for the point!
I changed to use `vm_checker` since with this change, we can just change `vm_checker.wait_migrated` in the future.

harvester_e2e_tests/fixtures/virtualmachines.py
```diff
@@ -342,6 +342,7 @@ def wait_migrated(self, vm_name, new_host, endtime=None, callback=default_cb, **
                 if (
                     not ctx.data['metadata']['annotations'].get("harvesterhci.io/migrationState")
                     and new_host == ctx.data['status']['nodeName']
+                    and "AgentConnected" == ctx.data['status']['conditions'][-1].get('type')
```
albinsun, line 346: This checks `ctx.data['conditions']`'s last item is `AgentConnected` type, e.g. 
```
[
  {"type": "type1", ...},
  {"type": "type2", ...},
  {"type": "AgentConnected"} 
]
```
However what if there is another event generated closely. Will this cuz problem?
```
[
  {"type": "type1", ...},
  {"type": "type2", ...},
  {"type": "AgentConnected"}
  {"type": "OtherEvent"}
]
```
If so, we may consider to use the `contains` logic, e.g.
```
>>> 'AgentConnected' in [c["type"] for c in conds]
True
```
WDYT?

asc5543: I agree with that, since the VM is created during the testing, it is okay to check all `type` in the history.
One concern is that if there are test cases do migrate in the same VM twice, it maybe a problem.
@lanfon72 WDYT?

lanfon72: Although Kubevirt didn't guarantee the order of agent condition, but ideally the condition should be the last because the agent connect to new pod will happen after the migration done.

We check whether it is the last condition would also help us to find whether there would have other events intercept/interact with the VM during migration state. (We won't know the event affecting migration or not, because it's not happening yet)

asc5543: Understood, I also ran the migration and checked the `type`, `AgentConnected` is always the latest one.
If there is another event generated closely, maybe we have to check why it happens since it is not happening before.

In conclusion, I prefer to keep checking the latest `type`, and will investigate if there is another event generated closely in the future.

lanfon72: You can also leave comment to ref the PR in the code as a reference for future changes.

albinsun: LGTM

albinsun (dismissed):

asc5543:
Add the comment for future reference.

albinsun (approved):

PR #2649, opened 2026-05-15
Add cpu and mem hotplug test cases

asc5543:

Issue #1695 
1. Optimize the variable and order of VMSpec
2. Add VMSpec160 for cpu/mem hotplug functions
3. Add basic test cases with cpu/mem hotplug

To support cpu and mem hotplug test in automation

The test result of `v1.5.2`, won't affect other test cases and the test cases are skipped as expected
Image
Image

And the test result of `v1.8.1-dev`
Image

### DIFF
--- a/harvester_e2e_tests/integrations/test_3_vm_functions.py
+++ b/harvester_e2e_tests/integrations/test_3_vm_functions.py
@@ -2217,3 +2217,114 @@ class TestHotPlugVolume:
 
         assert not scsi, "SCSI device still available in `/sys/block/`"
         assert not out, "SCSI device still available in `lsblk -r`"
+
+
+@pytest.mark.p0
+@pytest.mark.sanity
+@pytest.mark.virtualmachines
+@pytest.mark.skip_if_version(
+    "< v1.6.0",
+    reason=("feature available as of v1.6.0, ref: "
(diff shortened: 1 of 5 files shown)


harvester_e2e_tests/integrations/test_3_vm_functions.py
```diff
@@ -2217,3 +2217,112 @@ def test_remove(
...
+            name="cpu_mem_hotplug_delete_vm", depends=["cpu_mem_hotplug_create_vm"])
+    def test_delete(
+        self, api_client, unique_vm_name, vm_checker
+    ):
+        _, data = api_client.vms.get(unique_vm_name)
+        vm_spec = api_client.vms.Spec.from_dict(data)
+        _ = vm_checker.wait_deleted(unique_vm_name)
+        for vol in vm_spec.volumes:
+            vol_name = vol['volume']['persistentVolumeClaim']['claimName']
+            api_client.volumes.delete(vol_name)
```
albinsun, line 2330: I think it's better to `assert vm_deleted` if we make it a test case instead of a general teardown.

asc5543: Agree, update the assert vm_deleted

albinsun (approved):
lgtm

Issue #2654, opened 2026-05-18
[TEST] test_support_bundle.py::TestSupportBundleCustomFileName::test_create is run unexpected in v1.7.2-rc1


asc5543:
## What's the test to develop? Please describe
test_support_bundle.py::TestSupportBundleCustomFileName::test_create is run unexpected in `v1.7.2-rc1`
The test class should run when version >= `v1.8.0`.

```
@pytest.mark.skip_if_version(
    "< v1.8.0", reason="https://github.com/harvester/harvester/issues/9257 new feature in v1.8.0")
@pytest.mark.usefixtures("custom_filename_setting")
class TestSupportBundleCustomFileName:
```

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `TestSupportBundleCustomFileName` shouldn't run in `v1.7.2-rc1` test

## Additional context

Add any other context or screenshots about the test request here.

asc5543:
This issue is fixed.
<img width="1312" height="967" alt="Image" src="https://github.com/user-attachments/assets/d147f9e8-b4b5-4876-b7f3-3e67b025a62b" />

PR #2658, opened 2026-05-19
Add version check in custom_filename_setting

asc5543:

Issue #2654

Since automation call `custom_filename_setting` before `skip_if_version`, so it will shows error if the version doesn't support custom file setting.

Pass in `v1.7.2-rc1`, `v1.6.1` and `v1.8.0`
image
image
image


albinsun (approved):
lgtm

Issue #2653, opened 2026-05-18
[TEST] test_1_volumes.py::test_create_volume_bad_checksum fail due to TypeError: VolumeSpec.to_dict() got an unexpected keyword argument 'image_uid' in v1.7.2-rc1


asc5543:
## What's the test to develop? Please describe
test_1_volumes.py::test_create_volume_bad_checksum fail due to TypeError: VolumeSpec.to_dict() got an unexpected keyword argument 'image_uid' in `v1.7.2-rc1`
<img width="1911" height="990" alt="Image" src="https://github.com/user-attachments/assets/03c74791-764f-41fc-bc6a-f4ace2857724" />

## Prerequisite and dependency of test
Since the `image_uid` is started in `v1.8.0-rc1`, there shouldn't have this kind issue in `v1.7.2-rc1`

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `test_1_volumes.py::test_create_volume_bad_checksum` should be passed no matter the version

## Additional context

Add any other context or screenshots about the test request here.

asc5543:
The issue is fixed.

<img width="1312" height="967" alt="Image" src="https://github.com/user-attachments/assets/ec17e140-5123-45cc-a110-4f6628be99cd" />

PR #2659, opened 2026-05-19
Add image_uid in VolumeSpec for lower version

asc5543:

Issue #2653

Since in test_1_volumes, we send `image_uid=ubuntu_image_bad_checksum['info']['metadata']['uid']` into `VolumeSpec`
But there is no `image_uid` in `to_dict` of `VolumeSpec`
Add `image_uid` in `VolumeSpec` for supporting lower version

Pass in `v1.7.2-rc1`, `v1.6.1`, and `v1.8.0`
image
image
image


albinsun (approved):
lgtm

PR #2661, opened 2026-05-20
Update xfail range for those fixed in v1.7.1

asc5543:

No issue ID, just update the xfail range for those issues were fixed in v1.7.1, to let those test cases can be run in those fixed version.

Since in mark `xfail_if_version`, it will show xfail directly if the tested harvester version in the range when the setup step.
In this case, it won't run the test case, so with this PR, the test cases can be run and we can check the test result in those fixed version.


albinsun (approved):
lgtm

PR #2696, opened 2026-06-09
* [e2e] temporarily skip for upgrade test case

lanfon72:
refer to issue https://github.com/harvester/harvester/issues/10598

It's a temporarily way to skip this test case, I will figure out a better way to handle similar cases.


albinsun (approved):
lgtm

PR #2703, opened 2026-06-10
[e2e] fix VM teardown issue in upgrade

lanfon72:
Issue #

1. **ADD** the function `test_teardown_environment`  to clear up environment
2. **UPDATE** cluster/vm network fixture to not do tear down when there have any VM related test failed
3. **REMOVE** version skip decorate on `test_verify_restore_vm_from_snapshot`

To notice that this change will affect the context of `test_verify_restore_vm`

### DIFF
--- a/harvester_e2e_tests/integrations/test_upgrade.py
+++ b/harvester_e2e_tests/integrations/test_upgrade.py
@@ -31,10 +31,9 @@ logger = logging.getLogger(__name__)
 @pytest.fixture(scope="module")
 def cluster_state(request, unique_name, api_client):
     class ClusterState:
-        vm1 = None
-        vm2 = None
-        vm3 = None
-        pass
+        vms = dangling = None
+        version = ""
+        version_verify = False
 
     state = ClusterState()
 
@@ -42,7 +41,6 @@ def cluster_state(request, unique_name, api_client):
         state.version_verify = True
         state.version = request.config.getoption('--upgrade-target-version')
     else:
-        state.version_verify = False
         state.version = f"version-{unique_name}"
 
     return state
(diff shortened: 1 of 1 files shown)


harvester_e2e_tests/integrations/test_upgrade.py
```diff
@@ -1319,3 +1332,19 @@ def test_upgrade_image_deleted(self, api_client, wait_timeout):
...
+          teardown cleanly.
+        '''
+
+        # remove those VMs created before upgrade and verified
+        for name in cluster_state.vms['names']:
+            if name in cluster_state.dangling:
+                code, data = api_client.vms.get(name)
+                if 404 != code:
+                    vm_deleted, (code, data) = vm_checker.wait_deleted(name)
+                    if vm_deleted:
```
albinsun, line 1345: use `assert` instead？

albinsun (commented):
LGTM in general.
Could you please attach the test report to the PR for reference?

harvester_e2e_tests/integrations/test_upgrade.py
lanfon72: I hit the case Consider the case https://github.com/harvester/harvester/issues/10598 event we skip restore from snapshot (but the VM could be deleted when login and use UI's functionality).

So consider the case, we will not do **assertion** in here to let those VMs stay on final state.  (then cluster/VM network tear down will be skip.)

I will test whether we can remove the skip on the snapshot test case. If successful, this change will be ready to merge.

khushboo-rancher (approved):

lanfon72:
Add multi stage upgrade result based on this PR
v1.6.1 -> v1.7.1
image

v1.7.1 -> v1.8.1-rc1
image

cluster status
image

VM network, volumes and images was created by upgrade, we can still observing they still available on other upgrade runs or not, ideally they should be removed if VMs could be deleted gracefully.

Issue #2768, opened 2026-07-15
[BUG] Rancher tests aren't passing on v1.8 or v1.9


noahgildersleeve:
# Summary

The Rancher integration tests stopped passing on v1.8 after yesteday. We ran the tests for v1.9.0-rc1 yesterday and they failed with the cluster-registration-url having the incorrect token, but then the URL was removed from the Harvester cluster. This was after killing the pipeline to make sure that the environment wasn't cleaned up.

The v1.8 tests ran today and they failed to create Rancher with the pipeline. 

# Repro Steps
1. Run the Rancher integration tests pipeline on Jenkins with the info for testing v1.9.0-rc1

# Details

noahgildersleeve:
It seems that the failure on v1.8 was unrelated. @khushboo-rancher found that the failure on v1.9.0-rc1 was being caused by a change in Rancher that the token is being stored in a secret and not in the CRD. There will be a PR that fixes the issue and I will verify it when it's merged.

noahgildersleeve:
The pipeline was updated and fixed.

PR #2769, opened 2026-07-16
Pull token from secret

khushboo-rancher:
Starting from Rancher 2.15, the registration token is not available in plain text in the registration crd. It is moved to a secret in the same namespace


noahgildersleeve (approved):
LGTM.

noahgildersleeve:
This also fixed [2768](https://github.com/harvester/tests/issues/2768)

Issue #2756, opened 2026-07-08
[TEST] test_4_vm_template.py::TestVMTemplate::test_create_template_with_data fails due to AssertionError: (500, 'unexpected end of JSON input')


asc5543:
## What's the test to develop? Please describe
test_4_vm_template.py::TestVMTemplate::test_create_template_with_data fails due to AssertionError: (500, 'unexpected end of JSON input') in `v1.8.2-rc1`
<img width="1673" height="384" alt="Image" src="https://github.com/user-attachments/assets/a8b1477c-3b19-4a35-9989-d32ef28b03ce" />

Follow the test steps in description to run manual test will pass, so it maybe a script issue.

Steps:
    1. Create VM and write some data
    2. Create new template and keep data from the VM
    3. Create new VM from the template
    4. Check data consitency

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `test_4_vm_template.py::TestVMTemplate::test_create_template_with_data` should run success without issue

## Additional context

Add any other context or screenshots about the test request here.

asc5543:
Pass in `v1.6.1` but fail in `v1.7.2` and `v1.8.2-rc1`

<img width="1906" height="628" alt="Image" src="https://github.com/user-attachments/assets/30779e46-95d9-49dd-8972-42089bfdf1bc" />

<img width="1906" height="628" alt="Image" src="https://github.com/user-attachments/assets/22b924dd-404a-4d61-8854-eb71198a7385" />

<img width="1906" height="628" alt="Image" src="https://github.com/user-attachments/assets/81d44ada-57a3-4228-9ed7-c7811cf184cf" />

asc5543:
It seems the userdata has something wrong 
<img width="1606" height="376" alt="Image" src="https://github.com/user-attachments/assets/9a42759e-45ca-475e-b8ed-59c30617f6ed" />

After fixing it manually, the test case can pass
<img width="1606" height="376" alt="Image" src="https://github.com/user-attachments/assets/5a880f64-88a3-432b-9adb-45b790713a64" />

<img width="1591" height="396" alt="Image" src="https://github.com/user-attachments/assets/f729ec37-1869-4b47-8475-6b7f4194aa80" />

[...]

PR #2771, opened 2026-07-17
fix: missing harvesterhci.io/sshNames issue

asc5543:

Issue #2756 

Python script is missing the `harvesterhci.io/sshNames` annotation when creating the VM template, which causes a JSON unmarshalling error (AssertionError: (500, 'unexpected end of JSON input')) at [handler.go#L1501](https://github.com/harvester/harvester/blob/master/pkg/api/vm/handler.go#L1501).
After adding the `harvesterhci.io/sshNames`, the issue is gone.

The test pass.
image
image
image


irishgordo (approved):
thanks for the fix @asc5543 lgtm 😄

PR #2772, opened 2026-07-20
Reenable cases skipped by 10220

albinsun:

Not fix but just re-enable upgrade test cases which skipped by harvester/harvester/issues/10220

As #10220 seems covered by early fix (see https://github.com/harvester/harvester/issues/10220#issuecomment-4987641111), and a quick check on `v1.8` -> `v1.9` passed too, we can re-enable it and observe for more runs.

According to the fixed version, enable them since `v1.8`.

`v1.8.1` -> `v1.9.0-rc1`
image


asc5543 (approved):
lgtm

Issue #2778, opened 2026-07-22
[TEST] test_1_images.py::TestImageEnhancements::test_image_name_length_limits return 404 in automation but it should be 422


asc5543:
## What's the test to develop? Please describe
The test case fail due to the return code is not the expected one.
<img width="1080" height="501" alt="Image" src="https://github.com/user-attachments/assets/9077139c-94ac-4ffb-8195-e24b49f74c5b" />

## Prerequisite and dependency of test

Any prerequisite environment and pre-condition required for this test. 
Provide test case dependency here if any.

## Describe the items of the test development (DoD, definition of done) you'd like
> Please use a task list for items on a separate line with a clickable checkbox https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists

- [ ] `test_1_images.py::TestImageEnhancements::test_image_name_length_limits` return 422

## Additional context
It should return 422 if created via URL
<img width="1905" height="982" alt="Image" src="https://github.com/user-attachments/assets/26df308f-28f5-4794-9e64-a77cd8ce9462" />

If attempt to create via file, it will also shows 422
<img width="1905" height="982" alt="Image" src="https://github.com/user-attachments/assets/c9455024-ffcf-4220-9016-d1825201cdf0" />

asc5543:
The issue was fixed.
<img width="1232" height="970" alt="Image" src="https://github.com/user-attachments/assets/a2966285-23c1-461e-8023-05eb02efa55d" />

PR #2779, opened 2026-07-22
fix: let name length limits test get 422

asc5543:

Issue #2778 

1. In current test case, it use `create_by_file` for invalid long name, which will get 404 since the file is not created due to invalid long name. Use `create_by_url` instead to get 422 as expected
2. Change `xfail` to `xfail_if_version` since the issue is fixed.

image

### DIFF
--- a/harvester_e2e_tests/integrations/test_1_images.py
+++ b/harvester_e2e_tests/integrations/test_1_images.py
@@ -873,8 +873,8 @@ class TestImageEnhancements:
         delete_image(api_client, original_image, wait_timeout)
         delete_image(api_client, exported_image_id, wait_timeout)
 
-    @pytest.mark.xfail(
-            reason="https://github.com/harvester/harvester/issues/9515")
+    @pytest.mark.xfail_if_version(
+            "<= v1.8.0", reason="https://github.com/harvester/harvester/issues/9515")
     @pytest.mark.p1
     @pytest.mark.images
     @pytest.mark.negative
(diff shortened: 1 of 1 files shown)


irishgordo (dismissed):
lgtm

harvester_e2e_tests/integrations/test_1_images.py
```diff
@@ -895,11 +895,13 @@ def test_image_name_length_limits(self, api_client, unique_name, wait_timeout):
             long_base = "very-long-image-name-for-testing-kubernetes-name-length-limits-that-exceeds-maximum"  # NOQA
             invalid_long_name = f"{long_base}-{unique_name}"
 
-            resp = api_client.images.create_by_file(invalid_long_name, Path(f.name))
-            assert not resp.ok, f"Expected failure with long name but" \
-                                f" got success: {resp.status_code}"
-            assert resp.status_code in [400, 422], f"Expected 400/422 for " \
-                                                   f"invalid name, got: {resp.status_code}"
+            code, data = api_client.images.create_by_url(
```
lanfon72: If we change to use `create_by_url`, this test case would be different to existing one. The context was "upload file with long file name".
IMO, we should first check whether the API be changed to return 404 than update the test case.

asc5543: I checked the API return 422 on https://github.com/harvester/tests/issues/2778.
And the 404 here because in create_by_file, we trigger self.create first, which return 422 if the file name is too long.
After that, since the file doesn't exist, the function will return 404.
Otherwise, in normal use case, since user can only edit in yaml when use create_by_url, and both create_by_url and create_by_file are in the flow of create by yaml, I think it should be the same.

lanfon72: The change introduced in https://github.com/harvester/harvester/pull/10490, so in older version, it will return `[400, 422]` as expected.

After the fix introduced, empty file will not be created anymore, so we can expect it will return `404`. (which means we can just simply add `404` into the expected return code and ref the fix.)

To notice, I'm not suggest to change the test context like current PR because we did tested it in older version, and there have 2 sub tests, both depends on **create_by_file**.
we should keep the same test scenario when modify existing test case.

lanfon72 (approved):

PR #2811, opened 2026-08-06
ci: add containerized smoke test runner

Vicente-Cheng:
Add smoke-run.sh and Dockerfile.smoke to run the pytest smoke marker
    inside a container. Dependencies are baked into the image, so a test
    run itself needs no network and repeated runs hit the docker layer
    cache. Used by the harvester/release smoke workflows.

    Related issue: harvester/harvester#10211

Issue #

Used for the harvester release smoke test

### DIFF
--- a/Dockerfile.smoke
+++ b/Dockerfile.smoke
@@ -1,0 +1,15 @@
+FROM FROM python:3.10.16-bookworm@sha256:b240c5813c2fa23506dc314bb370da0e890dadce666c7886780757976dc3b54b
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends build-essential findutils wget && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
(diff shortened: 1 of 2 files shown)


Dockerfile.smoke
```diff
@@ -0,0 +1,15 @@
+FROM python:3.10.16-bookworm
```
khushboo-rancher: It will be nice to pin the python version:
`FROM python:3.10.16-bookworm@sha256:b240c5813c2fa23506dc314bb370da0e890dadce666c7886780757976dc3b54b`

khushboo-rancher (changes requested):

Dockerfile.smoke
Vicente-Cheng: good catch!
updated

khushboo-rancher (approved):

PR #2824, opened 2026-08-14
Fix smoke template count and overcommit

Vicente-Cheng:

Issue #

Since we introduced more VM templates in v1.9.0, we need to fix the pytest part.
Also, add the same test case (vm template) on the robot tests
Improve the overcommit comparsion of memory

### DIFF
--- a/harvester_e2e_tests/apis/test_vm_templates.py
+++ b/harvester_e2e_tests/apis/test_vm_templates.py
@@ -80,6 +80,7 @@ def expected_default_templates(api_client):
 @pytest.mark.templates
 @pytest.mark.negative
 class TestVMTemplateNegative:
+    @pytest.mark.robot_ported
     def test_get_not_exist(self, api_client, unique_name):
         code, data = api_client.templates.get(unique_name)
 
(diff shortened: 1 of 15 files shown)


Vicente-Cheng:
smoke result with this change https://github.com/harvester/release/actions/runs/31763724182/job/94661173698
[test-results-48190cbe84c9d6acb3e9f469b773474424ac1d11-119-1.zip](https://github.com/user-attachments/files/31060262/test-results-48190cbe84c9d6acb3e9f469b773474424ac1d11-119-1.zip)

4 failures related to the broken live migration issue
test_migrate_vm_with_user_data, test_migrate, test_migrate_vm_with_cpu_pinned, test_maintenance_mode_trigger_vm_migrate

1 error looks like flaky
test_migrate_vm_with_user_data

None of them are related to the PR changes.

khushboo-rancher (dismissed):
nit: Should we use vm_tmeplate instead of template naming?

Vicente-Cheng:
> nit: Should we use vm_tmeplate instead of template naming?

good suggestions!  updated

khushboo-rancher (approved):

albinsun (approved):
lgtm

Issue #2813, opened 2026-08-06
harvester-rancher-integration-test: Run Tests failing across all RKE2 versions (v1.34/v1.35/v1.36) for custom cluster and imported cluster


khushboo-rancher:
Build <https://harvester-jenkins/job/harvester-rancher-integration-test/291/|#291> (harvester v1.9) went UNSTABLE with the `Run Tests` stage failing on **all three** RKE2 versions exercised:
```
- RKE2 v1.36: FAILED
- RKE2 v1.35: FAILED
- RKE2 v1.34: FAILED
```

Custom cluster and imported cluster creation failed - build number 291

khushboo-rancher:
Fixed, results on pipeline looks good.

PR #2829, opened 2026-08-14
Fix the custom and imported cluster tests

khushboo-rancher:

Issue # https://github.com/harvester/tests/issues/2813

After the change of token being changed to secret not the plain text anymore, the changes were not flown down to test cases


noahgildersleeve (approved):
LGTM

PR #2846, opened 2026-08-20
integrations: fix vm_shell_do control flow and broaden SSH retry

Vicente-Cheng:

N/A (smoke flake observed in CI)

`TestVMSnapshot::test_vm_snapshot_create` fails intermittently in smoke runs with `paramiko AuthenticationException: transport shut down or saw EOF`. Two defects in `vm_shell_do` cause this:

1. The whole SSH-and-run-action block sits **inside** the readiness poll loop. When the VM is already Running with the guest agent connected, the loop `break`s past the SSH block, so the action is silently skipped (the data write/verify steps of the snapshot suite become no-ops). When the VM is *not* ready yet, the code falls through and logs in exactly while cloud-init has not finished installing the SSH key - which is the AuthenticationException flake.
2. The connect retry loop only catches `ChannelException`, so an auth failure during guest boot aborts immediately instead of retrying.

This PR dedents the SSH block out of the poll loop (login only happens after the agent is connected, and the poll now times out explicitly), and broadens the retry to `(SSHException, NoValidConnectionsError, ConnectionResetError, TimeoutError)` following the existing pattern in `test_upgrade.py`. The same narrow catch in `fixtures/virtualmachines.py` (`vm_shell_from_host` / `vm_checker`) is broadened too, which also covers the long-standing paramiko flake seen in the clone/hotplug tests.

`ChannelException` is a subclass of `SSHException`, so the previous catches remain covered. The retry loops are still bounded by `wait_timeout`, so a genuine auth misconfiguration still fails - just at the deadline instead of on first attempt.

Flake instance: harvester/release smoke run 32254373183 (`test_vm_snapshot_create`, guest sshd accepted TCP but key provisioning had not completed).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/harvester_e2e_tests/fixtures/virtualmachines.py
+++ b/harvester_e2e_tests/fixtures/virtualmachines.py
@@ -7,7 +7,7 @@ from contextlib import contextmanager
 
 import pytest
 from paramiko import SSHClient, RSAKey, MissingHostKeyPolicy
-from paramiko.ssh_exception import ChannelException, NoValidConnectionsError
+from paramiko.ssh_exception import NoValidConnectionsError, SSHException
 
 
 @pytest.fixture(scope="session")
@@ -103,7 +103,8 @@ def vm_shell_from_host(vm_shell, host_shell, wait_timeout):
             while endtime > datetime.now():
                 try:
                     vm_sh.connect(vm_ip, jumphost=h.client)
-                except (ChannelException, NoValidConnectionsError) as e:
+                except (SSHException, NoValidConnectionsError,
+                        ConnectionResetError, TimeoutError) as e:
                     login_ex = e
                     sleep(3)
                 else:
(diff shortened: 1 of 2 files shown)


Vicente-Cheng:
Please don't review this one; I will run smoke first

smoke: https://github.com/harvester/release/actions/runs/32329255999
target branch is master
test branch is this worktree

Vicente-Cheng:
See the corresponding release smoke test: https://github.com/harvester/release/actions/runs/32343675350

Rerunning twice did not see any error with test_vm_snapshot_create

I will run it 5 times to ensure the issue was resolved with this fix.

Vicente-Cheng:
I think we are good to review this.
cc @khushboo-rancher, @lanfon72

khushboo-rancher (approved):

PR #2873, opened 2026-09-02
Fix e2e ssh retry and teardown

Vicente-Cheng:

Issue #

- Ensure the SSH login retry mechanism on test_3_vm_functions.py and test_4_vm_backup_restore.py
- Migrate test teardowns; now wait for volume deletion before deleting the source image

These flaky cases were found by following smoke testing:
- https://github.com/harvester/release/actions/runs/33520980330
- https://github.com/harvester/release/actions/runs/33460089214

### DIFF
--- a/harvester_e2e_tests/fixtures/settings.py
+++ b/harvester_e2e_tests/fixtures/settings.py
@@ -22,6 +22,18 @@ def setting_checker(api_client, wait_timeout, sleep_timeout):
             return False, (code, data)
 
         @wait_until(wait_timeout, sleep_timeout)
+        def enable_storage_net(self, vlan_id, cluster_network, cidr):
+            """Update the storage-network setting, retrying while the webhook
+            rejects it because a node's VlanStatus is not Ready yet - enabling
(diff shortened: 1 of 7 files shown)


Vicente-Cheng:
~wait for the https://github.com/harvester/release/actions/runs/33589712134~
see latest comment

Vicente-Cheng:
The current result: https://github.com/harvester/release/actions/runs/33714718561

I am trying to catch one more flaky test since I added the VMIM dump for the migration issue.

Vicente-Cheng:
BTW, here is the real race case we found while fixing the flaky cases.
https://github.com/harvester/harvester/issues/11582

Vicente-Cheng:
OK... I cannot catch the flaky case that I want. But I already dumped the vmim in case we encounter it again.

This PR is now ready for Review.

cc @khushboo-rancher, @albinsun, @lanfon72 please check with that.

albinsun (approved):

khushboo-rancher (approved):

Issue #2877, opened 2026-09-04
Rancher integration suite setup fails to install harvester-rbac chart – lost connection to cluster / failed to find Session


khushboo-rancher:
## Summary
`harvester-rancher-integration-test` nightly build 350 for Harvester v1.9 is UNSTABLE. All RKE2 versions tested in the same run fail at suite setup.

## Evidence
* Build: harvester-rancher-integration-test/350/
* Robot summary:
  - RKE2 v1.36: FAILED (rc=33)
  - RKE2 v1.35: FAILED (rc=33)
  - RKE2 v1.34: FAILED (rc=33)
* Representative test case errors:
  - `RKE2 v1.34: Test Rancher Integration` – `Test TC1 - Assign Cluster View Role` FAILED
  - `Parent suite setup failed:
Failed to install chart harvester-rbac: Error from server (InternalError): an error on the server ("lost connection to cluster: failed to find Session for client stv-cluster-local") has prevented the request from succeeding`

The same `Failed to install chart harvester-rbac` / lost connection / failed to find Session error is present for all test cases across v1.34, v1.35 and v1.36, indicating a reproducible setup failure rather than version-specific flake.

## Impact
Entire Rancher integration test matrix is blocked at suite setup; no test cases actually execute.

Please triage whether this is a Rancher/HAL session issue, a Harvester RBAC chart install regression, or test automation connectivity problem.|Rancher integration suite setup fails to install harvester-rbac chart – lost connection to cluster / failed to find Session

PR #2879, opened 2026-09-06
Make Rancher tests run in parallel

khushboo-rancher:
1. Make Rancher tests run in parallel using pabot
2. Use rest calls to install chart to avoid a Rancher bug where kubectl create fails while installing rbac chart.
     1. This was working when test is running from laptop.
     2. It is probably a bug in Rancher side. Lost connection is observed.
3. Made rbac moved away from suite setup and tear down - rbac tests are independent and should not impact other tests. Test suite setup and tear down is only for creating necessary resources used for Rancher suite.

issue - https://github.com/harvester/tests/issues/2877


Yu-Jack (approved):
Thanks!

albinsun (approved):

PR #2889, opened 2026-09-15
Fix `ModuleNotFoundError: No module named 'pkg_resources'` error

votdev:
... which is caused due https://github.com/pypa/setuptools/pull/5007.
```
Running tests...
==============================================================================
Test Vm :: VM Test Cases: basic lifecycle, create-stopped, and volume hot-p...
==============================================================================
Test VM Basic Lifecycle :: Test basic VM creation, start, stop, de... | FAIL |
Parent suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'
------------------------------------------------------------------------------
Create Stopped VM Stays Stopped :: A VM created with runStrategy=H... | FAIL |
Parent suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'
------------------------------------------------------------------------------
Start Stopped VM :: Starting the stopped VM brings it to running      | FAIL |
Parent suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'
------------------------------------------------------------------------------
Hot-Plug Volume Into Running VM :: Hot-plug an existing volume int... | FAIL |
Parent suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'
------------------------------------------------------------------------------
Hot-Unplug Volume From Running VM :: Hot-unplug the disk; it must ... | FAIL |
Parent suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'
------------------------------------------------------------------------------
Test Vm :: VM Test Cases: basic lifecycle, create-stopped, and vol... | FAIL |
Suite setup failed:
ModuleNotFoundError: No module named 'pkg_resources'

5 tests, 0 passed, 5 failed
```

Additional fixes:
- Fix crash when using a nightly build cluster which returns a git hash instead a semver.
```
InvalidVersion: Invalid version: 'c6d323ec'
```


harvester_robot_tests/README.md
```diff
@@ -112,7 +112,12 @@ source venv/bin/activate  # On Windows: venv\Scripts\activate
 ```
 
 3. Install dependencies:
+
+> [!IMPORTANT]
+> You need to install `setuptools` version 81 because `pkg_resources` has been removed in version >= 82. See https://github.com/pypa/setuptools/pull/5007.
+
 ```bash
+pip install setuptools==81.0.0
```
m-ildefons: Why not just add this to `test-requirements.in`?

The fact that this isn't pinned there is even warned about in `test-requirements.txt`:
```
# WARNING: The following packages were not pinned, but pip requires them to be
# pinned when the requirements file includes hashes and the requirement is not
# satisfied by a package already installed. Consider using the --allow-unsafe flag.
# setuptools
```

m-ildefons (commented):
AFAIK, the dependencies of the robot tests are managed through `requirements.in`/`requirements.txt` in the `harvester_robot_tests` directory, not the top-level requirements files.

Aside from that, I'd prefer to pin this with the other dependencies rather than have it as separate install command documented.

harvester_robot_tests/README.md
albinsun: Yeah, if this package is missing in a fresh robot env., we should add it in `requirement.in` and regenerate `requirement.txt` too.

votdev: pip-tools is considering `setuptools` as unsafe. But we can force it with `--allow-unsafe`. I think that’s fine in our case, as we specifically need this version of setuptools.

votdev: With my first attempt at the PR, I didn’t want to do anything that went against the ‘normal’ behaviour of pip-tools; but I think forcing it is ultimately the better and simpler solution.

irishgordo (changes requested):
@votdev  - would you be able to regenerate this with Python 3.10.x?
The reason I ask is because we've got a really tight coupling to Python 3.10 in our anisble-container that we've not been able to address -> so all tests are based on that major.minor release.

We can update docs too.
But my thinking was perhaps you could leverage `pyenv` , https://github.com/pyenv/pyenv.

Then just with this repo:
```
$ pyenv virtualenv 3.10.14 PR2889
# leverage, for what I recall is a solid 3.10.x compatible pip-tools
$ pip install pip-tools==7.5.3
$ pip-compile --generate-hashes requirements.in -o requirements.txt
# then test locally
$ pip install -r requirements.txt
```


For what it's worth I've built out: https://github.com/harvester/harvester-baremetal-ansible/issues/321 - to more track the tight coupling & issues we have because we ideally should address them before end of October this year as Python 3.10.x is EOL'd Oct

cc: @khushboo-rancher

votdev:
@irishgordo I've now updated the `requirements.txt` using python 3.10.21.

